PR #19, opened 2019-07-26
Testing Inventory

Stebalien:


Stebalien:
Not sure if this is the right repo, but we don't really have a better one.

testing/infra/benchmarks.md
```diff
@@ -0,0 +1,34 @@
...
+* [ ]Pprof profiles (go)
+* [ ]Bandwidth usage
+* [ ]allocation/goroutine totals
+* [ ]Time till $event (time series)
+* [ ]Packet rate.
+
+## Targets
+
+* [x] js-ipfs
+* [ ] go-ipfs
```
daviddias: It targets go-ipfs as well but only as a way to compare (e.g. doesn't capture any DTrace probes to create Flamegraphs for go-ipfs as we don't have them in go)

PR #20, opened 2019-08-12
Let's get this started 🔥  Small benchmarks, here we go!

raulk:
This is a work-in-progress PR to serve as a Proof of Concept of the latest design iteration. It introduces a test plan called "small benchmarks", which relies on IPTB to create testbeds to orchestrate small test cases similar in spirit to those of JS Benchmarks: https://github.com/ipfs/benchmarks/tree/master/tests

We introduce two test scenarios:

* adding a file
* adding a file, swarm connecting from another node, getting that file

We run these scenarios for a variety of file sizes.

All observations are recorded in ndjson on stdout.

We introduce an abstraction encapsulating IPTB to make it... nicer and programmatic.


iptb/ensemble.go
```diff
@@ -0,0 +1,133 @@
...
+		te.tags[k] = node
+
+		if tn.opts.Initialize || tn.opts.Start {
+			wg.Add(1)
+
+			go func(n testbedi.Core) {
+				defer wg.Done()
+
+				var err error
+				if tn.opts.Initialize {
```
raulk: Race.

Issue #23, opened 2019-09-02
DHT find_peers distributed test over 1000 nodes


raulk:
A test scenario that we'll use as an example/template to build the system gradually.

1. Initially this will run within a single process.
2. Run as multiple processes locally, coordinated by the sync service (Consul). Local test runner.
3. Deploy to a Nomad cluster. Nomad test runner.

daviddias:
Work on specing tests at #33

daviddias:
Needs https://github.com/ipfs/testground/issues/103 to get access to those 1000 nodes

daviddias:
This is now possible. Note for the reader, we are not using neither consul or nomad (spec is out of date at the moment). Check the tutorial on the README to learn how to run the find_peers tests and how to setup a cluster for yourself https://github.com/ipfs/testground#contributing--testing

PR #33, opened 2019-09-10
Specify top 10 Test Plans to be created to test go-ipfs with confidence

jimpick:
We'll use the PR to discuss what the tests should be.

I've started a blank document which we should fill out collaboratively and then merge when consensus is reached.

-------- 

Resources:

- [Test Improvement Tracker](https://docs.google.com/spreadsheets/d/1xyqyGUF-oe3x9ln88YonVeOMWWdknik74lVgL_3dBY8/edit#gid=0)
- Issues in this repo
  - DHT tests https://github.com/ipfs/testground/issues/23


jimpick:
Some desired tests are identified (at a very high level) in the "Testing improvement plan tracker":

https://docs.google.com/spreadsheets/d/1xyqyGUF-oe3x9ln88YonVeOMWWdknik74lVgL_3dBY8/edit?usp=sharing

jimpick:
Issue #23 has been identified as a test scenario to help guide the development/design of testground - DHT find_peers over 1000 nodes

jimpick:
I did a personal [mapping exercise](https://github.com/ipfs/testground/wiki/Brainstorming-Notes---Jim---Test-"Kernels") to try to arrange all the things I think we could potentially test in IPFS over the next few years. Here's a boiled down set of "starter tests" I think we could start work on in the next month.

These are just ideas of where we could start.

Each test represents a potential area we could expand on. We would start by writing the simplest possible tests that capture the essence of what we're trying to test. Ideally, if the infrastructure is in place, we should be able to write any of the initial tests in a day. Once we have the basics down, we can then expand the complexity of the tests to increase the fidelity.

# 1. Bitswap

I did some experimentation with go-ipfs, js-ipfs, Bittorrent (Bittornado, based on the original Python source), Dat, and plain old netcat with split files and cat. If I place replicas of some random content on 2 nodes that are bandwidth throttled, and attempt to replicate a 3rd, IPFS doesn't do very well, speed-wise. All the other solutions can pump the data at the throttled speed, and finish the download with an equal amount of data from each replica in half the time it would take to download from a single replica. I think some of the upcoming bitswap improvements might fix that, so that would make a good trivial test case.

It would be interesting to provide a foundation so that we could expand the test case out into more of a benchmark, where we have numerous peers creating data, seeding and downloading simultaneously, sometimes full downloads, sometimes sparse, some peers acting as download clients only, some peers actively resharing, with different lifetimes, with statistical models of common download patterns (eg. steady state, spikes, etc.). It gets really interesting when we try to model real life networks, latencies, packet loss, etc.

# 2. DHT

The DHT is tricky as it needs to be correct and robust, but it also needs to be efficient and performant, all while working on a heterogenous pool of resources "donated" by IPFS peers spread across the Internet that come and go, or more around, and aren't always reachable or fast. It's a complex system, and there are a lot of proposals on ways to enhance it. We need a good torture test environment for evaluating changes before we roll them out into the wild. Being able to fully trace all peers end-to-end in a simulated environment will be valuable in validating the correctness of our implementations and to help debug performance issues.

It's probably going to be quite difficult to model and accurately simulate the conditions that the real life DHT network has - we'll have to be very creative and do a lot of testing and measurement on the live network as well.

# 3. Scalability to large data sets - Many files / large files

Identifying scalability bottlenecks will open up the potential of using IPFS for many "big data" applications, for example, reproducible scientific studies or machine learning corpora. Ideally, IPFS performance should scale linearly as more peers are added. A benchmark would test performance for adding huge numbers of files across a cluster of peers, and test retrieval and transfer speeds, as well as garbage collection speed.

# 4. Connectivity with NATs - Relaying / Hole Punching / WebRTC

Even though peers may be located behind firewalls, in many cases, they should still be able to fully participate in the IPFS Network using a variety of networking techniques. We would like to be able to build tests that allow peers to be located behind NATs so we can evaluate the best techniques to use, and the impact on the overall network.

Down the road, there are a number of other "exotic" networking technologies and topologies such as multicast, information centric networking, etc. that the tests could be expanded to include.

# 5. Saturation + efficiency tests - constrained cpu, memory, bandwidth

We would like to be able to force the peers under test into various states where they are fully saturated - cpu, memory, network bandwidth, etc. and evaluate how they react and the congestion effects on the rest of the network. This can be achieved by artificially capping certain resources and applying excessive loads in order to achieve the saturated states. We can add observability to make it easier to identify saturation and congestion in real life and test loads, as well as design strategies to cope with excessive loads (eg. load shedding, backpressure, queuing, replication, etc.).

# 6. Pubsub

Having a pubsub mechanism that works well between a small number of peers is extremely valuable. It's important that the peers can find each other and maintain membership. Scaling pubsub becomes very challenging as the number of peers increase and the message volume increases. Inevitably there will tradeoffs to consider when saturation starts to hit, or networks are poorly connected. Having an environment where we can test and generate traces on large numbers of peers participating in a channel will help a lot.

# 7. Package Managers - npm on IPFS

For 2019, the IPFS [top-ranked goal](https://blog.ipfs.io/78-ipfs-2019-roadmap/) is supporting the needs of package managers.
For this test, we would setup several peers running the [npm-on-ipfs](https://github.com/ipfs-shipyard/npm-on-ipfs) registry and schedule many peers to make builds that fetch dependencies from the registry.

# 8. Blogging benchmark (with IPNS, Gateway, Cluster, Companion)

A healthy decentralized web ecosystem would be characterized by a return of old school blogging ... individual blogs and microblogs, RSS feeds, Indieweb, image feeds, podcasts, videos, etc. in place of the highly centralized social media platforms that took over.

A full simulation could grow to be quite complex, but would be quite valuable as it would identify the pain points that are holding back the adoption of the decentralized web.

An initial test might be to just publish a static blog on IPFS with IPNS, and then use Puppeteer (Chrome) to download it via a standard HTTP gateway on a separate node.

A full benchmark would spin up a large number of IPFS nodes to host content (perhaps using IPFS Cluster), plus clients that would be creating blog content, uploading images, audio, video, as well as clients that retrieve content using realistic traffic patterns. Some clients would be web based and fetch via HTTP gateways, whereas others would be mobile apps and fetch raw data directly. Feeds would be published using IPNS. Some web clients may have IPFS Companion installed and will fetch content semi-directly using preload nodes or via delegated routing. Clients and servers may use spidering software to crawl the available sites and build indexes. Web archiving software would crawl non-IPFS sites and preserve them for the future.

# 9. Database Access Patterns Benchmark

A compelling use case for IPFS in the future is using it as a foundational layer for building highly distributed and scalable database systems. The access patterns are different than filesystem access patterns, and it is challenging to achieve high performance in a very decentralized environment.

There are many, many projects building on top of IPFS than could be tested.

* [IPLD](https://ipld.io/)
  * [js-ipld-hashmap](https://github.com/rvagg/js-ipld-hashmap)
* IPNS / IPRS
* [ipfs-geoip](https://github.com/ipfs/ipfs-geoip) (used by WebUI)
* [go-ds-crdt](https://github.com/ipfs/go-ds-crdt) (used by ipfs-cluster)
* [IDM](https://medium.com/moxystudio/the-need-for-decentralized-identity-4d369408e10e)
* [iGiS](https://github.com/ipfs-shipyard/IGiS)
* vmx: [GeoData on IPFS](https://media.ccc.de/v/bucharest-401-geodata-on-ipfs)
* [OrbitDB](https://github.com/orbitdb/orbit-db)
* [Textile Threads](https://docs.textile.io/concepts/threads/) - v2 coming soon
* [Qri](https://qri.io/)
* [Tupelo Chaintree](https://docs.quorumcontrol.com/docs/chaintree.html)
* [PeerGOS](https://peergos.org/)

# 10. Security and Availability with Adversaries

The Internet is a chaotic environment and there are bad actors that may wish to attack the network with the objective of knocking content off of it, or otherwise corrupt content. 

It would be useful to have a test environment where Denial-of-Service scenarios could be tested, and to measure what level of effort is needed to damage the network, and to develop resistance techniques.

momack2:
@Stebalien and @raulk likely have a more authoritative picture of what is required in order to release 0.5.0 with the significant DHT changes currently in the `stabilize` branch (and other features currently blocked), but my guess is that 1, 2, and 4 are the *highest* priority areas, with 3 as a close runner-up. I think we should focus this first set of tests there, though having the broader landscape of things we want to test is super useful!

Maybe something along the lines of:
1. Bitswap
   - spin up X nodes, were Y have desired data - try to spread data to all X nodes
2. DHT
   - add xx,000 nodes to DHT, try to find and get desired data on one node
   - add xx,000 nodes to DHT with half running 0.4.22 and half running master/nightly/etc, try to find and get desired data on one node
3. NAT
   - add xx,000 nodes to DHT, where 50% / 80% / 95% are NAT'd - try to find and get desired data on one node (that is / isn't NAT'd)
4. Scale
   - add XXMB to one node, unpin data as transfer to node 2, run GC?

autonome:
Overall, there are lots of standalone projects in this list... great candidates for devgrants/bounties?

> It would be interesting to provide a foundation so that we could expand the test case out into more of a benchmark...

+1 to a benchmark, automated to run regularly, and results published publicly. Even if we perform poorly at the beginning, every improvement shows, and the transparency/authenticity shines through. These types of "arewe*yet" metrics provide a cause to rally around and help keep project focus and community awareness on this type of issue - the kind of issue that technology selection decisions are made on.

> # 2. DHT
> 
>  all while working on a heterogenous pool of resources "donated" by IPFS peers spread across the Internet that come and go, or more around, and aren't always reachable or fast.
> ...
> we'll have to be very creative and do a lot of testing and measurement on the live network as well.

Given that the transient nature of the DHT is a core characteristic, I wonder if we should be measuring things like median node lifetime to better understand how that impacts performance, etc? Kinda like an ongoing "sparseness" measurement.

This likely falls into the area of live network testing... hm, actually if it's a core characteristic of the network IRL then we probably *should* be modeling it in our test scenarios.

> # 6. Pubsub

+ 1, the need for PubSub to move into production was a common request in the interviews I did at Camp. Performance, scale and torture tests for PubSub would help us communicate the relative level of stability of the feature so that IPFS users can evaluate it against the specifics of their user base and use-cases.

> # 7. Package Managers - npm on IPFS
> 
> For 2019, the IPFS [top-ranked goal](https://blog.ipfs.io/78-ipfs-2019-roadmap/) is supporting the needs of package managers.
> For this test, we would setup several peers running the [npm-on-ipfs](https://github.com/ipfs-shipyard/npm-on-ipfs) registry and schedule many peers to make builds that fetch dependencies from the registry.

Seems like this might be a high-value initial priority. As an end-to-end test it'll uncover problems we already know about (those first few test categories you described) but will let us know when we make progress in those areas that helps our highest priority use-case. It provides an external readiness signal for the package manager communities.

> # 8. Blogging benchmark (with IPNS, Gateway, Cluster, Companion)
>
> An initial test might be to just publish a static blog on IPFS with IPNS, and then use Puppeteer (Chrome) to download it via a standard HTTP gateway on a separate node.

This is a great candidate for bounty, devgrant, maybe should be in the ipfs mini projects repo?

Also, I made a puppeteer wrapper for Firefox a while back: https://github.com/autonome/puppeteer-fx

jimpick:
@autoname wrote:

> Given that the transient nature of the DHT is a core characteristic, I wonder if we should be measuring things like median node lifetime to better understand how that impacts performance, etc? Kinda like an ongoing "sparseness" measurement.

I was thinking it might be nice to have a "DHT Record Observatory" that monitored popular CID hashes on the live network, and polled which peers were serving records for them over time. It would be interesting to see how long things live the DHT before they drop out.

autonome:
Another thought - it would be ideal if these tests were all implementation-agnostic, so we can test Go / JS / Rust / Python / etc.

mgoelzer:
Given that spinning up ~10^5 nodes is expensive, it might be worth running the entire test suite first on 3 nodes, to catch trivial errors, then iff all green, re-run on the large cluster.

daviddias:
Folks, I'm starting to create the Test Narratives. I've just pushed the first one, more to come. If you any comments right now on the format in how it is written, let me know so that I adapt on the fly :) 

PS: Am I a weird person for getting super psyched on just the thought of seeing these tests run? :D

daviddias:
> Another thought - it would be ideal if these tests were all implementation-agnostic, so we can test Go / JS / Rust / Python / etc.

Agreed. Although, AFAIU this is a secondary goal given the short timeline. The best tooling that exists right now to deal with multiple implementations is in JS due to the complete http-client and the ipfsd-ctl which lets spin any kind of node.

daviddias:
> Given that spinning up ~10^5 nodes is expensive, it might be worth running the entire test suite first on 3 nodes, to catch trivial errors, then iff all green, re-run on the large cluster.

One should be able to run a 1 or 2 dozen nodes in their local machine. Agreed that a test shouldn't go into massive scale if one can't even get it to pass locally.

daviddias:
I believe these test narratives should make it very clear what is being tested, how and what we expect to measure so that we can then optimize. They should not, however, be overly perspective with details and rather give some freedom to the implementer or to testground to offer ways to implement them (e.g. how data is collected, visualized, etc).

@raulk, @Stebalien would love a comment if you feel I'm on the right direction here

@jimpick would love your comments to check if you were to happen to implement all of these, do you feel that the text has all the information you need and if not, what else?

docs/tests-mvp.md
```diff
@@ -0,0 +1,265 @@
...
+This test is not expected to support:
+
+- An heterogeneus network in which nodes have different configurations
+
+#### Test Narrative
+
+- **Warm up**
+  - Boot N nodes
+  - Connect each node to the node next to it (hash ring)
+  - Run multiple DHT random-walk queries to populate the finger tables
```
aschmahmann: Feels like we're mixing tests here that is bound to cause problems? Any reason not to test the DHT separately and manually control node connections for the Data Transfer tests?

raulk:
@daviddias I think these narratives are great discussion starters, and material we can refine into "musical scores" that explain _exactly_ the series of steps each test consists of.

I'd imagine I'm writing a script for a theatrical play. I'd need N actors, to play roles A, B, C. When the curtains opens, actor 1 is going to do X, actor 2 is going to do Y, actors 1 and 2 will have a dialogue, etc... Basically a step-by-step guide of what happens until the curtain call. You get the analogy.

We'd also benefit from defining the parameters of each test that can be adjusted when instantiating it, like you've already done, e.g. size of files 👍  We want to avoid hard-coding such parameters, in order to make tests flexible and useful under many circumstances (e.g. in the DHT lookup-peers test case, we can adjust the `bucket_size`).

Perhaps an appropriate mindset to adopt when writing this material is to think you're gonna outsource the writing of these cases to an external group that is not familiar with the priorities and use cases we want to subject IPFS to (so they don't know _what to test_), but they are experts in the technical design of IPFS (so you can talk about Bitswap, Graphsync, DHT, pubsub, etc.)

mgoelzer:
From a product standpoint, it would be awesome if we could have one or more simple but reproducible smoke tests that could give us confidence in a new PR or change.

It's sad panda to have a branch that we think will improve DHT performance, but is blocked from merging because we don't know what kind of side effects it might cause at scale.  Almost any test suite that outputs the same numerical metrics each time its run could help give us more insight into this, and thus confidence in deploying the change.

mgoelzer:
One other thought is a QUIC test.  FC and IPFS might need this soon for performance.

docs/tests-mvp.md
daviddias: @aschmahmann this step is specifically in the Warm up phase to "set up the network into the testing state" (see description at the top of the document), it is not testing the DHT, however, it is important to have a well connected fabric to be able to test the rest. There will be another test that focuses on the DHT connectivity.

daviddias:
> @daviddias I think these narratives are great discussion starters, and material we can refine into "playbooks" that explain exactly the series of steps each test consists of.

Great, this answers my question from https://github.com/ipfs/testground/pull/33#issuecomment-536281469. I read that you want more specificity 👍🏽

> I'd imagine I'm writing a script for a theatrical play. I'd need N actors, to play roles A, B, C. When the curtains opens, actor 1 is going to do X, actor 2 is going to do Y, actors 1 and 2 will have a dialogue, etc... Basically a step-by-step guide of what happens until the curtain call. You get the analogy.

Absolutely, 100% on the same page here. See some examples where I did this:

image

In this one I divided the nodes by %, % that describe the different properties and how heterogeneous the network is.

image

In this one I organized the % in Cohorts that take different Roles, a Role that will take a different set of actions.

> We'd also benefit from defining the parameters of each test that can be adjusted when instantiating it, like you've already done, e.g. size of files 👍 We want to avoid hard-coding such parameters, in order to make tests flexible and useful under many circumstances (e.g. in the DHT lookup-peers test case, we can adjust the bucket_size).

I was calling this "Execution Variants", but you called it right, Test Parameters it is what is being named today and there is no reason to rename. I'm fixing that.

I see that there are cases in which we have this parameters to adjust properties of the image (bucket size, bitswap strategy and so on) and in other cases we have parameters to adjust the network properties (% of nodes using QUIC, TCP, UTP, etc). I'll give a twist here to make these more explicit so that we can declare: How many images there are, the parameters to these images and also the network parameters. 

> Perhaps an appropriate mindset to adopt when writing this material is to think you're gonna outsource the writing of these cases to an external group that is not familiar with the priorities and use cases we want to subject IPFS to (so they don't know what to test), but they are experts in the technical design of IPFS (so you can talk about Bitswap, Graphsync, DHT, pubsub, etc.)

That's where I'm headed :D I guess from your point I take that I'm still far :) Good to know, that is the kind of feedback I needed.

> From a product standpoint, it would be awesome if we could have one or more simple but reproducible smoke tests that could give us confidence in a new PR or change.

@mgoelzer Could you be more quantitative and qualitative here? One could argue that we do have a ton of "simple & reproducible smoke tests" already, just check the massive test inventory we have here https://github.com/ipfs/testground/pull/19. However, we do all agree that the current test inventory is not enough (not large enough, not fast enough, not automated enough, not precise enough, not useful enough and so on).


Thank you all for this round of feedback, back to the writing! :D

daviddias:
As I improve the descriptions and try to come up with ways in how to describe what needs to happen, I'm starting to think that there is a lot of value to get one of these Test Plans implemented completely (the test does the thing, we capture the data points we need and we present the output in an organized way that one could infer if there has been improvements or not) so that we can fine tune the descriptions as we go (vs. water-falling a truck load of test narratives).

The one that seems closer to implement given the current set of features in TestGround is _Test Plan 1. Chewing strategies for Large DataSets_. @jimpick wanna tag team to get this one done fast? I believe it will also help you make progress on setting up the infro and gathering the profiling/metrics to be displayable :)

mgoelzer:
@daviddias 
> just check the massive test inventory we have here #19.
Thanks -- good point.  I think my issue with having some smoke tests before a deploy to production IPFS is probably better addressed in #19 anyway.

jimpick:
I'm loving the narratives... it's helping to clear away the fog and I can see pathways towards getting these implemented and parameterized in a shorter time frame.

@daviddias I'll set my sights on Test Plan 1 ... it will be a good exercise to try to map that into the framework that @raulk has built.

daviddias:
Awesome!

> @daviddias I'll set my sights on Test Plan 1 ... it will be a good exercise to try to map that into the framework that @raulk has built.

@jimpick I was working under the assumption that you were tackling the Infrastructure bit as discussed last week and I got started on writing (well, figuring out given that my go skillz aren't sharp yet) the Test 1. To avoid duplicating work, I believe it is still best for you to land the [remote deployments bit](https://trello.com/c/9uvG4oTv/30-set-up-aws-infrastructure-docker-machine-d-aws) & [piping results to Elastic Search](https://trello.com/c/JVnpC5e8/33-collect-container-outputs-into-elasticsearch-results-metrics-logs).

daviddias:
[Create 10~15 Test Case Scenarios](https://trello.com/c/0erPZId8/43-create-1015-test-case-scenarios)

docs/tests-mvp.md
aschmahmann: >There will be another test that focuses on the DHT connectivity

👍 

> it is not testing the DHT

Maybe not fully, but if the DHT is not working to some specification then the test will not perform as it should. We may be confident that the DHT currently works enough to satisfy this specification, but it doesn't mean that we haven't coupled the tests. Given that we already `Connect each node to the node next to it (hash ring)` shouldn't we just manually populate the finger tables with the appropriate random distribution?

docs/tests-mvp.md
```diff
@@ -0,0 +1,402 @@
...
+- **Warm up**
+  - Create the Bootstrapper nodes that are connected among themselves and support every transport
+- **Act I**
+  -
+- **Act II**
+  -
+- **Act III**
+  -
+
+## `Plan:` Providing Content (Content Routing / DHT)
```
aschmahmann: @daviddias Before I forget to comment, just wanted to point out that IMHO the biggest change we'll see in the DHT (until we have a message oriented DHT) will come from removing the clients behind NATs. The idea of having nodes remove themselves from the network based on information they receive from AutoNAT should hopefully do the job, but this change has the serious ability to make/break the network.

This is obviously a separate logical issue from testing DHT query resolution, but no matter where we decide to test it I think it's important we do so explicitly.

daviddias:
So that it is not lost
image

docs/tests-mvp.md
daviddias: Thanks! There is a Plan that is intended to capture the NAT challenges well. We will continue iterating on the Narratives as well.

daviddias:
As discussed during the last weekly sync https://github.com/ipfs/testground/pull/69, I'm merging this PR so that we can use it as a reference point as we iterate in each Test Plan independently.

PR #47, opened 2019-09-27
propagate test parameters from CLI to runenv.

raulk:
Now we can provide parameters to test cases. There are three types of parameters/options passable via LCI:

* builder configuration overrides, e.g. bypass the cache.
* runner configuration overrides, e.g. log level.
* test parameters: settings that are specific to the test behaviour, e.g. bucket size, timeout values, etc.


raulk:
Merging this, as I don't expect any pushback

PR #50, opened 2019-09-27
refactor sync/Writer + godocs

Stebalien:
* Avoids using a goto as a loop.
* Uses a defer for simpler locking.

### DIFF
--- a/sdk/sync/write.go
+++ b/sdk/sync/write.go
@@ -50,30 +50,34 @@ func NewWriter(runenv *runtime.RunEnv) (w *Writer, err error) {
 		refreshset: make(map[string]struct{}),
 	}
 
-	go w.refreshOwned()
+	go w.refreshWorker()
 	return w, nil
 }
 
-// refreshOwned runs a loop that refreshes owned keys every `RefreshPeriod`.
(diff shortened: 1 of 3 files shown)


daviddias:
@raulk if you approve, then merge? Taking you as the master tech lead here, less code floating in a fast moving project, the better.

sdk/sync/write.go
```diff
@@ -50,30 +50,34 @@ func NewWriter(runenv *runtime.RunEnv) (w *Writer, err error) {
...
 		}
-		w.lk.RUnlock()
-		goto Loop
+	}
+}
 
-	case <-w.doneCh:
-		return
+// refreshOwned refreshes all owned keys.
+func (w *Writer) refreshOwned() {
```
daviddias: Can I kindly ask for an improved explanation? Reading through this function the first time, it seems to me that it is checking which nodes are still alive and sets a TTL for their key. However, my understanding either doesn't seem complete or right. Can I haz some wisdom sprinkles?

raulk: Yeah, let me enhance the comments in that source file!

PR #58, opened 2019-10-03
feat: first tests of Chew Large Datasets

daviddias:
@daviddias when this issue was created:

> I'm slowly making progress as I get myself used to how the whole testground machine is set up. Creating this Draft PR so that I can ask questions as I go.
> 
> My apologies in advance for any obvious/beginner questions. I know that soon enough it will all be clear, it is just a matter of pouring hours into it :)

This adds a new plan called Chew Large Datasets. Please read the [README](https://github.com/ipfs/testground/blob/test-plan-1/plans/chew-large-datasets/README.md) for further information about each test plan and how it works.

Tracking comment: https://github.com/ipfs/testground/issues/93#issuecomment-562926159


daviddias:
So far, it has been useful already as I was able to adjust the language and structure of ipfs/interface-go-ipfs-core#33 with https://github.com/ipfs/testground/pull/33/commits/247bf38fff3bef22c1690d353c219c6aa796290e

manifests/chew-large-datasets.toml
```diff
@@ -0,0 +1,74 @@
...
+enabled = true
+
+[run_strategies."local:exec"]
+enabled = true
+
+[run_strategies."cluster:nomad"]
+enabled = true
+
+[[testcases]]
+name = "ipfs-add-defaults"
```
daviddias: ❓ I'm failing to replicate the dark magic that happens on the DHT manifest+plan in which a `lookup-peers` (https://github.com/ipfs/testground/blob/master/manifests/dht.toml#L28)  gets correctly set as an env var that it matches so that the test is selected at https://github.com/ipfs/testground/blob/master/plans/dht/main.go#L16 when `TESTGROUND_BASEDIR=`pwd` testground run dht/lookup-peers --builder=docker:go` is executed.

plans/chew-large-datasets/main.go
```diff
@@ -0,0 +1,26 @@
+package main
+
+import (
+	"github.com/ipfs/testground/sdk/runtime"
+	test "github.com/ipfs/testground/plans/chew-large-datasets/test"
+)
+
+var testCases = []func(*runtime.RunEnv){
+	test.IpfsAddDefaults,
```
daviddias: ❓ I'm trying to run this one manually with:

```
> TEST_CASE="IpfsAddDefaults" go run main.go
panic: test case sequence number not set
```

Assuming that I'm reading https://github.com/ipfs/testground/blob/master/sdk/runtime/runenv.go#L20 right. As you can see, it fails. Any tip please? :)

manifests/chew-large-datasets.toml
```diff
@@ -0,0 +1,78 @@
...
+
+# seq 3
+[[testcases]]
+name = "ipfs-url-store"
+instances = { min = 1, max = 1, default = 1 }
+
+  [testcases.params]
+  some_param = { type = "int", desc = "some param", unit = "peers" }
+
+# seq 4
```
daviddias: ❓ I learned now by tinkering with the dht tests that these seq numbers are needed (as in, it will output it doesn't have a seq number if missing from here or env variable), but it is not obvious why.

manifests/chew-large-datasets.toml
daviddias: Update: When running

```
> TESTGROUND_BASEDIR=`pwd` testground run chew-large-datasets/ipfs-add-defaults --builder=docker:go
```

I get

```
# ... docker logs
unknown runner:
```

What is this runner exactly?

plans/chew-large-datasets/main.go
daviddias: Update: learned through -- -- that TEST_CASE_SEQ is needed for the test selection. Now getting it to work locally

```
> TEST_CASE="ipfs-add-defaults" TEST_CASE_SEQ="0" go run main.go
Yo
{"runenv":{"test_plan":"","test_case":"ipfs-add-defaults","test_run":"","test_seq":0,"test_instance_count":-1},"timestamp":1570110821332210000,"result":{"outcome":"ok"}}
```

This means that the test case is actually being picked by the order in which they show up testCases array rather than the name. (wasn't obvious)

manifests/chew-large-datasets.toml
daviddias: For reference:

- https://github.com/ipfs/testground/blob/master/pkg/engine/engine.go#L183-L186
- https://github.com/ipfs/testground/blob/master/pkg/engine/engine.go#L43-L44

daviddias: Nvm. It requires one more env var

```
> TESTGROUND_BASEDIR=`pwd` testground run chew-large-datasets/ipfs-add-defaults --builder=docker:go --runner="local:docker"
```

or

```
> TESTGROUND_BASEDIR=`pwd` testground run chew-large-datasets/ipfs-add-defaults --builder=docker:go --runner="local:exec"
```

jimpick:
Is this working? Should I try to run it?

daviddias:
@jimpick It is working but it does nothing special yet. It is mostly bunch of scaffolding/setup code. I'm working my way through the testground codebase and learning how to yield IPFS daemons using the IPTB setup (#60) and/or direct in-proc nodes https://github.com/ipfs/go-ipfs/pull/6695.

I'll change the state of the PR from `draft` to `ready to review` when ready.

daviddias:
Quick update: Still slaying dragons at -- --, I am super close though and I should be able to go back to writing the tests here today using that as a base.

For reference, also learned the way that Netflix spawns their IPFS daemons, super clean https://github.com/Netflix/p2plab/blob/master/peer/peer.go

daviddias:
Getting the following panic when I run:

```
» go build . && TESTGROUND_BASEDIR=`pwd` ./testground run chew-large-datasets/ipfs-add-defaults --builder=docker:go --runner="local:exec"
resolved testground base dir from env variable: /Users/imp/code/go-projects/src/github.com/ipfs/testground
found cached docker image for: testground-chew-large-datasets:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
panic: reflect: New(nil)

goroutine 1 [running]:
reflect.New(0x0, 0x0, 0xc0004d5c80, 0x0, 0x0)
        /usr/local/go/src/reflect/value.go:2351 +0xc1
github.com/ipfs/testground/pkg/engine.coalesceConfigsIntoType(0x0, 0x0, 0xc0003164b0, 0x2, 0x2, 0xc0003164b0, 0x1, 0x2, 0x0)
        /Users/imp/code/go-projects/src/github.com/ipfs/testground/pkg/engine/engine.go:244 +0x360
github.com/ipfs/testground/pkg/engine.(*Engine).DoRun(0xc0000b3b00, 0x7ffeefbff8c9, 0x13, 0x7ffeefbff8dd, 0x11, 0x7ffeefbff90c, 0xa, 0xc0000d0b90, 0x0, 0x0, ...)
        /Users/imp/code/go-projects/src/github.com/ipfs/testground/pkg/engine/engine.go:208 +0x295
github.com/ipfs/testground/cmd.runCommand(0xc000196160, 0x0, 0xc0004fe380)
        /Users/imp/code/go-projects/src/github.com/ipfs/testground/cmd/run.go:128 +0x64b
github.com/urfave/cli.HandleAction(0x191ff20, 0x1adde90, 0xc000196160, 0xc000196160, 0x0)
        /Users/imp/code/go-projects/pkg/mod/github.com/urfave/cli@v1.22.0/app.go:523 +0xbe
github.com/urfave/cli.Command.Run(0x1aa32d8, 0x3, 0x0, 0x0, 0x0, 0x0, 0x0, 0x1ad1ebf, 0x62, 0x0, ...)
        /Users/imp/code/go-projects/pkg/mod/github.com/urfave/cli@v1.22.0/command.go:174 +0x51c
github.com/urfave/cli.(*App).Run(0xc00018a000, 0xc0000d0000, 0x5, 0x5, 0x0, 0x0)
        /Users/imp/code/go-projects/pkg/mod/github.com/urfave/cli@v1.22.0/app.go:276 +0x718
main.main()
        /Users/imp/code/go-projects/src/github.com/ipfs/testground/main.go:37 +0x244
```

Which comes from parsing the TOML manifest file -- https://github.com/ipfs/testground/blob/master/pkg/engine/engine.go#L228-L247

Scratching my head trying to figure out how is the manifest wrong. Tried using TOML validators including https://github.com/BurntSushi/toml and it parsed the manifest just fine.

Any tips?

Stebalien:
@daviddias The `LocalExecutableRunner` runner ("local:exec") doesn't specify a config type.﻿

https://github.com/ipfs/testground/blob/a428f18df934d0973417648f427ff4e5f14fa5ba/pkg/runner/local_exec.go#L97

Either `coalesceConfigsIntoType` should handle `typ == nil` (and likely simply return early) or we should actually return valid types from these methods.

https://github.com/ipfs/testground/blob/a428f18df934d0973417648f427ff4e5f14fa5ba/pkg/engine/engine.go#L228-L247

daviddias:
aaaah! Epic facepalm! you are right. There is no such runner as local:exec yet. I completely missed that it was my param that was wrong and was looking for the bug in the wrong place. Thank you @Stebalien !!

plans/chew-large-datasets/test/ipfs_add_trickle_dag.go
```diff
@@ -33,11 +33,12 @@ func IpfsAddTrickleDag(runenv *runtime.RunEnv) {
 		return
 	}
 
-	addOptions := coreopts.UnixfsAddOption([]coreopts.UnixfsAddSettings{
-		coreopts.UnixfsAddSettings.Layout(1), // 1 for Trickle Dag https://godoc.org/github.com/ipfs/interface-go-ipfs-core/options#Layout
-	})
+	addOptions := func(settings *coreopts.UnixfsAddSettings) error {
+		settings.Layout = coreopts.TrickleLayout
+		return nil
+	}
```
daviddias: Thank you @hacdias! I wasn't figuring this goism (if it is a goism) out

daviddias: Can you explain to me if this is in fact idiomatic go and if it is, why sometimes use functional arguments such as https://github.com/ipfs/testground/blob/master/plans/dht/test/create-dht-node.go#L22-L25 and other times do like this?

hacdias: @daviddias I've used Go in the past but I've never actually seen something like this. As far as I understand, what we have here is a function that takes a setting as argument and sets the option and returns **an error if it occurred**.

Whoever wrong the code in first place is certainly more capable of explaining why they choose to do it this way than I am. And I'm also curious

daviddias: @hacdias thank you! @Stebalien could you shed some of wisdom here?

aschmahmann: @hacdias this is a pattern we frequently use in our code base https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis. It essentially hacks around Go not supporting optional parameters, by leveraging its support of variadic arguments.

The idea is that you call some function (e.g. `ipfs.Unixfs().Add`) with the parameters it requires (e.g. `ctx`), and optionally pass in any additional parameters via functions (e.g. `ipfs.Unixfs().Add(ctx, option1, option2, ....). The `error` that you noticed is part of the **function signature** and as you can see no error is actually returned. However, it's possible that sometimes we might want to return an error. For example, if we want to only set settings.A or settings.B, then if B is set after A we can throw an error. Another example is if inside the option we do some IO operation that might fail, then we can still handle the IO error. Check out that link for more info.

daviddias: @aschmahmann I get that blog post and I believed I understood it well when I wrote https://github.com/ipfs/testground/blob/master/plans/dht/test/create-dht-node.go#L22-L25. 

However, the way to get the `trickle-dag` option to Unixfs is way different in nature, is is literally instead of passing the result of a func call (just like https://github.com/ipfs/testground/blob/master/plans/dht/test/create-dht-node.go#L22-L25 and what dave cheney describes) it takes in a function that received the config object to then change it.

aschmahmann: Yes, they are both part of the same solution.

If you look at "Way A" you'll see that you used `dhtopts.BucketSize(bucketSize)`, if you look at `dhtopts.BucketSize` https://github.com/libp2p/go-libp2p-kad-dht/blob/4b7d6fb886de322dd54d7c597e8ab163e087ad74/opts/options.go#L151 you'll see that it's a function that returns a function with the same signature as "Way B". In short this means "Way A" is "Way B" wrapped in another function for convenience and readability.

To make this even more explicit the code here should be replaced with `options.Unixfs.Layout(options.TrickleLayout)` where `options` is `github.com/ipfs/interface-go-ipfs-core/options`. We actually already have a "Way A" wrapper for the function created here 😄.

I'm not sure if we have an explicitly defined pattern for where/how we put these wrapper "Way A" functions. However, they're generally defined and searchable via code-completion or full-text searching the repo that the options are defined in.

hacdias:
@daviddias here, when you wrote:

```
 - `Directory Depth` - An Array containing objects that describe how deep/nested a directory goes and the size of files that can be found throughout (default to `[{depth: 10, size: 1MB}, {depth: 100, size: 1MB}]`
```

- How many files should there be per depth level?
- Each file is of size, right?

daviddias:
> How many files should there be per depth level?

I was just considering having the files at the directory leaves

> Each file is of size, right?

Yes

hacdias:
@daviddias agreed, having files scattered around every directory may not help with any debugging because we're just trying to make sure it can add a directory correctly!

hacdias:
Is anyone aware how to add files to MFS using CoreAPI? Couldn't find any of the MFS commands. What about the CG too?

/cc @daviddias @Stebalien

hacdias:
I am also hitting an issue: when adding big files (1/2 > GB), Unixfs().Add() fails with `too many open files`. I think I close everything I open so I believe it's somewhere on the API itself. Is anyone aware of that?

daviddias:
- https://github.com/ipfs/testground/pull/58#issuecomment-561586343 is a issue for https://github.com/ipfs/interface-go-ipfs-core/issues and https://discuss.ipfs.io/
- https://github.com/ipfs/testground/pull/58#issuecomment-561587740 seems a bug and must be reported at https://github.com/ipfs/go-ipfs/issues

Remember, although we are all under the same umbrella (the IPFS Team), it is important to use the right channels accordingly for each project. If you find issues in go-ipfs, then use the go-ipfs channels to report it and to get knowledge from the go-ipfs core team. Then you can link to those issues in this PR.

You can see how I've been practices what I preach with: https://github.com/ipfs/interface-go-ipfs-core/issues/48, https://github.com/ipfs/interface-go-ipfs-core/issues/46 and https://github.com/ipfs/go-ipfs/pull/6695

hacdias:
@daviddias done! I've also updated the main comment of this PR to show what's happening in a quick glance!

hacdias:
@daviddias I've noticed you created a `create-ipfs-daemon.go` file. What's the idea behind it? Run everything with a daemon and an instance?

hacdias:
One information here: I will be working on adding the needed interfaces for the Files (MFS) api to core api and go-ipfs :)

See more https://github.com/ipfs/interface-go-ipfs-core/issues/53#issuecomment-561889994

daviddias:
That's excellent! 

@hacdias here is a proposal. What about moving tracking things (the ones found in this thread) to https://github.com/ipfs/testground/issues/93 and get this PR merged so that you can make consequent smaller PRs?

daviddias:
> @daviddias I've noticed you created a create-ipfs-daemon.go file. What's the idea behind it? Test every test against a daemon and an instance?

It was both that and my understanding that some features would only be available through the daemon.

When you create that file, you could also create an example inspired by https://github.com/ipfs/go-ipfs/tree/master/docs/examples/go-ipfs-as-a-library for how to use go-ipfs as a daemon but from golang using the go-ipfs-api client, showcasing some cool unique daemon features. (I've this on my ideas/todo but you probably will get to this faster and do it better than I can :))

daviddias:
@hacdias move to Review/QA when you want our review
image

hacdias:
Thanks for the heads up @daviddias and yes, it's ready for review. Please note only the first two tests are fully implemented. All the others are missing implementation and tracking is being done on ipfs/interface-go-ipfs-core#93!

plans/chew-large-datasets/test/ipfs_mfs.go
```diff
@@ -0,0 +1,13 @@
+package test
+
+import (
+	"fmt"
+
+	"github.com/ipfs/testground/sdk/runtime"
+)
+
+func IpfsMfs(runenv *runtime.RunEnv) {
+	fmt.Printf("Yo - IpfsMfs")
```
daviddias:
```suggestion
	fmt.Printf("IpfsMfs - Not implemented yet")
```

plans/chew-large-datasets/test/ipfs_file_store.go
```diff
@@ -0,0 +1,53 @@
...
+		return nil
+	})
+
+	if err != nil {
+		runenv.Abort(err)
+		return
+	}
+
+	// TODO: Act II and Act III
+
```
daviddias:
```suggestion
    fmt.Printf("Test incomplete")

```

plans/chew-large-datasets/test/ipfs_mfs_dir_sharding.go
```diff
@@ -0,0 +1,13 @@
+package test
+
+import (
+	"fmt"
+
+	"github.com/ipfs/testground/sdk/runtime"
+)
+
+func IpfsMfsDirSharding(runenv *runtime.RunEnv) {
+	fmt.Printf("Yo - IpfsMfsDirSharing")
```
daviddias:
```suggestion
	fmt.Printf("IpfsMfsDirSharing - Not implemented yet")
```

plans/chew-large-datasets/test/ipfs_url_store.go
```diff
@@ -0,0 +1,27 @@
...
+	_, err := utils.CreateIpfsInstance(ctx, func(cfg *config.Config) error {
+		cfg.Experimental.UrlstoreEnabled = true
+		return nil
+	})
+	if err != nil {
+		panic(fmt.Errorf("failed to spawn ephemeral node: %s", err))
+	}
+
+	// TODO: ...
+
```
daviddias:
```suggestion
    fmt.Printf("Test incomplete")

```

plans/chew-large-datasets/test/ipfs_url_store.go
```diff
@@ -0,0 +1,27 @@
...
+	"context"
+	"fmt"
+
+	config "github.com/ipfs/go-ipfs-config"
+	utils "github.com/ipfs/testground/plans/chew-large-datasets/utils"
+	"github.com/ipfs/testground/sdk/runtime"
+)
+
+func IpfsUrlStore(runenv *runtime.RunEnv) {
+	fmt.Printf("Yo - IpfsUrlStore")
```
daviddias:
```suggestion
```

hacdias:
Are you sure about rebasing? This PR has 41 commits and its messages are not the best. I think it would be better to squash this ones

daviddias:
squashed :)

PR #60, opened 2019-10-03
Plan: smlbench 

daviddias:
Goal: get smlbench to work through testground


manifests/smlbench.toml
```diff
@@ -1,18 +1,27 @@
...
 enabled = true
 go_version = "1.13"
 module_path = "github.com/ipfs/testground/plans/dht"
+exec_pkg = "."
+go_ipfs_version = "0.4.22"
+
+# Builds latest go-ipfs
+[build_strategies."docker:ipfs"]
+enabled = true
+go_version = "1.13"
```
daviddias: ❓ I haven't been able to figure out what is the dark magic that makes these variables become available to the Docker file in an uppercase form

plans/smlbench/main.go
```diff
@@ -0,0 +1,81 @@
...
+		panic("test case sequence number not set")
+	}
+
+	testCases := testCasesSet
+
+	for i, tc := range testCases {
+		_ = os.Setenv("TEST_CASE", tc.Name())
+		_ = os.Setenv("TEST_CASE_SEQ", strconv.Itoa(i))
+
+		// ctx := api.NewContext(context.Background())
```
daviddias: @raulk What did you intend to do with this `api.NewContext`? I mean, where it is supposed to come from.

daviddias: It just dawned on me that probably all I need is to do this https://github.com/ipfs/go-ipfs/pull/6695/files#diff-5146146b33d1d2368a717a97d254547dR151. Will test in a sec

manifests/smlbench.toml
jimpick: For example, for go_version, in pkg/build/docker_go.go, you can see it is configured to look for "go_version" in the TOML file here:

https://github.com/ipfs/testground/blob/5284895c6c4bca284dc61e18c52a9f1094b711bf/pkg/build/docker_go.go#L34

I wasn't really familiar with the third element in the struct there ... those are an advanced feature called "tags" which let you add extra information that is available via reflection.

https://medium.com/golangspec/tags-in-golang-3e5db0b8ef3e

The tags are used by the TOML parser (executed by pkg/engine/engine.go) to map fields:

https://github.com/ipfs/testground/blob/5284895c6c4bca284dc61e18c52a9f1094b711bf/pkg/engine/engine.go#L245

It's an advanced feature of the TOML parser, documented here:

https://godoc.org/github.com/BurntSushi/toml#hdr-Key_mapping

Later, you can see how how it builds the GO_VERSION environment setting for Docker:

https://github.com/ipfs/testground/blob/5284895c6c4bca284dc61e18c52a9f1094b711bf/pkg/build/docker_go.go#L155

jimpick:
It's a little unclear on how the new version of the smlbench should interact with Docker...

The old version didn't use Docker, and simply used the wrapper around IPTB from the testground SDK + the version of IPFS found in the path.

There does appear to be an [IPTB plugin for Docker](https://github.com/ipfs/iptb-plugins/tree/master/docker) ... but I assume that's not what we want to use, as testground already spins up Docker instances (eg. you pass in the number of desired instances for the DHT test).

The [simple_add](https://github.com/ipfs/testground/blob/plan-smlbench/plans/smlbench/test/simple_add.go) test only runs on a single IPFS node (spawned by IPTB).

The [simple_add_get](https://github.com/ipfs/testground/blob/plan-smlbench/plans/smlbench/test/simple_add_get.go) test uses IPTB to spawn an "adder" node and a "getter" node.

The simplest thing would be to get IPTB built inside a single Docker image, and run both the "adder" and "getter" via IPTB in a single Docker container.

The [testground SPEC](https://github.com/ipfs/testground/blob/master/docs/SPEC.md#overview) mentions IPTB several times in order to explain concepts, but doesn't specifically say that we are using it.

https://github.com/ipfs/testground/blob/master/docs/SPEC.md#overview

For some future tests, we want to be able to use things like `tc` to do [traffic shaping](https://github.com/ipfs/testground/issues/29), which suggests that we want each node in a separate Docker container.

We could use an approach like the DHT tests, where the test binary is launched in separate Docker instances...

https://github.com/ipfs/testground/blob/master/plans/dht/test/lookup_peers.go

For the "simple_add_get" test, one possibility could be to run the test with the command-line argument `--instances=2`, and have each instance decide if it's the "adder" or the "getter". It looks like the Docker container name contains a [incrementing counter](https://github.com/ipfs/testground/blob/master/pkg/runner/local_docker.go#L156) for each instance - it might not be passed down via environment variables though to the `runenv`...

If we wanted to preserve the IPTB api for imperatively creating instances from within tests, but create them in separate Docker instances, that might be possible by making an IPTB plugin that communicated back to the 'engine'. I'm not sure if we want to do that or not.

jimpick:
The easiest thing just to get the test to pass is going to be to run the existing IPTB-based tests in a single Docker container ... I'll prototype that on a branch and link it here when it's working.

jimpick:
It wasn't the easiest thing to get working, but here it is:

https://github.com/ipfs/testground/tree/plan-smlbench-single-container

Try:

```
./testground -vv run --builder docker:go --runner local:docker --instances=1 smlbench/simple-add
```

or

```
./testground -vv run --builder docker:go --runner local:docker --instances=1 smlbench/simple-add-get
```

I had to do a lot of surgery to the Dockerfile to (optionally) download go-ipfs. I switched the base image from busybox to Debian because the prebuild go-ipfs needs a full glibc.

jimpick:
That's just a first stab at it ... I'm not entirely satisfied with how I modified the Dockerfile template to add IPFS to it. It probably makes more sense to use different templates depending on the test.

jimpick:
Another thought ... for published IPFS binaries, we could copy them out of an existing Docker image, which would cache a bit more nicely than downloading them each time.

jimpick:
I'm done for the night... so I'll leave this here to see if I'm on the right track...

I started another branch:

https://github.com/ipfs/testground/tree/plan-smlbench-two-containers

I made a "smlbench2" test suite, which is just "smlbench", but only with the simple_add_get test enabled. I'm going to try to spin it up in 2 different docker containers - and have one assume the "adder" role, and the other assume the "getter" role.

If I understand correctly, I can get a sequence number from redis by calling writer.Write() https://github.com/ipfs/testground/blob/944d3429bcf06137d0754ba3870ae35eb5504d55/sdk/sync/write.go#L109

If that works the way I hope it does, one peer can act as the "adder" and the other as the "getter". Once the peers are ready, I just need to connect them. Then I need to add the file on the adder (smlbench already has this), and signal the state, then get it on the "getter".

daviddias:
@jimpick lot's of progress, nice! Could you open a PR against this one so that I can see exactly the differences?

plans/smlbench/main.go
```diff
@@ -0,0 +1,82 @@
...
+
+	runenv := runtime.CurrentRunEnv()
+	if runenv.TestCaseSeq < 0 {
+		panic("test case sequence number not set")
+	}
+
+	testCases := testCasesSet[runenv.TestCaseSeq]
+
+	for i, tc := range testCases {
+		_ = os.Setenv("TEST_CASE", tc.Name())
```
daviddias: Hitting:
```
# github.com/ipfs/testground/plans/smlbench
./main.go:67:32: tc.Name undefined (type utils.SmallBenchmarksTestCase has no field or method Name)
```

Although the method is declared. Means that I'm missing some goism here. Will figure this out

jimpick: I added this line to add .Name() to the interface

https://github.com/ipfs/testground/blob/plan-smlbench-single-container/plans/smlbench/utils/testcase.go#L10

My guess is that when it is compiling in the container, it doesn't have the latest code... I wonder if it's pulling from github instead of copying the local source code into the docker image?

jimpick:
I've got the "2 container" version of the smlbench testplan (simple-add-get) partially working...

See #74 for all the details.

It's using the Testground Sync API to communicate between the peers. I'll have to tweak it a little bit to get it to work in Docker, but I'm really close.

daviddias:
@jimpick thank you so much for the help in landing this one :) With this and having done https://github.com/ipfs/go-ipfs/pull/6695/files, I now feel fully unblocked to land #58

PR #62, opened 2019-10-10
Document cloud setup on AWS

jimpick:
I'm in the middle of making 4 setup videos to guide new users (administrators, initially) through the process of setting up the base infrastructure for testground on AWS as it currently exists.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,6 @@ plans/**/go.sum
 .history
 
 # End of https://www.gitignore.io/api/go,visualstudiocode
+
+.vscode
+*.swp
--- a/docs/cloud-setup/index.md
+++ b/docs/cloud-setup/index.md
@@ -1,0 +1,30 @@
+Testground Cloud Setup
+======================
+
+Here are some preliminary instructions for getting started setting up a cloud infrastructure for running Testground.
+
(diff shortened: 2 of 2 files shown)


jimpick:
Is this a good space to drop things?

jimpick:
Just going to stick a gist for a working filebeat.yml here until I can document the setup a bit better:

https://gist.github.com/jimpick/fb3e5c841ae5d8c2436a5f8654bb0cf2

(You need to update the authentication to match your cluster - eg. cloud.id and cloud.auth for Elastic Cloud)

jimpick:
I'm going to try to do a bit more infra automation ... I have an AMI image built with Packer with the core dependencies that we can use for an AWS Autoscaling Group on EC2, but there are still a few manual setup steps for each VM in the cluster (eg. Filebeat Elasticsearch credentials, AWS ECR, networking, extra ssh keys, etc.). I'm thinking ansible is a good fit, although any similar tool would be fine. With that, setting up a fresh cluster should be pretty easy to document.

daviddias:
Now happening with #111. Can we close this one, @jimpick ?

jimpick:
Yep. Closing it.

PR #71, opened 2019-10-15
More clarifications about running locally

autonome:
@daviddias I've added some questions in here, if you can look!

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,18 +123,19 @@ smlbench/lookup-providers
 smlbench/store-get-value
 ```
 
-Then run this comment to... (TODO: What does this do?)
+This next command is your first test! It runs the lookup-peers test from the DHT plan, using the builder (which sets up the environment + compilation) named docker:go (which compiles go inside docker) and runs it using the runner local:docker (which runs on your local machine).
 
 ```
 > TESTGROUND_BASEDIR=`pwd` testground -vv run dht/lookup-peers --builder=docker:go --runner=local:docker --build-cfg bypass_cache=true
 ...
 ```
 
-You should see... (TODO: What should I see?)
+You should see a bunch of logs that describe the steps of the test, from:
 
-To check which Test Plan and Test Cases are available do:
-
-TODO: Is something supposed to be here?
+* Setting up the container
+* Compilation of the test case inside the container
+* Starting the containers (total of 50 as 50 is the default number of nodes for this test)
+* You will see the logs that describe each node connecting to the others and executing a kademlia find-peers action.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -121,15 +123,19 @@ smlbench/lookup-providers
 smlbench/store-get-value
 ```
 
-Then do
+Then run this comment to... (TODO: What does this do?)
```
daviddias: It runs the _lookup-peers_ test from the DHT plan, using the builder (setting up env + compilation) named docker:go (suggests that it compiles go inside docker) and runs it using the runner _local:docker_ (suggests that it is running in our local machine but locally).

README.md
```diff
@@ -121,15 +123,19 @@ smlbench/lookup-providers
...
 
 ```
 > TESTGROUND_BASEDIR=`pwd` testground -vv run dht/lookup-peers --builder=docker:go --runner=local:docker --build-cfg bypass_cache=true
 ...
 ```
 
+You should see... (TODO: What should I see?)
```
daviddias: You will see a bunch of logs that describe the steps of the test, from:

- Setting up the container
- Compilation of the test case inside the container
- Starting the containers (total of 50 as 50 is the default number of nodes for this test)
- You will see the logs that describe each node connecting to the others and executing a kademlia find-peers action.

daviddias: Do note that the Plans and Test cases are not final, so this might change meanwhile. We can keep updating it :)

README.md
```diff
@@ -121,15 +123,19 @@ smlbench/lookup-providers
...
+
 To check which Test Plan and Test Cases are available do:
 
+TODO: Is something supposed to be here?
```
daviddias: Ah! I moved it to the top and forgot to delete this sentence :)

README.md
```diff
@@ -121,15 +123,19 @@ smlbench/lookup-providers
...
+Then run this comment to... (TODO: What does this do?)
 
 ```
 > TESTGROUND_BASEDIR=`pwd` testground -vv run dht/lookup-peers --builder=docker:go --runner=local:docker --build-cfg bypass_cache=true
 ...
 ```
 
+You should see... (TODO: What should I see?)
+
 To check which Test Plan and Test Cases are available do:
```
daviddias:
```suggestion
```

README.md
daviddias:
```suggestion
```

README.md
daviddias: The command is indeed super verbose and definitely not easy to memorize. We haven't spent enough time on refining the user experience. It does need to happen though :)

PR #73, opened 2019-10-17
smlbench single container (for comparison)

jimpick:
For ease of comparison against PR #60 - not intended to be merged as is

This particular branch gets the smlbench test case working again ... however, it's only executing the old IPTB based test inside a single Docker instance, which is probably not what we ultimately want.

Run it with a single container, eg:

```
./testground -vv run --builder docker:go --runner local:docker --instances=1 smlbench/simple-add
```

or

```
./testground -vv run --builder docker:go --runner local:docker --instances=1 smlbench/simple-add-get
```

I modified the Dockerfile.template to call `install-ipfs.sh` if go_ipfs_version is set in the manifest. I also had to change the base image to debian from busybox in order to be compatible with the prebuilt IPFS binaries distributed from dist.ipfs.io. If I was to do it again, I think I'd prefer a mechanism where I could have a different Dockerfile for some test cases instead of sharing it between all test cases.


daviddias:
@jimpick mind resolving the merge conflicts for a clean merge/comparison? Thank you so much for pushing for this :)

PR #74, opened 2019-10-20
WIP: simplified smlbench - two containers (for comparison)

jimpick:
Don't merge ... for discussion

I took #73 and modified the test so it would run in 2 separate processes.

It uses the Testground Sync SDK (which uses Redis) to figure out which node is the getter and which is the setter. It uses the Sync SDK to communicate peer IDs, multiaddrs, state and to pass the CID.

testground-simple-add-get

It doesn't work in Docker yet (because it's using 127.0.0.1 for the Swarm addresses) ... I'll look into how to fix that next.


jimpick:
Looks like there is the possibility of passing in a 'swarmaddr' to IPTB, so we could set an IP address that is visible from the outside of the Docker container. 

https://github.com/ipfs/iptb-plugins/blob/0183b37641a0bf07aafb07dbb778060855cc3e74/local/localipfs.go#L80

jimpick:
I added some code to query the IPv4 address of eth0 and I pass that as the swarmaddr to IPTB. Now works between 2 Docker containers!

plans/smlbench/utils/testcase.go
```diff
@@ -1,11 +1,14 @@
-package smlbench
+package utils
```
daviddias: file changes will disappear on rebase from master

plans/smlbench/utils/metrics.go
```diff
@@ -1,4 +1,4 @@
-package smlbench
+package utils
```
daviddias: file changes will disappear on rebase from master

plans/smlbench/utils/files.go
```diff
@@ -1,4 +1,4 @@
-package smlbench
+package utils
```
daviddias: file changes will disappear on rebase from master

plans/smlbench/test/simple_add_get.go
```diff
@@ -1,4 +1,4 @@
-package cases
+package test
```
daviddias: file changes will disappear on rebase from master

plans/smlbench/test/simple_add.go
```diff
@@ -1,4 +1,4 @@
-package cases
+package test
```
daviddias: file changes will disappear on rebase from master

plans/smlbench/main.go
```diff
@@ -0,0 +1,77 @@
+package main
```
daviddias: file changes will disappear on rebase from master

plans/smlbench/doc.go
```diff
@@ -1,4 +1,3 @@
 // Package smlbench runs small benchmarks on IPFS.
```
daviddias: file changes will disappear on rebase from master

plans/smlbench2/main.go
```diff
@@ -0,0 +1,232 @@
...
+                } else {
+                        return time.Duration(t) * time.Second
+                }
+        }()
+
+	watcher, writer := sync.MustWatcherWriter(runenv)
+        defer watcher.Close()
+
+	spec := iptb.NewTestEnsembleSpec()
+	spec.AddNodesDefaultConfig(iptb.NodeOpts{Initialize: true, Start: true}, "local")
```
daviddias: Should this be "AddNodes" or "AddNode" (singular)? Given that it will only run a node per container?

plans/smlbench2/main.go
```diff
@@ -0,0 +1,232 @@
...
+	for i, addr := range swarmAddrs {
+		multiAddr, err := ma.NewMultiaddr(addr)
+		if err != nil {
+			runenv.Abort(err)
+		}
+		addrs[i] = multiAddr
+	}
+	addrInfo := &peer.AddrInfo{ID, addrs}
+
+	seq, err := writer.Write(sync.PeerSubtree, addrInfo)
```
daviddias: So, it gets the its sequence number when it does the first write because Redis returns the value of an incremental counter as peers appear?

plans/smlbench2/main.go
```diff
@@ -0,0 +1,232 @@
...
+	if err != nil {
+                runenv.Abort(err)
+        }
+        defer writer.Close()
+
+	client := localNode.Client()
```
daviddias: There is never a "remote" in this test as it always spawns a node for itself

plans/smlbench2/main.go
```diff
@@ -0,0 +1,232 @@
...
+		}
+		fmt.Println("cid", hash)
+
+		c, err := cid.Parse(hash)
+		if err != nil {
+			runenv.Abort(err)
+			return
+		}
+
+		runenv.EmitMetric(utils.MetricTimeToAdd, float64(time.Now().Sub(tstarted)/time.Millisecond))
```
daviddias: Note that for this Plan. The time adding the file isn't important. Only the time transferring it. (+ the other things to measure, such as number of repeated blocks, etc as described in the spec for the plan)

plans/smlbench2/test/simple_add_get.go
```diff
@@ -0,0 +1,77 @@
+package test
```
daviddias: It seams that this file is just c&p from smlbench as the test for smlbench2 is on main.go

plans/smlbench2/utils/files.go
```diff
@@ -0,0 +1,49 @@
...
+	"os"
+
+	"github.com/ipfs/testground/sdk/runtime"
+)
+
+// TempRandFile creates a file of the specified size (in bytes) within the
+// specified directory.
+//
+// It is the callers responsibility to delete this file when done.
+func TempRandFile(runenv *runtime.RunEnv, dir string, size int64) *os.File {
```
daviddias: This util will probably end up on the sdk as I'm sure it will be used many times across many plans :)

manifests/smlbench2.toml
```diff
@@ -0,0 +1,22 @@
...
+exec_pkg = "."
+go_ipfs_version = "0.4.22"
+
+[run_strategies."local:docker"]
+enabled = true
+
+# seq 0
+[[testcases]]
+name = "simple-add-get"
+instances = { min = 1, max = 1, default = 1 }
```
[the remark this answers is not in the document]
daviddias: Going through the code and your update. I read that you used 2 docker containers, however, I don't see anywhere where that instruction happens. I'm guessing you pass --instances=2 when running and it simply doesn't error and does indeed spawn 2 instances, is that the case?

manifests/smlbench.toml
```diff
@@ -1,51 +1,52 @@
 name = "smlbench"
```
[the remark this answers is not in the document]
jimpick: My original thinking was that this might go into the smlbench test - although I stripped out the bit where it runs multiple iterations of loading different file sizes. It's not clear to me if that should be handled internally in the test (eg. spin up the peers first, then loop and try the different file sizes) or if it might be better to loop over the various file sizes at the test runner level, and have completely independent tests. We probably want to support both patterns.

But, as you say, this will be going into "plan 2"... I can easily just rename this and we can backport what we want into smlbench later. I think smlbench (or something similar) can serve as a simpler test case to show to new developers that want to develop a new test case.

manifests/smlbench2.toml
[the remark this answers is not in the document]
jimpick: Yes, I pass --instances=2. Maybe I don't need that if the toml file is correct? I'll try that out.

pkg/build/Dockerfile.template
```diff
@@ -34,12 +39,21 @@ RUN cd ${PLAN_DIR} \
     && go env -w GOPROXY=direct \
     && CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o testplan ${TESTPLAN_EXEC_PKG}
 
+# Optionally install IPFS
+RUN if [ -n "${GO_IPFS_VERSION}" ]; then echo Install IPFS ${GO_IPFS_VERSION}; bash /install-ipfs.sh ${GO_IPFS_VERSION}; fi
```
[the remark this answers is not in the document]
jimpick: That was my first attempt ... I was trying to keep a single Dockerfile.template that worked for both the DHT test and the smlbench test - Dockerfile doesn't support conditionals like "if" statements, so I needed to have a shell script for that. I'm not really happy with how it turned out, so I think it might make more sense to have multiple templates, or a template that is generated dynamically by testground.

plans/smlbench2/main.go
jimpick: Correct.

plans/smlbench2/main.go
jimpick: In the original test, there's an IPTB node called "adder" and another one called "getter". In this one, the node gets spawned before it knows what it's role is going to be, so I just called it "localNode" as it's local to the Docker container ... but it could be named anything. Maybe just calling it "node" is good? Naming things is hard.

plans/smlbench2/main.go
jimpick: I just copied that from the original smlbench test. I'm not even really sure yet what schema for the JSON is going to work well with graphing things in Kibana - I think we need to try to build some reports/graphing there really soon.

I'd love to dump as many metrics, events, traces as we possibly can, but tag things in an understandable way so we can easily generate reports with the most important metrics that we care about. It should be clear in the test what the primary metrics are vs. secondary metrics.

plans/smlbench2/test/simple_add_get.go
jimpick: Oh yeah, I was going to delete that file, I just forgot to. I was copying/pasting from it.

plans/smlbench2/utils/files.go
jimpick: Once we write a few more tests, I think it's going to become clear what needs to go into the SDK. Already, I can see a lot of things that don't have to be boilerplate.

daviddias:
@jimpick when you refactor and rename this smlbench2 plan, please do so on top of this folder https://github.com/ipfs/testground/tree/master/plans/data-transfer-datasets-random that has the Test Plan that you are implementing.

jimpick:
@daviddias Sounds great. I should be starting that very soon.

jimpick:
I got the 2 container test to work across Docker containers running on two different EC2 machines!

The setup was:

* setup Redis on the first machine and export the REDIS_HOST environment variable with the IP address
* build the testground Docker image for the test, and manually push it to an AWS ECR Docker registry
* manually create a Docker network on each machine with a different subnet
* create GRE tunnel between the two machines, with each end of the tunnel terminated in the Docker bridge subnets https://www.tldp.org/HOWTO/Adv-Routing-HOWTO/lartc.tunnel.gre.html
* added routing on each EC2 machine for the subnets
* add a high-priority ACCEPT rule in the iptables FORWARD table for the tunnel
* match the IP range in the iptables masquerade tables to prevent NAT
* small [shell script](https://github.com/jimpick/testground-misc-scripts/blob/master/run-manual.sh) on each machine to invoke the test

Here's the demo GIF showing it working - transferring 100MB between Docker instances on two different EC2 machines:

testground-two-ec2-vms-bridged-gre

jimpick:
I wrote up an explanation of how the test works yesterday. I'll link it from here so we can find it again.

https://github.com/ipfs/testground/pull/78#discussion_r337809993

plans/smlbench2/main.go
```diff
@@ -0,0 +1,232 @@
...
+			fmt.Println("Jim cid", c)
+			cancel()
+			// Get the content from the adder node
+			tstarted := time.Now()
+			err = getter.Get(c.String(), ensemble.TempDir())
+			if err != nil {
+				runenv.Abort(err)
+				return
+			}
+			runenv.EmitMetric(utils.MetricTimeToGet, float64(time.Now().Sub(tstarted)/time.Millisecond))
```
daviddias: Probably move the metrics to its own package so that it is `metrics.TimeToGet`

daviddias:
Once #126 is merged, let's get this one to work with the cluster:runner as well

daviddias:
Alright, https://github.com/ipfs/testground/pull/126 has been merged.

## Next steps for this PR:

- [ ] Rebase master onto this branch
- [ ] Rename smlbench2 to data-transfer-of-random-datasets
- [ ] Make this test work with the Docker Swarm cluster in AWS

daviddias:
@jimpick anything left to be done in this PR?

jimpick:
Previously you suggested using this as a base for: https://github.com/ipfs/testground/tree/master/plans/data-transfer-datasets-random

But it's also completely different than the test plan described in the README... not sure how things converge.

I'll go ahead and prepare the PR to apply it against that test plan as requested.

Also, now that the sidecar-based traffic shaping is working, that would go really nicely with this.

jimpick:
I've renamed the test and squashed/rebased to master, but it's failing when I run it:

```
./testground --vv run --builder docker:go --runner local:docker --build-cfg bypass_cache=true data-transfer-datasets-random/simple-add-get
```

```
"swarm/connect: connect QmRqBzsk6MWGxKp1T7uAfpvie2T4aN9XaLg8wAcWzK4c3V failure: failed to dial : all dials failed\n  * [/ip4/172.20.0.7/tcp/46459] dial tcp4 172.20.0.7:46459: connect: network is unreachable"
```

(possibly related to sidecar?)

The branch modifies the builder and Dockerfile under pkg/build/golang in an unclean way ... I think it needs a different approach.

daviddias:
AFAIU, we won't be using the code present in this PR any longer. Closing it down, reopen if necessary.

jimpick:
I still need this for reference purposes, so I'm keeping the branch alive over at: https://github.com/jimpick/testground/tree/smlbench2

PR #78, opened 2019-10-22
Weekly Sync Notes 2019-10-22

daviddias:


_meeting-notes/2019-10-22.md
```diff
@@ -0,0 +1,128 @@
...
+    - follow-on work, two dockers containers, IPTB with 1 IPFS node per container, using Sync SDK (redis) to co-ordinate https://github.com/ipfs/testground/pull/74
+  - Lots of little puzzles solved:
+    - IPFS into Docker using Docker.template and conditional via shell script (works, but not satisfied)
+    - IPTB SDK usage
+    - Sync SDK usage
+  - Tested on a single EC2 vm ... going to test on multiple VMs today
+
+## :question: Questions (Blockers, Technical Design needs)
+
+- (@jimpick) - Error handling patterns - which ones?
```
daviddias: https://trello.com/c/dhedtclq/67-create-document-that-explains-the-conventions-around-error-checking-and-logging-for-the-project

_meeting-notes/2019-10-22.md
```diff
@@ -0,0 +1,128 @@
...
+### Top 3 things to get done this week
+
+- Finalize the spec for the 10 Test Plans
+- Land [Plan: Data Exchange with Datasets of Interest (BitSwap/GraphSync)](https://github.com/ipfs/testground/blob/master/docs/tests-mvp.md#plan-data-exchange-with-datasets-of-interest-bitswapgraphsync) end-to-end (Plan, Tests & Graphs) so that we can onboard Dirk that needs it for testing Bitswap changes
+- ~~Rope in Adin by onboarding him to the Network Connectivity Test Plans (needs the spec to be finalized and an example using TC done)~~ Given the limited amount of time, we might not be able to do it this week and only on the next.
+
+### Assignments
+
+- ==@stebalien==
+  - `P0` Reach out to Dirk to see if he can review the Test Plan https://github.com/ipfs/testground/tree/master/plans/data-transfer-datasets-random
```
daviddias: @Stebalien updated https://trello.com/c/XINMtY3S/29-what-are-the-exact-change-sets-0422v050-that-wed-like-to-test

_meeting-notes/2019-10-22.md
```diff
@@ -0,0 +1,128 @@
...
+  - Review all Test Plan https://github.com/ipfs/testground/blob/master/docs/PLAN_SPEC.md
```
daviddias: Same as above  https://trello.com/c/XINMtY3S/29-what-are-the-exact-change-sets-0422v050-that-wed-like-to-test

daviddias:
Updated Trello, merging these! :)

_meeting-notes/2019-10-22.md
```diff
@@ -0,0 +1,128 @@
...
+  - Created a draft Roadmap for review https://github.com/ipfs/testground/pulls
+  - Got sick, then got better :) ETOOMUCHLONGFLIGHTS
+  - Broke a tooth 🦷 (tip: watch out for those Swiss Ricolas!). **FYI** will be traveling this week to Lisbon for a emergency dentist appt (and a reschedule of others I was going to reschedule to post hack week, but this way I merge it all together rather than going to Lisbon 2x). This means that I'm almost at 2 days of lost work given flying + transportation + being at the dentist.
+- ==@jimpick==
+  - smlbench (adds a randomly generated file on one node, gets it from a 2nd node)
+    - initial work, one docker container, IPTB with 2 IPFS nodes https://github.com/ipfs/testground/pull/74
+    - follow-on work, two dockers containers, IPTB with 1 IPFS node per container, using Sync SDK (redis) to co-ordinate https://github.com/ipfs/testground/pull/74
+  - Lots of little puzzles solved:
+    - IPFS into Docker using Docker.template and conditional via shell script (works, but not satisfied)
+    - IPTB SDK usage
```
daviddias: What's this @jimpick ?

_meeting-notes/2019-10-22.md
```diff
@@ -0,0 +1,128 @@
...
+    - Sync SDK usage
```
daviddias: What did you learn @jimpick ?

_meeting-notes/2019-10-22.md
```diff
@@ -0,0 +1,128 @@
...
+  - Tested on a single EC2 vm ... going to test on multiple VMs today
```
daviddias: @jimpick when you say you tested in EC2. Did you spin a EC2 VM from your local machine or did you install TestGround on EC2 Vm and ran it there?

_meeting-notes/2019-10-22.md
jimpick: The test plan uses IPTB via Testground SDK documented here:

https://godoc.org/github.com/ipfs/testground/sdk/iptb

Rather than spin up a whole bunch of nodes, we're only using IPTB to spin up a single node (per Docker container) that we can control using the go-ipfs-api documented here:

https://godoc.org/github.com/ipfs/go-ipfs-api

Since all that infrastructure already existed as part of IPTB, it made sense to re-use it.

I think the "go-ipfs as a library" approach is also of high value - where we build IPFS directly into the test instead of calling an existing binary - that's probably useful for when we need deeper control or need to test a specific subset of things. https://github.com/ipfs/go-ipfs/pull/6695/

_meeting-notes/2019-10-22.md
jimpick: Sync SDK is here:

https://godoc.org/github.com/ipfs/testground/sdk/sync

It's the code that communicates between Docker containers using Redis.

The first thing I do with it is register each peer using `writer.Write()` ... it expects a peer.AddrInfo structure with the Peer ID and Multiaddrs - because I'm writing to the sync.PeerSubtree location, that enforces the type check and generates the prefix for the Redis key, eg. "run::plan::case:simple-add-get:nodes:..."

https://github.com/ipfs/testground/blob/f3746e82ce5ed06f93632bcb502d94e216c2ce3a/plans/smlbench2/main.go#L86

When I write to the sync API, it returns a sequence number. I have 2 peers running in separate Docker containers ... if the sequence number returns "1", then I make that peer act as the *"adder"* node. If the sequence is "2", then I make that peer into the *"getter"* node.

# The *"adder"* node goes first

It doesn't have to wait for anything, so it just makes some random data and adds it to the IPFS node.

We need to write the CID of the random data into Redis so the *"getter"* node can figure out what to fetch. We do that with `writer.Write()`:

https://github.com/ipfs/testground/blob/f3746e82ce5ed06f93632bcb502d94e216c2ce3a/plans/smlbench2/main.go#L123

I define a tiny namespace in Redis for it to store the state here:

https://github.com/ipfs/testground/blob/f3746e82ce5ed06f93632bcb502d94e216c2ce3a/plans/smlbench2/main.go#L25-L31

That will generate a key prefix in Redis that looks like `run::plan::case:simple-add-get:cid:...` and enforces the type and does the marshalling/unmarshalling for that automatically between when it is written from one Docker instance and read from another.

Once the CID is written, you can see that the "adder" node signals that it has reached the **"added"** state here:

https://github.com/ipfs/testground/blob/f3746e82ce5ed06f93632bcb502d94e216c2ce3a/plans/smlbench2/main.go#L129

At this point, it can't just exit, because the IPFS node needs to stay running so the *"getter"* node can fetch data from it. So it waits until the *"getter"* node signals that it has reached the **"received"** state using `watcher.Barrier()`:

https://github.com/ipfs/testground/blob/f3746e82ce5ed06f93632bcb502d94e216c2ce3a/plans/smlbench2/main.go#L136

# The *"getter"* node goes next

Once the *"getter"* node knows that it is the 2nd node, it has to get the multiaddr of the *"adder"* node so it can use IPFS to connect to it.

It uses the `watcher.Subscribe()` SDK method to get a channel of updates for the peers that have added their information to Redis. Since the *"adder"* node has already added data (even if it was added before the second node was even started) - the SDK will fetch that from Redis and immediately output an event on the channel with the contents of the peer.Addrinfo data structure ... all nicely unmarshalled and typed! (very nice DX @raulk, a joy to use).

https://github.com/ipfs/testground/blob/f3746e82ce5ed06f93632bcb502d94e216c2ce3a/plans/smlbench2/main.go#L159

With that, it's easy to call the the IPFS API to connect to the peer.

Once connected, the *"getter"* node has to wait for the *"adder"* node to reach the **"added"** state. That's done by setting a "barrier" using `watcher.Barrier()`

https://github.com/ipfs/testground/blob/f3746e82ce5ed06f93632bcb502d94e216c2ce3a/plans/smlbench2/main.go#L188

Once it sees the **"added"** state has been reached, it fetches the CID using `watcher.Subscriber()` again, but this time looking at the Redis namespace set aside for the CID (instead of the AddrInfo space as before):

https://github.com/ipfs/testground/blob/f3746e82ce5ed06f93632bcb502d94e216c2ce3a/plans/smlbench2/main.go#L197

The *"getter"* node now has all the info it needs, so it can just call the IPFS API to fetch the content and time how long it takes.

The last thing the *"getter"* node does is signal that it has reached the **"received"** state.

https://github.com/ipfs/testground/blob/f3746e82ce5ed06f93632bcb502d94e216c2ce3a/plans/smlbench2/main.go#L221

It just exits at this point because it is done.

# The *"adder"* node goes last

The adder node has been waiting for the "received" state using `watcher.Barrier()` ... now that the **"received"** state has been reached, there's nothing more to do, so it can safely exit (shutting down its IPFS node).

_meeting-notes/2019-10-22.md
jimpick: I just manually created an EC2 VM in the AWS web console and then I ssh into it, checkout out testground, and run the test.

I used a custom AMI image generated using Packer with this template: https://github.com/jimpick/testground-packer-templates

That pre-installs Docker, redis, Filebeat etc.

I've also got a few throwaway shell scripts here that might be useful...

https://github.com/jimpick/testground-misc-scripts

(shortcuts for clearing redis, cleaning up docker containers, images and networks, running various stages, no docs)

daviddias: Got it! Makes sense. How far (or are we even on track) to get the ability to spin the instances on AWS from our local machines through TestGround?

jimpick: I think that's totally possible with a small amount of extra work.

Right now, the existing local_docker runner assumes that it is "local".

I created a Trello card describing a possible ssh-based runner that captures the experimentation I've done so far - I was assuming that would run inside the first EC2 machine in the cluster, but there's no reason it couldn't run on a laptop, etc. When we implement an automated orchestrator that does CI, that will necessarily have to run in the cloud.

Currently, it's pretty easy to go into AWS directly, create an EC2 instance (ideally inside a scaling group) right from the AWS console, then install testground, configure Filebeat, and use it. I'll get some better tutorial documentation written very soon.

_meeting-notes/2019-10-22.md
raulk: Hey @jimpick, it would be great to add these explanations on the godocs. They are just not very discoverable here. Mind taking a quick stab at that?

PR #86, opened 2019-10-29
implement the exec:go builder, and the local:exec runner

raulk:
This PR introduces a new builder and a new runner:

* `exec:go` builder: builds a Go test plan with the system's Go toolchain, and produces an executable that conforms to the [Test Case Runtime Environment](https://github.com/ipfs/testground/blob/master/docs/SPEC.md#test-case-runtime-environment), as defined in the spec.
* `local:exec` runner: runs a test case in a test plan that has been built with an `exec:*` builder, forking as many system processes as instances required.

These components vastly improve DX, by allowing a test plan developer to iterate rapidly by avoiding Docker builds and runtime overhead.

Test it like this, after checking out this branch. ~Make sure you’re running a local Redis instance on port 6379 (default).~ (the runner now starts a temporary redis instance automatically).

```sh
$ go build .
$ ./testground -vv run --builder exec:go --runner local:exec --instances=20 --build-cfg bypass_cache=true --build-cfg go_version=\"1.13\" dht/lookup-peers
```


raulk:
Merged this as it's good to go and it works.

PR #87, opened 2019-10-29
WIP: Create Test Plan Spec :Data transfer with variable connectivity

dirkmc:


daviddias:
@dirkmc what about this Plan becoming part of the https://github.com/ipfs/testground/tree/master/plans/bitswap-tuning (your plan) as a test case?

dirkmc:
I think that's a good idea 👍 

The test plan I've written specifically tests Bitswap. In future we may want to have test plans that more generally test data transfer, be it Bitswap, Graphsync etc. We can cross that bridge when we come to it though

daviddias:
Thank you for the fast response! Left you a note at https://github.com/ipfs/testground/issues/100#issuecomment-564597466 so that the intent is not dropped. 

PSA: we are working towards making progress on Sidecar, hopefully by next Monday you will have what you need to tinker with latency and bandwidth for your bitswap tuning plan.

PR #91, opened 2019-10-30
flesh out a naming test plan

Stebalien:
I've merged them all into the same test as the _test_ is the same for each router (and we usually use multiple routers in parallel).


aschmahmann:
@Stebalien @daviddias it may not be that important, but should we also be adding a test for bandwidth usage in addition to latency?

Stebalien:
@aschmahmann yes.

Issue #110, opened 2019-11-01
[WIP] Preliminary cluster:swarm runner


jimpick:
Work-in-progress...

I have an automated Ansible setup for Docker swarm in the aws-ansible branch (needs just a bit more cleanup). I first tested it with a shell script, but then I realized that it wouldn't be too much work to modify the local:docker runner to talk to the swarm API.

Here's a demo running the smlbench2 2-container bitswap test across 2 different EC2 vms in the Docker swarm cluster:

![tg-docker-swarm-runner](https://user-images.githubusercontent.com/20886/68005983-553beb80-fc34-11e9-801c-eaa0f046033e.gif)

https://github.com/ipfs/testground/blob/aws-ansible/pkg/runner/cluster_swarm.go

It's not quite done yet ... I've hard-coded the image (pulling from ECR), the REDIS_HOST environment variable, and the network. We'll want to upload the image to ECR after building it before creating the service.

daviddias:
PR with WIP at https://github.com/ipfs/testground/pull/112 Continue the conversation there.

Issue #94, opened 2019-10-30
Impl Plan 2 - Data Transfer of Random DataSets (Bitswap/GraphSync)


daviddias:
Plan Description: https://github.com/ipfs/testground/tree/master/plans/data-transfer-datasets-random

Import from Trello:
  - Trello Card Import - [Impl Plan 2 - Data Transfer of Random DataSets (Bitswap:GraphSync) on Testground: Ignition 🚀 | Trello.pdf](https://github.com/ipfs/testground/files/3787379/Impl.Plan.2.-.Data.Transfer.of.Random.DataSets.Bitswap.GraphSync.on.Testground.Ignition.Trello.pdf)
  - Trello Card URL https://trello.com/c/6kEQv6Xn/36-impl-plan-2-data-transfer-of-random-datasets-bitswap-graphsync

daviddias:
Coalescing with https://github.com/ipfs/testground/issues/95 at https://github.com/ipfs/testground/pull/242 as suggested https://github.com/ipfs/testground/issues/83

PR #112, opened 2019-11-01
WIP: cluster:swarm runner

jimpick:
I have an automated Ansible setup for Docker swarm in the aws-ansible branch (needs just a bit more cleanup). I first tested it with a shell script, but then I realized that it wouldn't be too much work to modify the local:docker runner to talk to the swarm API.

Here's a demo running the smlbench2 2-container bitswap test across 2 different EC2 vms in the Docker swarm cluster:

tg-docker-swarm-runner

https://github.com/ipfs/testground/blob/aws-ansible/pkg/runner/cluster_swarm.go

It's not quite done yet ... I've hard-coded the image (pulling from ECR), the REDIS_HOST environment variable, and the network. We'll want to upload the image to ECR after building it before creating the service.

-------------- 

Not complete yet ... some hard-coded values. See demo of it running in #110 (along with notes about what is still missing)

I just hacked this together to test the ansible setup I've been prototyping in the aws-ansible branch:

https://github.com/ipfs/testground/tree/aws-ansible

(tested with the smlbench2 test, which isn't on master yet ... that's going to be the basis of #94)

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -3,6 +3,8 @@ package engine
 import (
 	"bytes"
 	"fmt"
+	"os"
+	"path/filepath"
 	"reflect"
 	"sync"
 
@@ -50,6 +52,7 @@ type Engine struct {
 		src  string
 		work string
 	}
+	env *api.EnvConfig
 }
 
 type EngineConfig struct {
@@ -84,6 +87,16 @@ func NewEngine(cfg *EngineConfig) (*Engine, error) {
 
 	e.dirs.src = srcdir
 	e.dirs.work = workdir
+
+	// load the .env.toml file.
+	e.env = new(api.EnvConfig)
+	path := filepath.Join(srcdir, ".env.toml")
+	switch _, err := toml.DecodeFile(path, e.env); err.(type) {
+	case nil:
+		fmt.Println("loading env from", path)
+	case *os.PathError:
+		// TODO no .env.toml file located, maybe print a warning.
+	}
 
 	return e, nil
 }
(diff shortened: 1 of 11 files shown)


pkg/engine/engine.go
```diff
@@ -23,6 +23,7 @@ var AllBuilders = []build.Builder{
 var AllRunners = []runner.Runner{
 	&runner.LocalDockerRunner{},
 	&runner.LocalExecutableRunner{},
+	&runner.ClusterSwarmRunner{},
```
daviddias:
```suggestion
	&runner.ClusterDockerSwarmRunner{},
```

daviddias: I'm not entirely sure if naming yet another Runner just for Docker Swarm is the best path. It probably worth to just have Docker Swarm as the default Docker Runner

daviddias: Or perhaps this is me just getting confused by seeing Docker Swarm being called a Runner (given that in the end it is just the same thing running in a Docker container, but managed by Docker Swarm)

pkg/runner/cluster_swarm.go
```diff
@@ -1 +1,162 @@
...
+	// Start service
+	// name := fmt.Sprintf("tg-%s-%s-%s", input.TestPlan.Name, testcase.Name, input.ID)
+	name := fmt.Sprintf("tg-%s-%s", input.TestPlan.Name, testcase.Name)
+
+	log.Infow("creating service", "name", name)
+
+  // docker service create --replicas 1 --name helloworld alpine ping docker.com
+
+  replicas := uint64(input.Instances)
+  spec := swarm.ServiceSpec{
```
daviddias: Can I get a clarification of this way of specing the service vs. using docker-compose.yml? (https://docs.docker.com/compose/)

pkg/engine/engine.go
raulk: I think calling this `ClusterSwarmRunner` is fine. I was using the convention `scope:technology`, e.g. `local:exec`, `local:docker`, etc. to name runners. We can also rename "runners" to "schedulers" if we feel that's more accurate.

pkg/runner/cluster_swarm.go
raulk: Now that we've decided to create a temporary overlay network per run, we _could_ use Docker Compose to create both the network and the service in one go. However, IIRC Docker Compose is really a CLI tool written in Python that transforms a YAML into docker calls. We cannot use it from Golang :-(

raulk:
Taking this to land it on top of @jimpick's work on the infra side of things.

raulk:
Closing this PR in favour of #126. Thanks for the head start here, @jimpick! Let's land it!

PR #127, opened 2019-11-06
Fix locateSrcDir

lanzafame:
Running `testground` in $GOPATH/src/github.com/ipfs/testground returned:
```
$ testground
attempting to guess testground base directory; for better control set ${TESTGROUND_SRCDIR}
attempting to guess testground base directory; for better control set ${TESTGROUND_SRCDIR}
attempting to guess testground base directory; for better control set ${TESTGROUND_SRCDIR}
attempting to guess testground base directory; for better control set ${TESTGROUND_SRCDIR}
attempting to guess testground base directory; for better control set ${TESTGROUND_SRCDIR}
attempting to guess testground base directory; for better control set ${TESTGROUND_SRCDIR}
panic: unable to determine testground src dir

goroutine 1 [running]:
github.com/ipfs/testground/cmd.glob..func1(0x0)
	/home/lanzafame/src/github.com/ipfs/testground/cmd/all.go:12 +0x5b
github.com/ipfs/testground/cmd.init()
	/home/lanzafame/src/github.com/ipfs/testground/cmd/all.go:15 +0x37
```

Discovered that `locateSrcDir` was getting the directory of the executable, instead of the working directory, i.e. where the executable was called, and it was also truncating the path before checking if it was the source directory.

### DIFF
--- a/pkg/engine/dirs.go
+++ b/pkg/engine/dirs.go
@@ -46,7 +46,6 @@ func locateSrcDir() (string, error) {
 
 	for len(path) > 1 {
 		fmt.Printf("attempting to guess testground base directory; for better control set ${%s}\n", EnvTestgroundSrcDir)
-		fmt.Println(path)
 		if isTestgroundRepo(path) {
 			os.Setenv(EnvTestgroundSrcDir, path)
 			fmt.Printf("successfully located testground base directory: %s\n", path)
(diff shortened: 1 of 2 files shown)


lanzafame:
Closing as I have misunderstood the use of `os.Executable` and my fix only works if you are running `testground` in the source directory, which I was because I had just pulled it down. The reason that the original code didn't work though was because I had run `go install` instead of `go build` therefore putting the binary in $GOPATH/bin instead of $GOPATH/src/github.com/ipfs/testground.

raulk:
No, I think you're right! We can improve this logic, and also document it in the README.

raulk:
Did so in #130.

PR #116, opened 2019-11-04
Impl Plan DHT Part I (Plan Spec + Test Find Peers)

daviddias:
WIP PR to Match #115

PR #129, opened 2019-11-06
Impl Plan DHT - Part deux (Find Providers + Provide Stress + Factoring out the network set up logic)

daviddias:
Continues the work started on #116


plans/dht/test/find_providers.go
```diff
@@ -2,19 +2,100 @@ package test
...
+	for i := 0; i < nCidsToNotPublish; i++ {
+		v := fmt.Sprintf("%d -- not gonna publish", i)
+		mhv := ipfsUtil.Hash([]byte(v))
+		cidsToNotPublish = append(cidsToNotPublish, cid.NewCidV0(mhv))
+	}
+
+	for i := 0; i < nCidsToPublish; i++ {
+		err := dht.Provide(ctx, cidsToPublish[i], true)
+		if err != nil {
+			runenv.Abort(fmt.Errorf("Failed on .Provide - %w", err))
```
daviddias: I'm always getting this error with:

```
<< 83046fff3822e0002d612cae80e53b0222c2ffb3127b193ff49e610503db556f >> {"runenv":{"test_plan":"dht","test_case":"find-providers","test_run":"aea7286e-07f5-4255-a19a-40b9319fd6f8","test_seq":1,"test_instance_count":5},"timestamp":1573212021273521900,"result":{"outcome":"aborted","reason":"Failed on .Provide - no known addresses for self. cannot put provider."}}
```

Which comes from https://github.com/libp2p/go-libp2p-kad-dht/blob/master/routing.go#L466-L471

Does https://github.com/libp2p/go-libp2p-kad-dht/blob/master/routing.go#L466-L471 that now by default, the DHT doesn't work on LANs?

daviddias: nvm. Was calling context cancel in the wrong place

plans/dht/test/set-up.go
```diff
@@ -0,0 +1,116 @@
...
+		return nil, nil, nil, err
+	}
+
+	watcher, writer := sync.MustWatcherWriter(runenv)
+	defer watcher.Close()
+	defer writer.Close()
+
+	/// --- Tear down
+
+	defer func() {
```
Stebalien: Won't this end the entire _test_ when we exit this function?

daviddias: Still needs to wait for the barrier to signal that it is complete.

Stebalien: Previously, this barrier would happen after the _test_ completes. Now, it happens after _setup_ completes.

daviddias: You are right. Fixing it

daviddias: fixed

plans/dht/test/find_peers.go
```diff
@@ -169,5 +74,6 @@ Loop:
 
 	/// --- Ending the test
 
+	defer TearDown(ctx, runenv, watcher, writer)
```
Stebalien: Given that we're defering this, we should run it immediately after setup.

go.mod
```diff
@@ -15,16 +15,15 @@ require (
 	github.com/docker/docker v1.4.2-0.20190910181529-415f8ecb65e8
 	github.com/docker/go-connections v0.4.0 // indirect
 	github.com/docker/go-units v0.4.0 // indirect
-	github.com/gogo/protobuf v1.3.0 // indirect
 	github.com/google/uuid v1.1.1
 	github.com/gorilla/mux v1.7.3 // indirect
 	github.com/hashicorp/go-cleanhttp v0.5.1 // indirect
 	github.com/hashicorp/go-getter v1.4.0
 	github.com/imdario/mergo v0.3.7
+	github.com/ipfs/testground/plans/dht v0.0.0-20191106194925-0d10fef12a43 // indirect
```
raulk: These should not be necessary. I quite haven't identified why the go toolchain adds them some time. But you should try to remove them and do a `go build . && go mod tidy`, and they shouldn't reappear.

plans/dht/test/find_providers.go
```diff
@@ -1,20 +1,110 @@
...
+		cidsToPublish = append(cidsToPublish, cid.NewCidV0(mhv))
+	}
+
+	for i := 0; i < nCidsToNotPublish; i++ {
+		v := fmt.Sprintf("%d -- not gonna publish", i)
+		mhv := ipfsUtil.Hash([]byte(v))
+		cidsToNotPublish = append(cidsToNotPublish, cid.NewCidV0(mhv))
+	}
+
+	for i := 0; i < nCidsToPublish; i++ {
```
Stebalien: nit: one typically uses `for _, c := range cidsToPublish { ...`.

plans/dht/test/find_providers.go
```diff
@@ -1,20 +1,110 @@
...
+		}
+	}
+
+	runenv.Message("Provided a bunch of CIDs")
+
 	/// --- Act II
 
+	// TODO, only `p-resolving` of the nodes attempt to resolve the records provided before
+
+	for i := 0; i < nCidsToPublish; i++ {
```
Stebalien: nit: one typically uses `for _, c := range cidsToPublish { ...`.

plans/dht/test/find_providers.go
```diff
@@ -1,20 +1,110 @@
...
+			runenv.Abort(fmt.Errorf("Should have found Providers - %w", err))
+			return
+		}
+	}
+
+	runenv.Message("Found a ton of providers for the CIDs I was looking for")
+
+	// TODO, only `p-failing` of the nodes attempt to resolve records that do not exist
+
+	for i := 0; i < nCidsToNotPublish; i++ {
```
Stebalien: nit: one typically uses `for _, c := range cidsToNotPublish { ...`.

plans/dht/test/create-dht-node.go
```diff
@@ -1,25 +1,19 @@
...
 
 // CreateDhtNode creates a libp2p Node and a DHT on top of it
-func CreateDhtNode(ctx context.Context, runenv *runtime.RunEnv) (host.Host, *kaddht.IpfsDHT, error) {
-	// Test Parameters
-	var (
-		bucketSize  = runenv.IntParamD("bucket_size", 20)
-		autoRefresh = runenv.BooleanParamD("auto_refresh", true)
-	)
-
+func CreateDhtNode(ctx context.Context, runenv *runtime.RunEnv, bucketSize int, autoRefresh bool) (host.Host, *kaddht.IpfsDHT, error) {
```
raulk: I'd encapsulate the parameters in a struct for a cleaner interface. We may want to create a little helper method that binds test params from the RunEnv into a struct. Actually, this already exists: https://github.com/mitchellh/mapstructure. Wanna try integrating it inside the `sdk/runtime` module?

```go
type TestParams struct {
    BucketSize *int 	`mapstructure:"bucket_size"`
    AutoRefresh *bool 	`mapstructure:"auto_refresh"`
}

runenv := runtime.CurrentRunEnv()

var tp TestParams
if err := runenv.DecodeTestParams(&tp); err != nil {
    runenv.Abort(err)
	return
}
```

raulk: And to deal with defaults, you could have a variable with the defaults, and use mergo to merge the incoming parameters into it.

plans/dht/test/find_peers.go
daviddias: done

raulk: Shouldn't we be deferring at the top?

plans/dht/test/find_providers.go
```diff
@@ -1,20 +1,110 @@
...
+
+	var (
+		cidsToPublish     []cid.Cid
+		nCidsToPublish    = 5
+		cidsToNotPublish  []cid.Cid
+		nCidsToNotPublish = 5
+	)
+
+	for i := 0; i < nCidsToPublish; i++ {
+		v := fmt.Sprintf("%d -- gonna publish", i)
```
raulk: Log into the runenv instead?

```
runenv.Message()
```

plans/dht/test/provide-stress.go
```diff
@@ -1,20 +1,75 @@
...
 
 // ProvideStress implements the Provide Stress test case
 func ProvideStress(runenv *runtime.RunEnv) {
-	fmt.Printf("Not implemented yet")
+	// Test Parameters
+	var (
+		timeout     = time.Duration(runenv.IntParamD("timeout_secs", 60)) * time.Second
+		randomWalk  = runenv.BooleanParamD("random_walk", false)
+		bucketSize  = runenv.IntParamD("bucket_size", 20)
+		autoRefresh = runenv.BooleanParamD("auto_refresh", true)
```
raulk: We should be adding all the new parameters to the `dht.toml` manifest. This will come in handy when you do:

```sh
$ testground describe dht/provide-stress
Stress tests the provider subsystem of the DHT by doing bla bla bla

Parameters:
  timeout_secs	duration	how long the test is allows to run (default: 60s)
  random_walk	bool		bla bla bla (default: false)
  bucket_size	int			bucket size of the routing table (default: 20)
  ...
```

plans/dht/test/find_peers.go
daviddias: ✅

plans/dht/test/provide-stress.go
daviddias: ✅

plans/dht/test/find_providers.go
daviddias: This string is being used only as the data that gets hashed to create the CID

go.mod
daviddias: Gotcha! Didn't know of `go mod tidy`

plans/dht/test/create-dht-node.go
daviddias: Of course! :D

daviddias: I'm moving this last refactor bit to a separate PR, tracked at https://github.com/ipfs/testground/issues/160, so that all of this code doesn't get blocked until Monday/Tuesday with all the travels and catch up.

daviddias: This way, the next refactor will be already supported by CI #157  :)

PR #139, opened 2019-11-07
aws/ecr: fix b64 encoding.

raulk:


raulk:
Went ahead and merged this tiny fix.

PR #155, opened 2019-11-08
Adding Windows support

aschmahmann:
Get testground to run on Windows.

Four fixes:
- Remove /x/sys go.mod dependency (for some reason this just breaks things)
- Switch from `$HOME` environment variable to the OS agnostic `os.UserHomeDir`
- Connect to docker containers using named pipes instead of unix sockets
   - While Windows 10 does support unix sockets, for some reason it's not utilized by the `go-connections` utilized by docker
   - @raulk we might be able to make this go away with environment variables, so there's now an additional reason for configuring docker via environment variables
- Manifests now uses the correct file protocol handler `file://` (instead of `file:`)


pkg/build/golang/docker.go
```diff
@@ -62,8 +63,17 @@ func (b *DockerGoBuilder) Build(in *api.BuildInput) (*api.BuildOutput, error) {
 	// TODO support specifying a docker endpoint + TLS parameters from env.
 	// raulk: I don't see a need for this now, as we certainly want to do builds
 	// locally, and push the image to a registry.
+	// aschmahmann: No host config parameters means we need some OS check
+
+	var host string
+	if runtime.GOOS == "windows" {
```
raulk: We should be able to replace this with `client.DefaultDockerHost`. The docker project uses OS build constraints (see https://github.com/moby/moby/blob/master/client/client_windows.go and https://github.com/moby/moby/blob/master/client/client_unix.go) to vary the value based on the target OS.

aschmahmann: even better!

daviddias: @aschmahmann can you test this?

aschmahmann: @daviddias If we're using the `client.DefaultDockerHost` it should be fine.

raulk:
> @raulk we might be able to make this go away with environment variables, so there's now an additional reason for configuring docker via environment variables

We don't want to build the docker client using the env variables (`DOCKER_HOST`) because those could be pointing elsewhere. We really want these builds to be executed locally, so that the `local:docker` runner can pick up the local images.

However, if we want to get fancy, we should add support for specifying the docker endpoint in this builder like the `cluster:swarm` runner does, by picking up explicit values from the env file.

raulk:
Thanks for making the changes. I think this is OK now, but I'll let David pronounce himself.

PR #111, opened 2019-11-01
Ansible + Terraform of Docker Swarm Cluster in AWS - (Temporary ansible scripts - will be replaced with 'roles' layout in a separate PR)

jimpick:
Work-in-progress for automating cluster setup using Ansible.

PR #156, opened 2019-11-08
Proposal to separate Provisioning, Building and Running into separate Phases

daviddias:
@raulk this is where my head was headed a bit ago. The PR adds some light touches to a Plan config and a Plan spec. The general idea is:

- Create a separate option for Provisioning
- Separate the elements of the provision step out of the runner (i.e. running in a cluster of docker swarm nodes or multiple docker containers is effectively the same thing for the runner, what changes is the provisioning sets up that cluster)
- We keep a Vagrant provisioner so that we can have a reproducible env in all architectures
- We keep the option open to treat some pre-provisioned "neighboor:boxes" to run our tests.
- This also helps map a plan spec to the actual plan manifest

### DIFF
--- a/manifests/dht.toml
+++ b/manifests/dht.toml
@@ -1,62 +1,81 @@
+[[base]]
 name = "dht"
-# hashicorp/go-getter URLs, so in the future we can support fetching test plans
-# from GitHub.
+# hashicorp/go-getter URLs, so in the future we fetch plans from anywhere
 source_path = "file:${TESTGROUND_SRCDIR}/plans/dht"
+    [defaults]
+    provisioner = "terraform:aws"
+    builder = "docker:go"
+    runner = "local:docker"
 
-[build_strategies."docker:go"]
-enabled = true
-go_version = "1.13"
-module_path = "github.com/ipfs/testground/plans/dht"
-exec_pkg = "."
+[[provisioners]]
+    [provisioner_strategies. "vagrant:local"]
+    enabled = true
 
-# TODO: exec:go is not ready yet
-[build_strategies."exec:go"]
-enabled = true
-module_path = "github.com/ipfs/testground/plans/dht"
-exec_pkg = "."
+    [provisioner_strategies. "terraform:aws"]
+    enabled = true
 
-[run_strategies."local:docker"]
-enabled = true
+    [provisioner_strategies. "terraform:aws:docker-swarm"]
+    enabled = true
 
-[run_strategies."local:exec"]
-enabled = true
+    [provisioner_strategies. "terraform:do"]
+    enabled = true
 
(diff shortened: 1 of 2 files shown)


manifests/dht.toml
```diff
@@ -1,62 +1,81 @@
...
 source_path = "file:${TESTGROUND_SRCDIR}/plans/dht"
+    [defaults]
+    provisioner = "terraform:aws"
+    builder = "docker:go"
+    runner = "local:docker"
 
-[build_strategies."docker:go"]
-enabled = true
-go_version = "1.13"
-module_path = "github.com/ipfs/testground/plans/dht"
```
daviddias: I've removed this as it was appearing repeated in multiple locations

manifests/dht.toml
```diff
@@ -1,62 +1,81 @@
...
+    go_version = "1.13"
+    exec_pkg = "."
+
+    # TODO: exec:go is not ready yet
+    [build_strategies."exec:go"]
+    enabled = false
+    exec_pkg = "."
+
+[[runners]]
+instances = { min = 2, max = 200, default = 25 }
```
daviddias: It makes more sense if instances is a runners param

manifests/dht.toml
```diff
@@ -1,62 +1,81 @@
+[[base]]
 name = "dht"
-# hashicorp/go-getter URLs, so in the future we can support fetching test plans
-# from GitHub.
+# hashicorp/go-getter URLs, so in the future we fetch plans from anywhere
 source_path = "file:${TESTGROUND_SRCDIR}/plans/dht"
+    [defaults]
+    provisioner = "terraform:aws"
```
raulk: I don't think Terraform should appear here. We are certainly not going to run a terraform script from the testground; IMO that's too much magic / "spooky action at a distance". We should just assume that if the user is hitting an AWS backend, they've already prepared it accordingly (with Terraform scripts).

What we can do is print out of a warning with instructions if we detect that the environment is not properly set up, e.g. if the docker swarm commands fail. The warning could look like this:

```
WARNING

Testground has detected that your AWS environment is not properly set up.
Follow the instructions here to fix this: git.io/...
```

raulk: In the same vein, I don't think we should be calling this a provisioner, but rather a backend. We are not provisioning any machines, we are just using an infrastructure backend to run our tests.

daviddias: Gotcha. So for the `test run` itself this would be just the `backend` or `infra`.

`Testground` can still offer a command for `provision`, no? Or do we want to keep all the infra scripts in a folder forever like it is coming up on #111

daviddias:
Separation of concerns convo continues at #224

PR #159, opened 2019-11-08
add a describe command.

### LINKED ISSUE #45: CLI describe command
Outputs information about a test plan or a test case for now. In the future it may be extended to inspect builders and runners, and their configuration properties.

raulk:
Fixes #45.

Test with: 
- `testground describe dht`
- `testground describe dht/find-peers`

### DIFF
--- a/pkg/api/definitions.go
+++ b/pkg/api/definitions.go
@@ -1,5 +1,13 @@
 package api
 
+import (
+	"fmt"
+	"io"
+	"text/tabwriter"
+
+	"github.com/mitchellh/go-wordwrap"
+)
+
 // TestPlanDefinition represents a test plan known by the system.
 // Its name must be unique within a test census.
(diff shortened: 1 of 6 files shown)


pkg/api/definitions.go
```diff
@@ -47,3 +56,51 @@ func (tp *TestPlanDefinition) TestCaseByName(name string) (seq int, tc *TestCase
 	}
 	return -1, nil, false
 }
+
+func (tp *TestPlanDefinition) Describe(w io.Writer) {
+	p := func(w io.Writer, f string, a ...interface{}) {
```
daviddias:
```suggestion
	print := func(w io.Writer, f string, a ...interface{}) {
```

tiny nit not worth changing but I felt sharing anyway :) having a word doesn't bite and saves a reader to have to look at the args to understand that it is for printing

PR #163, opened 2019-11-09
feat(runner): separate control and per-run data networks

Stebalien:
1. Use separate control/data networks everywhere.
2. Create a new data network for every run.

I've tested the local docker runner changes but am unsure how to test it on AWS.

### DIFF
--- a/pkg/runner/cluster_swarm.go
+++ b/pkg/runner/cluster_swarm.go
@@ -155,12 +155,47 @@ func (*ClusterSwarmRunner) Run(input *api.RunInput) (*api.RunOutput, error) {
 		return nil, fmt.Errorf("testground-redis service doesn't exist in the swarm cluster; aborting")
 	}
 
+	// Create the network.
+
+	log.Infow("creating network", "name", sname)
+
+	networkSpec := types.NetworkCreate{
+		Driver:         "overlay",
+		CheckDuplicate: true,
+		EnableIPv6:     true, // TODO: params?
+		Internal:       true,
+		Attachable:     true,
+		Scope:          "swarm",
+		Labels: map[string]string{
+			"testground.plan":     input.TestPlan.Name,
+			"testground.testcase": testcase.Name,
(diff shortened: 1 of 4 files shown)


pkg/runner/cluster_swarm.go
```diff
@@ -155,12 +155,47 @@ func (*ClusterSwarmRunner) Run(input *api.RunInput) (*api.RunOutput, error) {
...
+		Driver:         "overlay",
+		CheckDuplicate: true,
+		EnableIPv6:     true, // TODO: params?
+		Internal:       true,
+		Attachable:     true,
+		Scope:          "swarm",
+		Labels: map[string]string{
+			"testground.plan":     input.TestPlan.Name,
+			"testground.testcase": testcase.Name,
+			"testground.runid":    input.RunID,
```
Stebalien: I'm using this to find all networks associated with the given test run.

pkg/runner/cluster_swarm.go
```diff
@@ -155,12 +155,47 @@ func (*ClusterSwarmRunner) Run(input *api.RunInput) (*api.RunOutput, error) {
...
+			"testground.name":     "default", // default name. TODO: allow multiple networks.
```
Stebalien: This is the name that tests and the sidecar will use. (we may support more network interfaces in the future)

pkg/runner/cluster_swarm.go
```diff
@@ -172,6 +207,11 @@ func (*ClusterSwarmRunner) Run(input *api.RunInput) (*api.RunOutput, error) {
 			ContainerSpec: &swarm.ContainerSpec{
 				Image: image,
 				Env:   env,
+				Labels: map[string]string{
+					"testground.plan":     input.TestPlan.Name,
```
raulk: Nice one. These labels are gonna come in handy in general.

raulk:
Fixes #153.

Stebalien:
Thanks for looking into that.

PR #165, opened 2019-11-12
docker:go builder: add support for goproxy + travis streamlining

### LINKED ISSUE #162: Spawn local GOPROXY when building docker images
We should consider exporting the user's locally cached packages via a local GOPROXY to docker containers when we build. This should speed up docker builds considerably.

raulk:
**We will want to squash this PR before merge, unless we fancy merging the delightful commit history into master. It _is_ a piece of history, after all.**

---

This PR introduces three modes of operation in the `docker:go` builder:

* The `local` mode (default) starts a proxy container (if one doesn't exist yet) with bridge networking, and will configure the build to use that proxy.
* The `direct` mode sets the `GOPROXY=direct` env var on the go build.
* The `remote` mode specifies a custom proxy. The `GoProxyURL` field must be non-empty.

---

This PR also creates a new `pkg/docker` page for Docker utility and helper functions.

---

This PR also streamlines our `.travis.yml`. The smoke runs were a bit flaky, especially the Docker one, so I disabled the latter. The linter needs to be added outside of the source folder to avoid it overriding dependency versions. Also consolidated all the build steps into two, which speeds things up drastically. Added Travis caches.

---

## End-to-end build time benchmark

Build time reduced by ~45%.

```sh
⟩ time ./testground -vv build dht --builder=docker:go --build-cfg bypass_cache=true --build-cfg go_proxy_mode=direct
...
0.03999s    INFO    testground-build network found  {"buildId": "4e125a1f-4619-4d6f-bc27-22301d822eff", "networkId": "9923f83a4c8da1d1c29a7d3cebfaaccf13c9c7f258edaaa0369fb0ef3411c895"}
...
      136.15 real         0.06 user         0.05 sys

⟩ time ./testground -vv build dht --builder=docker:go --build-cfg bypass_cache=true --build-cfg go_proxy_mode=local
...
0.08073s    INFO    testground-build network found  {"buildId": "c60b2ecd-3780-4d7e-916f-d8353ebff88d", "networkId": "9923f83a4c8da1d1c29a7d3cebfaaccf13c9c7f258edaaa0369fb0ef3411c895"}
0.08080s    INFO    go proxy enabled; ensuring testground-goproxy container is started  {"buildId": "c60b2ecd-3780-4d7e-916f-d8353ebff88d"}
0.08083s    DEBUG   checking state of container {"buildId": "c60b2ecd-3780-4d7e-916f-d8353ebff88d", "containerName": "testground-goproxy"}
0.08458s    INFO    container found {"buildId": "c60b2ecd-3780-4d7e-916f-d8353ebff88d", "containerName": "testground-goproxy", "containerId": "34e17595467ac58caf47a96c54827a00f333afc0efe95ae665c29862395a61ab", "state": "running"}
...
       74.49 real         0.05 user         0.06 sys
```

---

Fixes #162.


.travis.yml
```diff
@@ -11,44 +15,57 @@ services:
...
-  - test
-  - run
-
+# NOTE: The smoke test itself will fail because the Travis redis instance does
+# not have keyspace notifications enabled. But we're interested in exercising
+# the testground logic, not the test plan itself. So it's fine.
 jobs:
   include:
-    - stage: check
+    - stage: build_check_test_run
```
daviddias: All smashed together?

daviddias:
```suggestion
    - stage: lint_build_check_test_run
```

.travis.yml
```diff
@@ -11,44 +15,57 @@ services:
...
-    - stage: test
-      os: linux
-      script:
-        - go test -v -race ./...  # Run all the tests with the race detector enabled
-
-    - stage: test
-      os: osx
-      script:
         - go test -v -race ./...  # Run all the tests with the race detector enabled
+        - go build .
```
daviddias: why two build times?

.travis.yml
raulk: yeah, each steps adds significant latency and there's no clear benefit to separating them because they were executed sequentially anyway

.travis.yml
raulk: first one builds all packages, but produces no executable. second one builds the testground executable.

.travis.yml
```diff
@@ -11,44 +15,57 @@ services:
...
```
daviddias:
```suggestion
```

only build once

.travis.yml
```diff
@@ -11,44 +15,57 @@ services:
...
+        - ./testground -vv run dht/find-peers --builder=exec:go --runner=local:exec --test-param timeout_secs=15
 
-    - stage: test
+    - stage: build_check_test_run
       os: windows
       script:
+        - echo 'travis_fold:start:go.mod.graph' && echo 'go mod graph' && go mod graph && echo 'travis_fold:end:go.mod.graph'
+        - go build ./...          # build the project
         - go test -v -race ./...  # Run all the tests with the race detector enabled
+        - go build .
```
daviddias:
```suggestion
```

only build once

.travis.yml
daviddias: ah, got your memo now. So I guess we do need to build twice?

.travis.yml
```diff
@@ -63,12 +80,6 @@ jobs:
 #      script:
 #        - go run main.go run dht/find-peers --builder=exec:go --runner=local:exec
 
-    - stage: run
```
daviddias: So no more smoke testing? This will expose us to refactor/reshuffle erros until we have wide test coverage.

.travis.yml
raulk: AFAIK, yes.

.travis.yml
raulk: The smoke testing is covered here on Linux: https://github.com/ipfs/testground/pull/165/files#diff-354f30a63fb0907d4ad57269548329e3R32 No docker thing because that's flaky, and we're playing tricks with networks in a way that Travis likely won't like.

pkg/build/golang/Dockerfile.template
```diff
@@ -32,15 +34,16 @@ COPY /plan/go.mod ${PLAN_DIR}
 COPY /sdk/sync/go.mod /sdk/sync/go.mod
 COPY /sdk/iptb/go.mod /sdk/iptb/go.mod
 COPY /sdk/runtime/go.mod /sdk/runtime/go.mod
+
 # Download deps.
 RUN cd ${PLAN_DIR} \
-    && go env -w GOPROXY=direct \
+    && go env -w GOPROXY=${GO_PROXY} \
```
Stebalien: nit: quotes?

pkg/docker/container.go
```diff
@@ -0,0 +1,101 @@
...
+
+type EnsureContainerOpts struct {
+	ContainerName    string
+	ContainerConfig  *container.Config
+	HostConfig       *container.HostConfig
+	NetworkingConfig *network.NetworkingConfig
+}
+
+// EnsureContainer ensures there's a container started of the specified kind.
+func EnsureContainer(ctx context.Context, log *zap.SugaredLogger, cli *client.Client,
```
Stebalien: :heart:

daviddias:
@raulk this PR takes out the smoke tests (the two runs, on exec and docker) in CI and only checks if the code builds in two platforms. I’m afraid this will reduce the actual ability to ensure that the Test Ground code is fully working (for a definition of working that is testing the most complex test we have). That said, I acknowledge the velocity tradeoff and I take your decision as the architect 
🙂

pkg/build/golang/Dockerfile.template
raulk: good call

raulk:
@daviddias we do still smoke test the `local:exec` runner: https://github.com/ipfs/testground/pull/165/files#diff-354f30a63fb0907d4ad57269548329e3R32

The docker runner was just not working well on Travis, so for now I commented it out. We may want to invest time in the future to get that working.

daviddias:
Ah, you are right. My bad, I missed that line

Issue #154, opened 2019-11-08
Infra: AWS Autoscaling


jimpick:
The Terraform scripts already have the workers configured in an Autoscaling group with a manual strategy.

To get the clusters to "scale in" when not in use, we could use a custom autoscaling strategy:

https://docs.aws.amazon.com/autoscaling/plans/userguide/gs-configure-scaling-plan.html

It seems we could simply publish a custom metric (perhaps with the desired container count?) to AWS Cloudwatch and hook the autoscaling strategy to that:

https://docs.aws.amazon.com/autoscaling/plans/userguide/gs-specify-custom-settings.html

The only thing the new nodes need to do on startup is setup filebeat and join the docker swarm. We could fetch the needed data from the manager machine which is always up (I think we can give it a common static IP in the VPC).

When the cluster does a "scale in" (we should probably only do this when no jobs are running), we'll get orphaned entries in the docker swarm node table, so to keep things tidy, we'll want to periodically remove them, perhaps with a daily cron job.

We could have a script to manually set the CloudWatch network for times when we want the cluster to be scaled out even when no jobs are running.

dirkmc:
I remember Edgar mentioning that we should be careful about scaling up and down too much as it can cause inconsistent results in benchmarks. @hinshun could you confirm?

hinshun:
It's not the benchmarks per-say, but we observed higher number of failures with ASGs bigger than 200 nodes. I think combined with the `libp2p` connection bugs, it was hard to configure the network topology reliably.

Small clusters (like 3 peers) may have some inconsistent results because `bitswap` may end up transferring all data only from 1 peer. This doesn't occur in bigger clusters (4+).

I suggest you opt-in to the beta EC2 limits based on vCPUs because it's much easier to manage than requesting limit increases on every instance type you want to scale higher in.

jimpick:
Yep, I requested a raise to 128 vCPUs and got it ... might need to get that raised again if we go really big. I'm new to this aspect of AWS. :-)

daviddias:
I'm not sure if I understand why would we use auto scaling in the first place. The test plan writer should have control over the number instances created for the test case (number of docker containers spawned, each with an independent node) and how over many machines those instances are going to be instantiated (during the step of setting up the cluster).

jimpick:
Related PR: #175 

As the cluster "scales out", the worker nodes will automatically configure themselves and join the swarm.

jimpick:
[...]

PR #175, opened 2019-11-19
Infra: Setup worker nodes automatically on boot

jimpick:
When the worker nodes boot (for example, when the autoscaling group
scales "out"), they will run a startup script that will fetch
data files via HTTP from the manager node. The following services
are automatically set up:

 * ssh keys so the manager node can ssh to any worker
 * Filebeat
 * the worker nodes join the Docker Swarm

On the manager node, we now use ansible to generate an ssh keypair,
as will as set up an Apache server.

The packer image now calls the startup script as a new service. (This
will need to to updated once merged to master)

I also updated the terraform files to use the newer syntax for
terraform 0.12.x to avoid deprecation warnings (I was using an older
version previously).


jimpick:
There's a bit of follow-on work to run the Ansible scripts on the manager node automatically from Terraform. I'd like to get this merged and submit that as a separate PR.

jimpick:
The terraform config now set a default number of workers to zero ... because we want to run the Ansible scripts on the manager node before they start up. It's probably OK to spin them up before because the Ansible scripts will find them and configure them, but I'd prefer it if all worker nodes were setup using the startup script. My plan is to do some more work on the terraform config in a separate PR to use `depends_on` to bring up the cluster and automatically run the Ansible scripts in a controlled order.

jimpick:
I discovered it's missing one thing ... the Docker labels for the workers can only be applied on the manager node, so I'll have to implement something to do that ... probably a simple CGI script on the manager node will do the trick.

jimpick:
There is now a CGI script on the web server on the manager node so new workers can join the swarm and be properly labeled.

Also, I'm now naming the AMI images in AWS when building from Packer, and doing a lookup for the latest image from the Terraform configuration - this makes using multiple AWS regions easier.

This is ready to merge. (I'm hoping to do some more automation work ... eg. get Terraform to run the Ansible scripts automatically after this is merged)

daviddias:
> When the worker nodes boot (for example, when the autoscaling group
scales "out")

Can you unpack this a bit more. AFAIU, AWS autoscalling will hurt our test performance as it is a reactive measure to counter tremendous amount of load -- https://github.com/ipfs/testground/issues/154 --. 

Each cluster size should be managed by the tester or the Infra admin.

infra/ansible/files/register-worker.cgi
```diff
@@ -0,0 +1,22 @@
...
+if [ -z "$WORKER" ]; then
+  echo 'Need worker=<hostname> in body'
+  exit 0
+fi
+
+echo docker node update --label-add TGRole=worker $WORKER
+docker node update --label-add TGRole=worker $WORKER
+
+exit 0
+~
```
daviddias: Are we installing an Apache Server, just so that we can run a single .cgi script to call `docker node update`? Seems a bit over the top. Could this be done by a tiny go program so that in the future is part of the testground daemon?

daviddias:
@olizilla your ansible + terraform wisdom would be timely as we make progress in making all of its usage simpler :)

infra/ansible/files/register-worker.cgi
lanzafame: I agree with @daviddias on this. In my afternoons worth of digging, we are already controlling docker swarm from within `pkg/runner/cluster_swarm.go` so it doesn't make sense to me that we are doing metadata coordination outside of this in the infrastructure layer, i.e. ansible, as well.

infra/terraform/testground.tf
```diff
@@ -66,35 +66,44 @@ resource "aws_security_group" "default" {
...
+    values = [var.aws_ami]
+  }
+}
+
 resource "aws_instance" "testground_manager" {
-  ami                    = "${lookup(var.aws_amis, var.aws_region)}"
-  availability_zone      = "${var.aws_availability_zone}"
-  instance_type          = "${var.aws_instance_type_manager}"
-  key_name               = "${var.key_name}"
+  ami                    = data.aws_ami.base.id
```
nonsense: If we are to use these scripts for long-lived infrastructure for Testground, we might want to lock the `ami` to a constant value. Otherwise a few days/weeks after we run `terraform apply` for the first time, Terraform would start suggesting that we replace the VMs with the newest AMI, which becomes a problem if we want to apply other changes to the infrastructure.

infra/terraform/testground.tf
```diff
@@ -66,35 +66,44 @@ resource "aws_security_group" "default" {
...
 
 resource "aws_instance" "testground_redis" {
-  ami                    = "${lookup(var.aws_amis, var.aws_region)}"
-  availability_zone      = "${var.aws_availability_zone}"
-  instance_type          = "${var.aws_instance_type_redis}"
-  key_name               = "${var.key_name}"
+  ami                    = data.aws_ami.base.id
+  availability_zone      = var.aws_availability_zone
+  instance_type          = var.aws_instance_type_redis
+  key_name               = var.key_name
```
nonsense: AFAIK it doesn't matter which key we use to create VMs (so using personal keys such as `jim-us-west-2` is fine), as long as we then replace the `authorized_keys` file. I think it would be a good practice to collect the keys from team members and add everyone's key to the `authorized_keys` right away as a first step when configuring masters and workers.

infra/ansible/files/register-worker.cgi
lanzafame: @jimpick Using something like [AWS Launch Templates](https://capstonec.com/auto-scaling-docker-nodes-in-aws/) we could put a message onto redis and have the `manager` node listen for that?

infra/ansible/setup-worker.yml
```diff
@@ -0,0 +1,53 @@
...
+  tasks:
+    - name: Get hosts
+      debug: var=ansible_play_hosts
+    - name: Get hostname
+      debug: var=ansible_facts.hostname
+    - name: Add keys to authorized_keys
+      authorized_key:
+        user: ubuntu
+        state: present
+        key: http://172.16.0.10/authorized_keys
```
nonsense: I guess you are fetching the `authorized_keys` from the `master`, since the prv/pub SSH key for the `master` is currently dynamic and auto-generated during provisioning time.

I am not sure why we need the workers to have SSH access using the master's SSH key - I think we should be just fine by applying a static `authorized_keys` file which contains 3-4 public keys for all team members.

infra/ansible/setup-worker.yml
```diff
@@ -0,0 +1,53 @@
...
+    - name: get filebeat.yml from http server on manager
+      get_url:
+        url: http://172.16.0.10/filebeat.yml
```
nonsense: Similar question as the one above - why do we need to fetch the `filebeat.yml` from the `master` and expose over HTTP ? I think this file should be static, and we should be able to apply it to `/etc/filebeat/filebeat.yml` directly with Ansible.

infra/ansible/files/register-worker.cgi
nonsense: AFAICT `cluster_swarm.go` is used to run a `docker service` on an already established `docker swarm` cluster.

I believe we are applying labels to the workers here, so that we can assign workloads only to the workers through the use of labels... is that right @jimpick ? I think this actually should be done when we setup the infrastructure on the Terraform/Ansible layer, and not when we run `testground`.

I also think we should be able to do this without Apache web server though - we could just run `docker node update` as part of the Ansible playbook to setup the necessary labels on workers.

infra/ansible/setup-worker.yml
olizilla: +1, this is what we do on the gateways'n'bootstrappers

olizilla:
As @jimpick already mentioned some of the manual steps could be collapsed into one by having the steps as ansible roles, and a playbook that applies them.

More generally, I'm unclear on the pattern and constraints here... ansible is good for mutating things in place, terraform is good for defining immutable things. Are test clusters expected to be long lived and altered in place from run to run? or destroyed and rebuilt per testcase?

I think if this PR works, then great, it should go in. Using a go program vs apache vs redis messaging are all ways to achieve the same thing. Then we should start talking about how to simplify it... as @nonsense points to in his comments, it feels like we could be doing more static configuring rather than runtime registering. @jimpick I'd be happy to jump on a call and talk it through... I'd help me to get a better understanding of the constraints and hoops your jumping through.

And one weird trick to speed up ansible... 

**ansible.cfg**
```
[ssh_connection]
pipelining = true
```
https://github.com/protocol/bifrost-infra/pull/288
https://www.toptechskills.com/ansible-tutorials-courses/speed-up-ansible-playbooks-pipelining-mitogen/

jimpick:
> > When the worker nodes boot (for example, when the autoscaling group
> > scales "out")
> 
> Can you unpack this a bit more. AFAIU, AWS autoscalling will hurt our test performance as it is a reactive measure to counter tremendous amount of load -- #154 --.
> 
> Each cluster size should be managed by the tester or the Infra admin.

"Autoscaling group" is AWS nomenclature for the concept of a group of servers managed as a single unit. We're using the "group" part of that currently. We're not currently using the "autoscaling" feature.

infra/ansible/files/register-worker.cgi
jimpick: The Apache server also serves the filebeat.yml, authorized_hosts and docker swarm join key to any new worker nodes that boot up.

I initially entertained the idea of writing a single-purpose HTTP server in Go that does the exact same thing as what Ansible does with installing Apache + the CGI script, but since we were already using Ansible, it seemed cleaner to me to keep it all together in one place.

jimpick: > AFAICT `cluster_swarm.go` is used to run a `docker service` on an already established `docker swarm` cluster.
> 
> I believe we are applying labels to the workers here, so that we can assign workloads only to the workers through the use of labels... is that right @jimpick ? I think this actually should be done when we setup the infrastructure on the Terraform/Ansible layer, and not when we run `testground`.
> 
> I also think we should be able to do this without Apache web server though - we could just run `docker node update` as part of the Ansible playbook to setup the necessary labels on workers.

Agree, this is a step that happens when preparing infrastructure before the `testground` binary is run. (I imagine the scope of the binary could be expanded to include managing infrastructure, but it currently doesn't do that)

This PR is for a boot script that gets run automatically on a new worker node when it is added to the swarm. It can't run Ansible commands on the manager node in order to run `docker node update` (that command can't be run on worker nodes), so that's why I had to do the CGI script. 

Alternatively, I could have distributed the private ssh key and installed it on the worker nodes (via HTTP) so they could run Ansible commands on the manager node, but that felt wrong to me.

jimpick: > @jimpick Using something like [AWS Launch Templates](https://capstonec.com/auto-scaling-docker-nodes-in-aws/) we could put a message onto redis and have the `manager` node listen for that?

Looks like they are just putting the Docker Swarm join key into the user data, which I think is possible with either Launch Templates or Launch Configurations. We could potentially do that because we create the Launch Configuration from Terraform, but I'd have to experiment ... it would have to create the Docker Swarm first, and then export the join key to Terraform somehow so it could use it when generating the Launch Configuration or Launch Template.

We do spin up Redis as a Docker service, but I don't think it's exposed externally.

infra/terraform/testground.tf
jimpick: I previously had the hard-coded AMIs (there's a different AMI for each region), but I think it's going to be more flexible to use AWS Tags and the matching logic (see `data "aws_ami" base {}`) ... that way, we don't don't have to update them every single time we change the Packer image, but we can still version things with the tags).

infra/terraform/testground.tf
jimpick: That is a feature I'm hoping to implement as part of #186 - agree that we could have a default that matches a keypair we have on our AWS account. Non-PL users would have to supply their own, of course.

infra/ansible/setup-worker.yml
jimpick: I generally use ssh-agent forwarding, so I can usually run the Ansible scripts on the manager node and it can access the other nodes. Sometimes I forget to add the ssh key to my agent and I see a failure. Where this will be really useful is when we want to run Ansible scripts from Terraform, which won't have ssh-agent.

infra/ansible/setup-worker.yml
jimpick: Alternatively, I could have used the same strategy that I use on the manager node where I build the filebeat.yml from a template using the secrets stored in the S3 bucket.

jimpick:
> And one weird trick to speed up ansible...
> 
> **ansible.cfg**
> 
> ```
> [ssh_connection]
> pipelining = true
> ```
> 
> [protocol/bifrost-infra#288](https://github.com/protocol/bifrost-infra/pull/288)
> https://www.toptechskills.com/ansible-tutorials-courses/speed-up-ansible-playbooks-pipelining-mitogen/

I didn't know that one, thanks!

nonsense:
> > > When the worker nodes boot (for example, when the autoscaling group
> > > scales "out")
> > 
> > 
> > Can you unpack this a bit more. AFAIU, AWS autoscalling will hurt our test performance as it is a reactive measure to counter tremendous amount of load -- #154 --.
> > Each cluster size should be managed by the tester or the Infra admin.
> 
> "Autoscaling group" is AWS nomenclature for the concept of a group of servers managed as a single unit. We're using the "group" part of that currently. We're not currently using the "autoscaling" feature.

I completely agree with @jimpick that we are just using the group feature right now, but in its current form the auto-scaling group prevents use from `stopping` instances for some reason. It might be a good idea to be able to `stop` instances overnight and `start` them back up, rather than re-create them - I am not 100% convinced we need this, but could be necessary for individual developer clusters.

daviddias:
What's missing to get this PR merged? @nonsense ?

nonsense:
@daviddias I have opened a new PR - https://github.com/ipfs/testground/pull/210, so that we compare more easily the different approaches. I believe I have simplified a few things in the other PR, so let's review it, and maybe then we close this one.

jimpick:
Superceded by #210

PR #183, opened 2019-11-22
initial daemon implementation

nonsense:
This PR is addressing https://github.com/ipfs/testground/issues/149 by:
- adding an HTTP server and converting testground into `thinclient <-> server` architecture ala `docker`.
- moving initialisation of config away from `Engine` as it is used for configuring `Server` and `Client` now as well

TODO:
- [x] Errors propagation - it would be good to propagate errors happening on `server` to the `client`. This is currently done only for `testground describe` with `errf`.
- [x] Decide whether `.env.toml` is required or optional. Gracefully handle config initialisation if `.env.toml` is missing is not there yet.
*Solution*: I think we should consider `.env.toml` to be optional for as long as we can. We should use meaningful defaults.
- [x] Decide on logger timestamp - I think RFC3339 format is more useful for a long-running daemon than elapsed time
- [x] Decide whether to start a daemon automatically if `[client]` section is missing in `.env.toml`, making it simpler to run `testground` when client/server are on the same machine. This will probably be handy during development. Also currently tests fail because of this.
*Solution*: `daemon` automatically runs if the `Client.Endpoint` is missing or an empty string. If it is set to anything else, we assume that a daemon must be running. `testground daemon` returns an fatal error if `Daemon.Listen` is missing or an empty string though.
- [x] Introduce protocol between `server <-> client` when receiving chunked progress responses with a final message containing result and maybe additional error information. For example when calling `testground build` we want to see progress of build, but also get the `ArtifactPath` at the end of the request.
*Solution*: for now I added a simple JSON message protocol where an API call returns an `io.ReadCloser` (see `build`, and `run`) and then the client processes all messages. Message type currently looks like:
```
type Msg struct {
	Type    string      `json:"type"` // progress or result or error
	Payload interface{} `json:"payload,omitempty"`
	Error   *Error      `json:"error,omitempty"`
}
```
And clients should know how to decode the `Payload`. For a given API request, in general the client should expect 0-n `progress` messages, and exactly 1 `result` message.
- [x] Document HTTP API for testground daemon
- [x] Godoc for `daemon` package and for `client`.

MAYBE FUTURE PRS:
- [ ] API between `testground client -> testground server` cleanup
- [ ] API between `testground server -> engine` cleanup


daviddias:
@nonsense you mentioned on the Weekly call #188 that you were looking for some Code Review in this PR. However, I see that there is a bunch of TODOs, CI is red and that this PR is still declared as "draft" (i.e. not ready to review). Maybe I misunderstood, do you want to have a review now or later when you get into a ready to review state?

nonsense:
@daviddias probably best when the PR is marked as ready for review. During the Weekly call I mentioned it might be worth checking the PR, because there were questions regarding the `daemon`, and this part is mostly established now, and I don't think will change much.

nonsense:
Windows build appears to fail because of https://github.com/golang/go/issues/34610 and https://github.com/moby/moby/pull/40021 . I tried upgrading `docker/docker` in `go.mod`, but this results in:

```
go: github.com/ipfs/testground/pkg/build/golang imports
        github.com/docker/docker/pkg/archive imports
        github.com/Sirupsen/logrus: github.com/Sirupsen/logrus@v1.4.2: parsing go.mod:
        module declares its path as: github.com/sirupsen/logrus
                but was required as: github.com/Sirupsen/logrus
```

I am not sure how to fix this right now, so will leave Windows build broken for now, and continue with the rest of the todos. If anyone knows how to update `docker/docker` to latest version, so that Windows build passes, ideas are welcome :)

raulk:
@nonsense unfortunately Docker doesn't use go.mod and they don't either make releases that I know of 😨 To upgrade to the laster master, you want to do `go get github.com/docker/docker@master`. However, the version we're using works well with the `x/sys` version we depend on. In the past, we've gone through hell to make ends meet because the Go team breaks `x/sys/windows` frequently, and the interplay of the various downstream dependencies makes picking the right version really tough. We've literally had to do a binary search for a version that makes everyone happy in the past. That's the version that's currently locked in go.mod. @aschmahmann can tell you more about it.

pkg/server/server.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+func GetEngine() (*engine.Engine, error) {
+	_engineOnce.Do(func() {
+		_engine, _engineErr = engine.NewDefaultEngine()
+	})
+	return _engine, _engineErr
+}
+
+type Server struct {
```
raulk: Can we document the HTTP API here? In the future we might want to introduce Swagger for API docs, especially if/as the API becomes richer. But for now, enumerating methods in the following form would suffice to get an API overview:

```
* POST /build: builds ...
```

nonsense:
I think this PR is mostly ready for review. It is hard to further cleanup APIs at this point, they seem good enough for the time being, I think it would be better to iterate with future functionalities.

I am not sure who exactly to mark for review, so feel free to remove yourself if you should not be a reviewer, and assign someone else.

I have tested this locally, and will now look into setting up a remote cluster and trying out Docker Swarm.

I am not happy that we have almost 0 tests, so probably a good idea to start adding some in the near future, so that we have a basic safety net. If you feel strongly that we should do that now before moving forward, let's have a chat.

pkg/server/describe.go
```diff
@@ -69,11 +69,19 @@ func (srv *Server) describeHandler(w http.ResponseWriter, r *http.Request, log *
 ----------
 `
 
-	tgw.Write([]byte(header))
+	_, err = tgw.Write([]byte(header))
+	if err != nil {
+		log.Errorw("write error", "err", err)
```
daviddias: Can you make it more descriptive? e.g. "which write?"

nonsense: I am not very familiar with the zap logger, but in development, every ERROR, WARN and FATAL show a stack trace, so you know exactly the line where all ERRORs are logged from. If the stack trace is not enabled once we run this in prod, then we should enable caller location (file:line) on all log line, and again we would know where a given log message is coming from.

You could argue that when we do that, we don't even have to add the verbose `write error` which doesn't add much value, but I think it makes it easier to search once you see the error coming up in log files. Still with (file:line) granularite `log.Error(err.Error())` might also be just enough information.

I hope this makes sense.

pkg/runner/cluster_swarm.go
```diff
@@ -162,8 +162,14 @@ func (*ClusterSwarmRunner) Run(input *api.RunInput) (*api.RunOutput, error) {
 			// if we are keeping the service, we must also keep the network.
 			return
 		}
-		if err := cli.NetworkRemove(ctx, networkID); err != nil {
```
nonsense: That's something I lost a bit of time trying to figure it out - if we don't explicitly define a `ctx` within the scope of this `defer` statement, we actually pick up the `global ctx` defined in the beginning of the function, which gets canceled by the time we call `cli.NetworkRemove`, so the `cli.NetworkRemove` is never successful - because we try to execute it with an already canceled context.

pkg/server/describe.go
nonsense: Nevertheless I see no harm in this being more descriptive, so - changed!

daviddias:
Suggestion: squash this PR commits and rebase master onto it again so that the changes that exist already on master don't appear here (hence, dropping that 1000 line change to something smaller).

nonsense:
> Suggestion: squash this PR commits and rebase master onto it again so that the changes that exist already on master don't appear here (hence, dropping that 1000 line change to something smaller).

I believe this is the correct number of changes, squashing won't help. Everything that you see in the Github diff is NOT on `master` as far as I can tell. I usually don't squash after opening a PR, so that if someone has reviewed up to a given commit, can continue to review from that commit on after comments have been addressed. The `force` push was done since I `rebased` on master twice due to conflicts with README.md.

Nevertheless, since I don't think anyone has spent much time reviewing already, will squash to one commit right now :)

daviddias:
Github diff shows changes on the main README.md (58 lines to be exact) and I assumed that there would be more. Maybe just those ones. Nvm then :)

README.md
```diff
@@ -66,22 +68,36 @@ To learn how this team works together read [HOW_WE_WORK](./docs/HOW_WE_WORK.md)
...
+
+If you plan to contribute code, make sure to install the [pre-commit](https://github.com/pre-commit/pre-commit) tool, which manages our pre-commit hooks for things like linters, go fmt, go vet, etc.
+
 [![](https://cdn.rawgit.com/jbenet/contribute-ipfs-gif/master/img/contribute.gif)](https://github.com/ipfs/community/blob/master/CONTRIBUTING.md)
 
 This repository falls under the IPFS [Code of Conduct](https://github.com/ipfs/community/blob/master/code-of-conduct.md).
 
 You can contact us on the freenode #ipfs-dev channel or attend one of our [weekly calls](https://github.com/ipfs/team-mgmt/issues/674).
 
+
```
daviddias: README updates are good 👍🏾

client/client.go
```diff
@@ -0,0 +1,164 @@
+package client
```
daviddias: Probably a nitpick from me, but I would prefer to have a `daemon` folder that has the RPC API and the Respective client rather than having two folders at the root.

This would follow the IPFS approach where there is a package for the "Core" of the service/implementation, and then daemon & clients are ways to interact with that Core.

client/client.go
```diff
@@ -0,0 +1,164 @@
...
+// close it.
+func (c *Client) Run(ctx context.Context, r *RunRequest) (io.ReadCloser, error) {
+	var body bytes.Buffer
+	err := json.NewEncoder(&body).Encode(r)
+	if err != nil {
+		return nil, err
+	}
+
+	return c.request(ctx, "POST", "/run", bytes.NewReader(body.Bytes()))
+}
```
daviddias: So, we always send JSON to the server and receive a streaming response of only bytes? Are these bytes base64 encoded and encoded in JSON as well?

I'm asking this as I'm considering future browser clients / dashboards etc. We've had to tackle many problems in JS land because Go often assumes that everything is a TCP socket and makes things super painful to deal with in a Browser or JS contexts (some notes from those wars can be found at https://github.com/ipfs/interface-js-ipfs-core/issues/81)

lanzafame:
> Decide whether .env.toml is required or optional. Gracefully handle config initialisation if .env.toml is missing is not there yet.
Solution: I think we should consider .env.toml to be optional for as long as we can. We should use meaningful defaults.

Was `.env.toml` being optional implemented? It is panicking for me currently.

lanzafame:
Some usage notes that may or may not need fixing in this PR:
- killing the daemon in the middle of a run returns the unhelpful error of just `EOF` to the client, would be ideal for the client to know that the daemon has died or at least suggest that that may be the cause of the error
- daemon doesn't have a logging flag, ideally the daemon pipes logs about what it is doing to journald for debugging

nonsense:
> > Decide whether .env.toml is required or optional. Gracefully handle config initialisation if .env.toml is missing is not there yet.
> > Solution: I think we should consider .env.toml to be optional for as long as we can. We should use meaningful defaults.
> 
> Was `.env.toml` being optional implemented? It is panicking for me currently.
Currently:
`.env.toml` is optional when `testground` is run as a client, then since there are no configs, an in-process daemon is start, and you can execute all the test runs locally.
`.env.toml` is required when `testground` is run as a daemon - it needs to know which address to listen to at the very least.

client/client.go
nonsense: We have two types of API endpoints - those that immediately return a response (for example `testground list`), and those that stream response back to the client (for example `testground build` and `testground run`).

Currently all requests/responses are JSON marshalled.

For the streaming endpoints, we use chunked HTTP responses and return JSON encoded `Msg`s (see PR description). I will review the linked issue, as I am not sure what could be a potential problem with this protocol.

I basically copied whatever Docker client/server was doing, because the model seemed like a good fit for `testground`.

nonsense:
> Some usage notes that may or may not need fixing in this PR:
> 
> * killing the daemon in the middle of a run returns the unhelpful error of just `EOF` to the client, would be ideal for the client to know that the daemon has died or at least suggest that that may be the cause of the error

The daemon should be stable and should not die in the middle of a run. If this happens then whoever supports the service should fix it by reviewing logs, etc. Pretty much I'd consider this as an internal server error. I agree that `EOF` is very unhelpful now, although if the daemon crashes, probably we can't do much better than returning `Something went wrong`.

> * daemon doesn't have a logging flag, ideally the daemon pipes logs about what it is doing to journald for debugging

Not sure what you mean with this. I think every command has a logging flag currently - we can run `testground -vv daemon` and pipe that to any logging system.

client/client.go
nonsense: You're right `client` should not have been top level - moved it to `pkg/daemon/client` as suggested.

README.md
nonsense: I don't consider the README complete, but wanted to at least add a structure where the top section describes `TestGround` to outsiders and has a good overview of the system.

What we had in `Description` read more of a history/background for the project, rather what `TestGround` actually is.

nonsense:
@lanzafame I updated a bit the error handling to provide additional context when something bad happens. The way I see it, we have at least two different types of errors, and I am not sure to what degree we want to handle them differently
1. Errors due to API call failing - for example if the daemon is killed as you suggested - I wrapped those with `fatal error from daemon`.
2. Errors due to something failing on the daemon side that we want to propagate to the client - for example daemon losing connection with its docker daemon, or build/run failing, etc. - Those are propagated as `Msg`s with `type=error` and written to the response with `errf`.

I hope that makes sense. Let me know if you have ideas on how to improve this :)

raulk:
I'm reviewing.

pkg/runner/cluster_swarm.go
```diff
@@ -162,8 +162,14 @@ func (*ClusterSwarmRunner) Run(input *api.RunInput) (*api.RunOutput, error) {
...
-			log.Errorw("removing network", "name", sname)
+		err = retry(5, 1*time.Second, func() error {
```
raulk: Are we sure that this is fundamentally different than setting a context deadline of 30 seconds and doing a single call? Is there a bug in Docker or something?

pkg/runner/cluster_swarm.go
```diff
@@ -348,3 +354,19 @@ func (*ClusterSwarmRunner) ConfigType() reflect.Type {
...
+			return
+		}
+
+		if i >= (attempts - 1) {
+			break
+		}
+
+		time.Sleep(sleep)
+	}
+	return fmt.Errorf("after %d attempts, last error: %s", attempts, err)
```
raulk:
```suggestion
	return fmt.Errorf("failed after %d attempts, last error: %w", attempts, err)
```

raulk: (to wrap the error, thanks go1.13)

README.md
```diff
@@ -5,9 +5,24 @@
 
 > ⚠️ **Heavy WIP.** beware of the Dragons 🐉..
 
-> **This repository is incubating the Testground. 🐣**
+> **This repository is incubating the TestGround. 🐣**
```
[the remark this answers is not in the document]
nonsense: I have to agree I also prefer `Testground` to `TestGround` - much easier to type.

cmd/describe.go
```diff
@@ -1,71 +1,45 @@
...
-	ArgsUsage: "[term], where " + termExplanation,
+	ArgsUsage: "[term], where " + server.TermExplanation,
 	Action:    describeCommand,
 }
 
-func describeCommand(c *cli.Context) error {
-	if c.NArg() == 0 {
-		_ = cli.ShowSubcommandHelp(c)
-		return errors.New("missing term to describe; " + termExplanation)
+func describeCommand(ctx *cli.Context) error {
```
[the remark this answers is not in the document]
nonsense: Fixed it. Sorry - I changed to `ctx` because at some point I had a `var c` that was better suited for `c`, but then apparently I refactored it out.

pkg/runner/cluster_swarm.go
nonsense: I think `docker` was returning from the previous command while the `network` was still in use, so I think this is different to setting ctx deadline to 30sec. Then again it could have been triggered to the bug with the global ctx being picked up in the closure - I am not sure.

Will add a note to review once again, and will remove the `retry` in future PR if it is indeed not needed.

pkg/runner/cluster_swarm.go
nonsense: Agree, it would be good to adopt the new errors package, but we wrap errors at many places, so will leave this for a future PR.

nonsense:
Thanks for the reviews everyone! I know this is lacking tests, so I will make sure I add a few in the near future so that we have a better safety net around testground client and testground daemon commands.

pkg/runner/cluster_swarm.go
raulk: No need to adopt the errors package, it's part of `fmt.Errorf()` now: https://golang.org/pkg/fmt/#Errorf

PR #190, opened 2019-11-27
docs: fix dead links

hacdias:
Just looking around and found some dead links.


daviddias:
Thank you @hacdias :)

PR #198, opened 2019-12-02
feat: add support for array of strings and json parameters

### LINKED ISSUE #195: Pass Arrays and Objects to Test Parameters
I'm now taking a look at #58 ("Chewing Large Datasets" Plan) and I want to make it better! @daviddias set some ideas for the [test parameters](https://github.com/ipfs/testground/tree/test-plan-1/plans/chew-large-datasets#test-ipfs-add-defaults):

> - `File Sizes` - An array of File Sizes to be tested (default to: [1MB, 1GB, 10GB, 100GB, 1TB])
> - `Directory Depth` - An Array containing objects that describe how deep/nested a directory goes and the size of files that can be found throughout (default to [{depth: 10, size: 1MB}, {depth: 100, size: 1MB}]

As far as I understood right now, we're passing the parameters with the flag `--test-param key=value`, right? If so, how will we pass arrays and other types of variables. I can work on that, I just to come to an agreement on how to do it 😃 

## Way 1

We could pass arrays like this:

```
--test-param arrKey=value1 \
--test-param arrKey=value2 \
--test-param arrKey=value3
```

It's easy to parse and we can easily define arrays. However, this wouldn't solve how to pass custom objects.

## Way 2

Just pass a JSON-like object:

```
--test-param array=[1MB, 1GB, 10GB, 100GB, 1TB]
```

Then, we could call `runenv.StringArrayParam` or `runenv.StringSliceParam`. The same for other basic data types that we already have right now. This would work for slices.

We could also add a `runenv.JsonParam` that would unmarshal to a specified data type (taken as an argument). This would work for slices and objs.

## Way 3

Just implement `runenv.JsonParam` from Way 2.

/cc @daviddias @raulk

hacdias:
This adds support for two new different types of test parameters and fixes #195.

1. Array of Strings, through the functions:
	- `func StringArrayParam(name string) (a []string, ok bool)`
	- `func StringArrayParamD(name string, def []string) []string`
2. And arbitrary JSON structures:
	- `func JSONParam(name string, v interface{}) (ok bool)`

---

In reality, `StringArrayParam` and `StringArrayParamD` are just wrappers for `JSONParam`. 

I didn't add `IntArrayParam` nor `BoolArrayParam` because I think they are more uncommon. However, I'm open to see what you've to say!

License: MIT

### DIFF
--- a/sdk/runtime/runenv.go
+++ b/sdk/runtime/runenv.go
@@ -287,7 +287,6 @@ func (re *RunEnv) JSONParam(name string, v interface{}) (ok bool) {
 	}
 
 	if err := json.Unmarshal([]byte(s), v); err != nil {
-		fmt.Println(err)
 		return false
 	}
 
--- a/sdk/runtime/runenv.go
+++ b/sdk/runtime/runenv.go
@@ -287,6 +287,7 @@ func (re *RunEnv) JSONParam(name string, v interface{}) (ok bool) {
 	}
 
 	if err := json.Unmarshal([]byte(s), v); err != nil {
+		fmt.Println(err)
 		return false
 	}
 
--- a/sdk/runtime/runenv.go
+++ b/sdk/runtime/runenv.go
@@ -2,6 +2,7 @@ package runtime
 
 import (
 	"crypto/sha1"
+	"encoding/json"
 	"fmt"
 	"math/rand"
 	"os"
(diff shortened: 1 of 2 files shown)


hacdias:
Done @daviddias!

sdk/runtime/runenv.go
```diff
@@ -258,6 +259,41 @@ func (re *RunEnv) BooleanParamD(name string, def bool) bool {
...
+// JSONParam unmarshals a JSON parameter in an arbitrary interface.
+// It returns if the parameter was (un)set or (in)valid.
+func (re *RunEnv) JSONParam(name string, v interface{}) (ok bool) {
+	s, ok := re.TestInstanceParams[name]
+	if !ok {
+		return false
+	}
+
+	if err := json.Unmarshal([]byte(s), v); err != nil {
+		fmt.Println(err)
```
nonsense: I think we shouldn't print errors to stdout, but rather return them to the caller to decide what to do with them, or just ignore them, depending on circumstances.

hacdias: Completely missed that! Thanks for noticing. The plan is to ignore them in this case.

hacdias:
For how many reviews should I wait before merging PRs?

nonsense:
I think two reviewers should be fine here.

Issue #196, opened 2019-12-02
Unblock the testing for go-ipfs 0.5.0 release


daviddias:
The primary goal for the TestGround MVP is to unblock the testing for go-ipfs 0.5.0 release. To achieve that goal, it was instructed by @Stebalien as the IPFS Tech Lead that the requirement is to test the "Stabilize the DHT" branch(es) of go-libp2p-kad-dht. This requirement was identified during the  Co-Location Hack Week of early November.

The current state of TestGround and where we are headed can be consulted at [State of TestGround, mid November Q4 2020](https://docs.google.com/document/d/1lpifz6CSEYhas1a3ZpgCcDWAFl9Ysmr7o_JH0n8Vny0/edit#) report.

### In tl;dr; what we need to deliver on the goal is:

- [ ] Expand the TestGround Infrastructure so that it can run multiple thousands of nodes (owner: @nonsense)
- [x] Add support to Traffic Shaping & Network Configuration to TestGround (owner: @Stebalien)
- [ ] Instrument go-libp2p so that the DHT operations are observable and traced in a Distributed Context (owner: @raulk)
- [ ] Creation of a TestGround Dashboard that can aid the developers of go-ipfs & go-libp2p to grok the results outputted by every node in a TestGround run (owner: @daviddias)
- [ ] Keep on iterating the DHT Test Plan (owner: @aschmahmann )

### Other current active endeavors, not on the critical path are:

- [x] Daemonizing the TestGround (owner: @nonsense)
- [ ] Creation of the following Test Plans
  - [ ] [Bitswap Plan (owner: @dirkmc)](https://github.com/ipfs/testground/issues/100)
  - [ ] [Naming Plan (owner: @aschmahmann)](https://github.com/ipfs/testground/issues/99)
  - [ ] [Message Delivery Plan (owner: @vyzo)](https://github.com/ipfs/testground/issues/98)
  - [ ] [Datasets Plan (owner: @hacdias)](https://github.com/ipfs/testground/pull/58)
  - [ ] [Interop Plan (owner: @achingbrain)](https://github.com/ipfs/testground/issues/138)

momack2:
I think there is more than this, no? @Stebalien mentioned needing to run multiple different go-ipfs versions within a test plan to ensure the "partial upgrade" state works as well.

[...]

PR #201, opened 2019-12-03
Design document for test ensembles

raulk:
Needed for #196.


docs/ENSEMBLES.md
```diff
@@ -0,0 +1,57 @@
...
+2. Running 50 instances of each build artifact, enlisting them in the same run
+   ID, providing total instance count=200. This results in all instances
+   attaching to the same sync subtree, and awaiting for 200 instances total to
+   appear.
+
+As the test scenario runs, each instance outputs metrics and assets. The
+challenge now becomes how do we differentiate the metrics generated by each
+dependency graph / test parameter set.
+
+In practice, we want to answer questions like: "how many dials per second did
```
nonsense: Just to be clear, when you say `how many dials per second did version X generate`, you mean a specific library version, not a specific group, right?

1. If we care about stats from a given Group, we just need to tag the binary/docker image and sent stats with a given tag.
2. If we care about stats from `go-libp2p-swarm/v0.3.0`, then we need to filter for results from both Group A and Group C, which is indeed more complicated than 1. and we need the dep graph tracking, as described below.

docs/ENSEMBLES.md
```diff
@@ -0,0 +1,57 @@
...
+Testground has been designed to offer the primitives to compose such **test
+ensembles**. We coin the term **test ensemble** to refer to a set of test
+instances, potentially built against different dependency graphs, potentially
+running with different parameters, but orchestrated by the same runner, and
+participating under the same *run id*.
+
+Thus, all instances comprising the ensemble have visibility over one another,
+yet they are configured differently, and such variance is totally opaque.
+
+As long as the DHT plan can be compiled against all upstream versions of the
```
yusefnapora: Sorry if this is off topic, but this reminds me of a conversation with @aschmahmann and @daviddias during the testground unconf about how to run a test between versions that aren't fully API compatible.

For example, if I add a new public method to some type in version 0.5 that I'd like to use in a test plan, how can I test the protocol-level interactions with version 0.4 that doesn't have the new method? Since the plan depends on a method that doesn't exist in 0.4, we can't just swap out the dependencies.

I think that @daviddias was recommending that you write a shim of some kind which would stub out the new method when linked against the old version but would call the real implementation if it exists. Does anyone have thoughts on how to write that shim layer? In a language like Go it seems like the shim would have to use reflection to check for the existence of the real method. Is that the right way to do this?

FWIW, I was initially thinking I'd need to do this to test a (future) DHT with signed address records against the current version. But after thinking it over, I may just be able to add some instrumentation to the new version, so the test plan wouldn't have to call any new methods & instead could just collect the new metrics. Still, it seems like something we'll probably want to do eventually, so I figure it's worth asking.

raulk: Reflection or build tags would be the way to go IMO. But this doesn't work always. Some behaviours cannot be shimmed -- they are novel features.

For the particular case of comparing how a DHT or pubsub performs with signed peer routing records vs. unsigned records, there shouldn't really be any public interface changes to deal with.

docs/ENSEMBLES.md
raulk: I'm talking about concrete upstream dependency versions. IMO groups are a higher-level construct. We'll want to record which groups ran with a particular test run, but when it comes to a comparative analysis, I'm not entirely sure they'll provide much value. The kind of question I'd like to answer is: how did go-libp2p-swarm/v0.2.0 perform against 0.1.0?

raulk:
Ensembles were superseded by Compositions, and the latter are documented here: https://github.com/ipfs/testground/blob/master/docs/COMPOSITIONS.md

PR #204, opened 2019-12-03
feat: track dependency tree

### LINKED ISSUE #147: Dependency graph tracking
When we produce a build, we want to capture the entire dependency graph in effect and store it in the state database. This will enable us to:

* search for runs that tested `go-libp2p-kad-dht` commit `7aa5b0ef` (for example). 
* easily generate performance charts where module `X` was in the runtime across all revisions of module `X`, contextualised to a specific plan `A`.
* dashboard UI: find a build to set as a comparative baseline.

**Capturing this data is responsibility of the builder.**

The implementation language of a test plan (e.g. Go, Javascript, etc.) will have its own tooling to calculate a dependency graph. For Go, we can use `go mod graph`.

In builder `exec:go`, this is as easy as running `go list -m all` as an `exec.Command()` and processing the output.

In builder `docker:go`, my suggestion is to add a `RUN` in the `Dockerfile` that dumps the dependency graph in a demarcated block so the builder can slice this output and process it as per above:

```
RUN echo "-----BEGIN GO MOD LIST-----" && go list -m all && echo "-----END GO MOD LIST-----"
```

The dependency graph needs to be sent off to the run state database, which will be specified in another issue.

hacdias:
This closes #147. Adds support for tracking the dependency graph changes by adding a new `Dependencies` variable into `BuildOutput`. It contains a map in the format module=>version. It allows for easy checks of versions changes and/or if a module was added/removed.

Note that in Docker mode, if the image is cached, nothing will be printed, and hence the dependencies map will be empty. Not sure if it's supposed to be this way or not, so I'll wait for your feedback!

License: MIT

### DIFF
--- a/pkg/api/builder.go
+++ b/pkg/api/builder.go
@@ -49,4 +49,6 @@ type BuildOutput struct {
 	// ArtifactPath can be the docker image ID, a file location, etc. of the
 	// resulting artifact. It is builder-dependent.
 	ArtifactPath string
+	// Dependencies is a map of modules (as keys) to versions (as values).
+	Dependencies map[string]string
 }
--- a/pkg/build/golang/Dockerfile.template
+++ b/pkg/build/golang/Dockerfile.template
@@ -22,7 +22,7 @@ ENV TESTPLAN_EXEC_PKG ${TESTPLAN_EXEC_PKG}
 ENV PLAN_DIR /plan/
 
 # Optionally install IPFS
-RUN if [ -n "${GO_IPFS_VERSION}" ]; then echo Install IPFS ${GO_IPFS_VERSION} && cd /tmp && wget https://dist.ipfs.io/go-ipfs/v${GO_IPFS_VERSION}/go-ipfs_v${GO_IPFS_VERSION}_linux-amd64.tar.gz && tar xf go-ipfs_v${GO_IPFS_VERSION}_linux-amd64.tar.gz; fi 
+RUN if [ -n "${GO_IPFS_VERSION}" ]; then echo Install IPFS ${GO_IPFS_VERSION} && cd /tmp && wget https://dist.ipfs.io/go-ipfs/v${GO_IPFS_VERSION}/go-ipfs_v${GO_IPFS_VERSION}_linux-amd64.tar.gz && tar xf go-ipfs_v${GO_IPFS_VERSION}_linux-amd64.tar.gz; fi
 RUN touch /tmp/delete_me
 
 # Copy only go.mod files and download deps, in order to leverage Docker caching.
@@ -44,6 +44,10 @@ RUN cd ${PLAN_DIR} \
     && go env -w GOPROXY="${GO_PROXY}" \
     && CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o testplan ${TESTPLAN_EXEC_PKG}
 
+# Store module dependencies
+RUN cd ${PLAN_DIR} \
+  && go list -m all > ./testground_dep_list
+
 #:::
 #::: RUNTIME CONTAINER
 #:::
@@ -51,6 +55,7 @@ RUN cd ${PLAN_DIR} \
 #FROM busybox:1.31.0-glibc # Doesn't work with prebuilt IPFS binaries
 FROM debian
 
+COPY --from=0 /plan/testground_dep_list /
 COPY --from=0 /plan/testplan /
 RUN mkdir -p /usr/local/bin
 COPY --from=0 /tmp/delete_me /tmp/go-ipfs/ipfs* /usr/local/bin/
--- a/pkg/build/golang/Dockerfile.template
+++ b/pkg/build/golang/Dockerfile.template
@@ -46,7 +46,7 @@ RUN cd ${PLAN_DIR} \
 
 # Store module dependencies
 RUN cd ${PLAN_DIR} \
-  && go list -m all > ./testground_dep_list
+  && go list -m all > /testground_dep_list
 
 #:::
 #::: RUNTIME CONTAINER
@@ -55,7 +55,7 @@ RUN cd ${PLAN_DIR} \
 #FROM busybox:1.31.0-glibc # Doesn't work with prebuilt IPFS binaries
 FROM debian
 
-COPY --from=0 /plan/testground_dep_list /
+COPY --from=0 /testground_dep_list /
 COPY --from=0 /plan/testplan /
 RUN mkdir -p /usr/local/bin
 COPY --from=0 /tmp/delete_me /tmp/go-ipfs/ipfs* /usr/local/bin/
--- a/pkg/build/golang/deps.go
+++ b/pkg/build/golang/deps.go
@@ -1,0 +1,87 @@
+package golang
+
+import (
+	"bytes"
+	"context"
+	"strings"
+	"time"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/client"
+)
+
+func parseDependencies(raw string) map[string]string {
+	rawModules := strings.Split(raw, "\n")
+	modules := map[string]string{}
+
+	for _, module := range rawModules {
+		module = strings.TrimSpace(module)
+		if module == "" {
+			continue
+		}
+
+		parts := strings.Split(module, " ")
+		path := strings.TrimSpace(parts[0])
+		version := ""
+
+		if len(parts) == 2 {
+			version = strings.TrimSpace(parts[1])
+		}
+
+		modules[path] = version
+	}
+
+	return modules
+}
+
+func parseDependenciesFromDocker(cli *client.Client, imageID string) (map[string]string, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+	defer cancel()
+
+	ccfg := &container.Config{
+		Image: imageID,
+		Tty:   true,
+		Entrypoint: []string{
+			"cat",
+			"/testground_dep_list",
+		},
+	}
+
+	// Create container
+	res, err := cli.ContainerCreate(ctx, ccfg, nil, nil, "")
+	if err != nil {
+		return nil, err
+	}
+
+	// Start container
+	err = cli.ContainerStart(ctx, res.ID, types.ContainerStartOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	// Get logs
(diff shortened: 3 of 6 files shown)


nonsense:
`Note that in Docker mode, if the image is cached, nothing will be printed, and hence the dependencies map will be empty. Not sure if it's supposed to be this way or not, so I'll wait for your feedback!` - I think this is a problem - we need the dependencies to be reliable and always present, no matter if the part or all of the build was cached.

Since the output is missing when an image is cached, maybe this is not the right way to grab the dependencies - why don't we store them in a file within the docker image, and not just `cat` the file from that image when we need the dependencies?

hacdias:
@nonsense possible, and perhaps easier! I will try it out! Thanks for the idea.

raulk:
Exactly what I was thinking, @nonsense!

pkg/util/docker.go
```diff
@@ -3,22 +3,25 @@ package util
...
 )
 
-// PipeDockerOutput pipes a reader that spits out jsonmessage structs into a
-// writer, usually stdout. It returns normally when the reader is exhausted, or
-// in error if one occurs.
-func PipeDockerOutput(r io.ReadCloser, w io.Writer) error {
+// PipeDockerOutput pipes a reader that spits out jsonmessage structs into one
+// or many writer, usually stdout. It returns normally when the reader is exhausted,
+// or in error if one occurs.
+func PipeDockerOutput(r io.ReadCloser, writers ...io.Writer) error {
```
raulk: I'd say accept a single `Writer`, and have the caller pass a [`MultiWriter`](https://golang.org/pkg/io/#MultiWriter) if they want to tee outputs.

nonsense:
@hacdias just heads up that this PR will probably conflict with https://github.com/ipfs/testground/pull/183 - whoever gets their first, the other one will have to rebase ;)

pkg/util/docker.go
```diff
@@ -3,22 +3,25 @@ package util
...
 	var msg jsonmessage.JSONMessage
 Loop:
 	for dec := json.NewDecoder(r); ; {
 		switch err := dec.Decode(&msg); err {
 		case nil:
-			_ = msg.Display(w, true)
-			if msg.Error != nil {
-				return msg.Error
+			for _, w := range writers {
+				_ = msg.Display(w, w == os.Stdout || w == os.Stderr)
```
raulk: Ah, I see that you need some fine-grained control here.

pkg/build/golang/utils.go
```diff
@@ -0,0 +1,45 @@
...
+			version = strings.TrimSpace(parts[1])
+		}
+
+		modules[path] = version
+	}
+
+	return modules
+}
+
+func parseDependenciesFromDocker(raw string) map[string]string {
```
raulk: With piping the dependency graph into a file inside the image, all of this disappears. Instead I'd create a container (without starting it), and I'd use `docker cp` (API equivalent) to copy the dependency file out. Then delete the container.

nonsense: Or just run a container with `--rm` and override the entrypoint to `cat`... this way no need to `docker cp`, and it'd just be one command. Just an idea.

pkg/util/docker.go
hacdias: Not needed anymore since I don't need to use this function anymore. We can change it in the future if needed!

pkg/util/docker.go
hacdias: Same as previous comment!

pkg/build/golang/utils.go
hacdias: I basically went with @nonsense's idea.

hacdias:
@nonsense @raulk I modified this according to your comments. It now starts a container afterwards, cats the dependency list, and cleans it up.

raulk:
> @hacdias just heads up that this PR will probably conflict with #183 - whoever gets their first, the other one will have to rebase ;)

I think it might be easier to merge the daemon first, then have @hacdias rebase his PR. The daemon PR essentially moves around code, so I suspect it will be easier for @hacdias to find the right spots in the new code, than for @nonsense to have to rebase the whole thing on top of these small changes.

pkg/build/golang/utils.go
raulk: As you prefer. The reason I suggested a `docker cp` is because the Docker API doesn't have a single command for `docker run`. You actually have to call create, **then** start, and actually process the output. It's clunky. And starting a container implies a higher latency.

Since you _anyway_ have to create the container, you can just extract the file without starting it and doing all the above dance.

hacdias:
Agreed!

hacdias:
@raulk @nonsense just rebased this and there were no issues! :)

raulk:
I still think I have a point with https://github.com/ipfs/testground/pull/204#discussion_r354357844

pkg/build/golang/utils.go
hacdias: @raulk missed this comment! Yes, I will try to do that to see if it's faster (it must be) :)

hacdias:
@raulk @nonsense done!

pkg/api/builder.go
```diff
@@ -49,4 +49,6 @@ type BuildOutput struct {
 	// ArtifactPath can be the docker image ID, a file location, etc. of the
 	// resulting artifact. It is builder-dependent.
 	ArtifactPath string
+	// Dependencies is a map of modules (as keys) to versions (as values).
```
raulk: Let's document what the map actually contains.

```suggestion
	// Dependencies is a map of modules (as keys) to versions (as values),
	// containing the collapsed transitive upstream dependency set of this
	// build.
```

pkg/build/golang/Dockerfile.template
```diff
@@ -44,13 +44,18 @@ RUN cd ${PLAN_DIR} \
     && go env -w GOPROXY="${GO_PROXY}" \
     && CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o testplan ${TESTPLAN_EXEC_PKG}
 
+# Store module dependencies
+RUN cd ${PLAN_DIR} \
+  && go list -m all > ./testground_dep_list
```
raulk: I'd prefer if we store it in the root (/), outside the plan's source code, to avoid unnecessary mixing.

pkg/build/golang/deps.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil, err
+	}
+
+	deps, err := ioutil.ReadFile(path.Join(dir, "testground_dep_list"))
+	if err != nil {
+		return nil, err
+	}
+
+	// Remove container
+	err = cli.ContainerRemove(context.Background(), res.ID, types.ContainerRemoveOptions{Force: true})
```
raulk: If something fails above, the container will not be removed. You can schedule this cleanup in a `defer` statement after the container is created.

pkg/build/golang/deps.go
```diff
@@ -48,6 +49,14 @@ func parseDependenciesFromDocker(cli *client.Client, imageID string) (map[string
 		return nil, err
 	}
 
+	defer func() {
+		// Remove container
+		err = cli.ContainerRemove(context.Background(), res.ID, types.ContainerRemoveOptions{Force: true})
+		if err != nil {
+			fmt.Printf("error while removing container %s: %v", res.ID, err)
```
hacdias: @raulk @nonsense when deferring, what's the best practice to check for errors? I decided to just print out a message about it. I could also replace the return value in case there is an error here too. Wdyt?

pkg/build/golang/deps.go
```diff
@@ -0,0 +1,83 @@
...
+	"strings"
+	"time"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/client"
+	"github.com/docker/docker/pkg/archive"
+)
+
+func parseDependencies(raw string) map[string]string {
```
nonsense: This is a good candidate for a unit test with a few test cases. For example I see we have a version where version stays empty, but I am not sure this can be the case with `go mod`. I believe you always get a version, even if it is pseudo-version.

pkg/build/golang/deps.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+
+		modules[path] = version
+	}
+
+	return modules
+}
+
+func parseDependenciesFromDocker(cli *client.Client, imageID string) (map[string]string, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
```
nonsense: We have a `ctx` one level up - maybe worth including `ctx` as an argument to this function, so that all container actions here get canceled if the ctx in the argument gets canceled.

pkg/build/golang/deps.go
nonsense: I think in this case, it's best for the return value to be replaced.

When you want to `handle` an error (which half of the times actually means just logging it, then better to use a logger, rather than fmt.Printf, because the logger is structured, and has additional information - timestamp, optional context, caller location, etc.)

nonsense: Actually on a second thought about this particular case - this is the `cleanup` phase. If nothing returned an error until the defer statement, then we get the dependencies correctly. Then if the cleanup phase fails for some reason, we can also consider it just as a warning, and move on (although something might fail in the future).

Therefore I think it is just as valid to just log the error, and not replace the returned value - whatever you decide.

Another point is that if something in the function does return an error and then you have another error in the cleanup phase - you would overwrite the first error.

hacdias: @nonsense yes, I completely agree with you about the error overwriting and that's why I didn't replace it on the first turn. I updated it to be a  `log.Warnf`. It doesn't worry me much if the cleanup fails even if something Docker-related might fail in the future, we will get the error afterwards anyways.

pkg/build/golang/deps.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+
+		modules[path] = version
+	}
+
+	return modules
+}
+
+func parseDependenciesFromDocker(ctx context.Context, log *zap.SugaredLogger, cli *client.Client, imageID string) (map[string]string, error) {
+	// Create container
```
nonsense: Nitpick, but these comments are redundant IMO - it is pretty clear from the function call below. Same goes for `Remove container`, and `Copy file from container` and `Unpack file`.

hacdias:
@raulk @nonsense I'm merging this. I added three small unit tests and addresses all the issues you've mentioned!

PR #206, opened 2019-12-04
docs: link to how we work

daviddias:


CONTRIBUTING.md
```diff
@@ -4,4 +4,20 @@ IPFS as a project and community follows the [standard IPFS Community contributin
...
 We appreciate your time and attention for going over these. Please open an issue on [ipfs/community](https://github.com/ipfs/community) if you have any question.
 
+## Contributing Code
+
+If you plan to contribute code, make sure to install the [pre-commit](https://github.com/pre-commit/pre-commit) tool, which manages our pre-commit hooks for things like linters, go fmt, go vet, etc.
+
+We provide a `Makefile` rule to facilitate the setup:
+
+```sh
+$ make pre-commit
```
nonsense: Just a nitpick - I find it a bit weird that we require people to install python and pip and pre-commit for contributing.

I think we should just explain that code needs to be go fmtted, go vetted, etc. - something that is tested by the CI and common sense in the Go community. If you submit code that is not go fmtted, then CI will fail, it is normal to not accept PRs that fail the CI runs, and sometimes not even bother to review them prior to CI passing.

daviddias: This was an addition back from Hack Week Tenerife. I've been mixed feelings about it as well (as most people will just do `git commit -m "" -n` and skip the hooks anyway) but not looking to bikeshed too much on it. If you feel it is an important nitpick, could you open an issue ?

nonsense: Just sharing my experience as a user - I never ran this, I tried to install it and it failed and I didn't bother again. It seems we just add more text and rules that ultimately don't matter to our docs. I like opinionated and meaningful defaults in general as long as they are adopted by the community in general.

I like docs that are concise and to the point, without unnecessary information.

I don't have hard feelings, but if someone else on the team shares the opinion, maybe we should remove it ;)

Issue #38, opened 2019-09-16
Acquire a domain name


jimpick:
It would be nice to have a URL where we could host things like:

* dashboard
* grafana
* blog for infra test updates
* sub-namespaces for specific teams / endeavours to post results

Keep in mind that we might want to share resources between IPFS, libp2p, and possibly even Filecoin and community projects / collaborations down the road.

jimpick:
https://benchmarks.ipfs.team/ is currently in use as a place to publish benchmark results. (appears to be broken right now though)

daviddias:
Why not just `testground.ipfs.team` for now? I'm able to update the DNS record, just let me know where you want me to point it at.

jimpick:
`testground.ipfs.team` works for now, I think. 

I'm thinking we could publish a static site at that address (built on CI like our other sites).

daviddias:
@olizilla can we haz testground.ipfs.team? We won't be using right away but want to squat it if available.

olizilla:
yas. I was in the dns anyhoo, so `testground.ipfs.team` is now a thing.

daviddias:
Wohooo! Thank you for being super fast. Can you make it redirect to https://github.com/ipfs/testground for the time being?

nonsense:
testground.ipfs.team has an A record pointing to an S3 bucket, which does a redirect to https://github.com/ipfs/testground

Not sure if worth spending more effort to get the https before we actually have a proper website.

Marking this as complete for now, feel free to re-open if you think we need to do more now.

PR #208, opened 2019-12-04
create route53 hosted zone for testground.ipfs.team

nonsense:
This PR is adding Terraform playbooks to manage a hosted zone for `testground.ipfs.team` in AWS. I have already applied the playbooks, so the backend has been created in S3 and the hosted zone exists.

We need to add docs to point to the correct account on AWS where this is hosted, but for now this information is implicit, until we have some valuable records there (for example our website, or record for the production TestGround service, or something else.)

Next steps:
- [x] Update `ipfs.team` domain to point the `testground` subdomain to AWS Route53.

### DIFF
--- a/infra/dns/README.md
+++ b/infra/dns/README.md
@@ -1,6 +1,6 @@
-# Terraform: TestGround DNS
+# Terraform: Testground DNS
 
-A list of DNS zones used by the TestGround team:
+A list of DNS zones used by the Testground team:
 
 | Cloud | DNS Zone                                     |
 | ----- | -------------------------------------------- |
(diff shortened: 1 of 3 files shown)


nonsense:
NS records for `testground.ipfs.team` have been updated, and these Terraform playbooks have already been applied, so I think we could merge this.

nonsense:
@jimpick added you as reviewer here as well, just to showcase that this is also PL-specific. @raulk @daviddias what do you think about such changes - see discussion about PL-specific code at https://github.com/ipfs/testground/pull/210#discussion_r354504666

nonsense:
I also added an S3 bucket and redirect to github.com/ipfs/testground as per #38

This works only for HTTP.

Bonus points if someone wants to modify this for HTTPS (AFAIK we have to go through CloudFront and also get a certificate for testground.ipfs.team , I don't think it is possible only via AWS S3)

Issue #176, opened 2019-11-19
sidecar: Remove the dependency on docker hub


raulk:
Instead, introduce a `Makefile` that builds both the testground and the Docker image. Make the `local:docker` runner use the latest locally tagged image (don't pull from anywhere).

raulk:
This is fixed. We don't pull the image from Docker Hub any longer, but the user still has to manually rebuild it when they change the sidecar logic, which sucks a bit, but I can't think of ways to streamline that.

PR #209, opened 2019-12-05
Add Makefile to build ipfs/testground image (don't pull it)

jimpick:
Run `make docker-ipfs-testground` to build the docker image locally.

I've modified EnsureContainer() to take a PullImageIfMissing boolean
option - it is set to false for the ipfs/testground image, so it
will not automatically pull it from Docker Hub.

If the image does not exist, it displays the following error:

> Docker image ipfs/testground not found, run `make docker-ipfs-testground`


jimpick:
For #176

PR #210, opened 2019-12-05
Testground infrastructure playbooks

nonsense:
This PR is a continuation of https://github.com/ipfs/testground/pull/175 as well as refactoring some of the existing playbooks and scripts. It introduces the following changes to the TestGround infrastructure and addresses comments from the previous PR:

Terraform:
- [x] Replace `terraform-aws-modules/autoscaling/aws` module with `aws_instance` - since we don't want the autoscaling feature, it is simpler to just have a number of instances which we can then `stop` and `start` on-demand.
- [x] Added `terraform backend` so that state of infrastructure is kept on AWS S3. This is necessary when multiple people will concurrently work on the TestGround as a Service.
- [x] Removed static allocation of private IPs
- [x] Removed `mosh` from security group rules
- [x] Removed `EC2-Full-API-Access-Dangerous` IAM instance profile from instances
- [x] ~Added default AWS key pair, so that we don't manually manage keys in AWS. This key is used only to create infrastructure, and once Ansible playbooks are run, they don't have to (and currently don't) keep it in `authorized_keys`, so it is not important which key is used, as long as the person running the playbooks has access to it. I suggest we share this key via 1Password. It is referenced from the Ansible playbooks from `ansible.cfg` as default.~
- [x] Locked version of AWS Terraform provider

Terraform infrastructure <-> Ansible inventory:
- [x] Used https://github.com/adammck/terraform-inventory to generate Ansible inventory from Terraform state. This way we have nice groups of hosts in Ansible as soon as infrastructure is generated by Terraform.

Ansible:
- [x] Added a few default and external roles, namely: `bootstrap`, `setup`, `docker`, `logrotate`. These need to be installed prior to running the main playbook, or we could vendor them. I suggest once the PR is reviewed/approved, we vendor them in case they disappear from Github (highly unlikely, but better safe than sorry).
- [x] Added roles for installing Docker Swarm and initialising `manager` and `worker` nodes.
- [x] Added external role for installing Redis - this is cleanly separated from everything else, so we can easily decide to change the way we run Redis if needed. Not sure we actually need this, as Redis is currently running in a container on Docker Swarm on the worker node marked as `TGRole=redis`. There is no data container, and no backup or upgrade strategy.
---
As a result of some of these changes we:
- Manage most software through Ansible, and not Ansible and Packer.
- Don't need special permissions for `aws describe-instances` and `http://169.254.169.254/`
- Don't need Apache and CGI scripts
- Don't need to generate SSH keys on remote hosts, or to install Ansible on remote hosts.

OPEN QUESTIONS:
- [ ] Why is Redis virtual machine part of the Docker Swarm when we don't want to schedule test plans on it? AFAICT Redis is just a dependency that we can run anywhere we see fit. Is there a requirement for Redis to be part of the swarm? I guess we are just piggy-backing on the docker service discovery, or?
- [x] Secrets management - currently `filebeat` config contains secret config for the sink. How do we want to handle that?
a) Plaintext file on AWS S3.
b) Keybase FS
c) Ansible Vault
~~d) AWS Secrets Manager (has a module in Ansible) - it is very expensive, so I vote `no` on this.~~
**e) Use 1Password - this is the solution I have picked for now. The dev running the Ansible playbooks, needs to have access to the respective secrets on 1Password and must be logged in prior to running the playbooks. This is described in the Ansible README.**
- [ ] Authentication to `testground daemon` on a remote instance - since we will already get it on one of these instances, we could either open a port on the security group (i.e. no authentication), or assume that users setup SSH tunnels to it (for this we'd need user's public keys).

TODO:
- [x] Add labels to some docker swarm workers (or all), because this is currently required by `testground`
- [x] Add and configure `filebeat`.
- [x] Continue the refactor and remove most (or all) from `scripts` folder.
- [x] Remove unnecessary config files from `files`.
- [x] Remove unnecessary config files from `templates`.
- [x] Update docs and maybe consolidate everything under one `infra/README.md` rather than split between Ansible and Terraform READMEs.
- [x] ~Add default `ssh` `testground` key to 1Password so that devs creating infra could use it?~
- [x] ~Collect ssh pubkeys from team members - @jimpick @daviddias @raulk @hacdias anyone else?~
- [ ] Run `testground` on remote Docker Swarm cluster and make sure it is discovering correctly `redis` and that `filebeat` is forwarding logs.
- [ ] Make sure playbooks work with different instance types (`hvm instance store`, `ebs`, etc.)
- [x] Cleanup AWS account (S3 buckets, keys, networks, etc.) from left-overs of runs with outdated terraform playbooks
- [x] `setup.yaml` fails on first run on `filebeat`. When retried it passes... fix this.
`ERROR   instance/beat.go:906    Exiting: Error in initing input: Error creating input. No such input type exist: 'container'` - could be due to no containers being present on the machine when we try to run this...

NICE TO HAVE:
- [x] `swarm-init` and `swarm-join` roles are not idempotent. If something fails after them, we cannot re-run `setup.yaml` again without manually disabling them :/
---

One thing I found confusing is that currently `workers` and `redis` are separated in Terraform, although once the VMs are running, we actually add all EC2 instances to the Docker Swarm... so ultimately we end up with workers+1 nodes managed under Docker Swarm.


infra/ansible/inventories/group_vars/all.yaml
```diff
@@ -0,0 +1,10 @@
+ansible_user: ubuntu
+
+# geerlingguy.docker
+docker_users:
+  - ubuntu
+
+filebeat:
+  version: 6.x
+  cloud_id: "{{ lookup('onepassword', 'Filebeat configuration', field='filebeat_cloud_id', vault='Test Infra') }}"
```
olizilla: oh nice! I did not know you could `lookup('onepassword'`

nonsense: I'm not mad about it, since you have to install `op` and get a session token, but at least everyone on the team gets access to 1Password during onboarding, so at least no need to share additional tokens. Keybase encrypted file system is a nice solution as well, but I don't want to push another tool onto the team :)

olizilla: i hear you. those are both neat tricks.

infra/ansible/files/authorized_keys
```diff
@@ -0,0 +1,10 @@
+# Anton
+ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDckh8AO6+am3nFudUrRgmkqfXn2al6pSSAvnKIzHAnRrkhozXoO6yXpA+oei4wNglpIr0GO3FgZGkuddSZfxAcIaGdCCEj3LKXkIr7Oc+DjpFT/LjECMaeVncsK6c9DrDtE/Yjg76BfxkM6gP5VP773lrioFSnzytZPaJE/ZDc3wywh3K345D4LPufuP3b447Q/k51ZxsPHuOhh/201Gpp1ZjX6mk1gPDRBuX5cWyU6MBNt7M/r5/xfpM144BXz50ARVZGv8AGVAhubn2VYpMaq8vM4rsS8AACjJGQ/GlkZDuH5YFflsJ4D0UK1jbrvgt/16Az40GaY8gX2QiMFDb9
```
jimpick: One concern I have with baking in the ssh keys directly is that if a non-PL community member wants to set up testground on their own AWS account, they might not see this and might accidentally grant access.

infra/README.md
```diff
@@ -6,39 +6,23 @@ In this directory, you will find:
...
-- 2. Create a `~/.aws` folder to store your AWS credentials (Access Key ID and Secret Access Key). See how [here](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html)
-- 3. Register a SSH Key Pair at the EC2 dashboard in the AWS web console ([EC2 Dashboard, Network & Security](https://us-west-2.console.aws.amazon.com/ec2/home?region=us-west-2#KeyPairs:sort=keyName)).
-- 4. `cd terraform` and run `terraform init` to install all the deps necessary
-- 5. Create a file with the name `terraform.tfvars`. It should look like:
-
-```
-key_name = "<ssh key pair name registered in AWS>"
-tag      = "<name for your cluster, use only alphanumeric chars and underscores>"
-```
+- 1. Confiugre your AWS credentials
```
jimpick: s/Confiugre/Configure/

infra/ansible/files/authorized_keys
nonsense: I totally agree!

I think we should split the parts that are open source about our infrastructure, from the parts that are specific to PL. At the same time we want our infrastructure to be defined in code and under version control, so ultimately we need this data somewhere in a repo.

For now I consider `ipfs/testground` to be that repo (as visible in this PR), but maybe that's not right. I am open to suggestions on how and where to define the PL-specific parts.

In the past we've started with private repos, and then open sourced bits and pieces, but I have found this approach a bit lacking, because ultimately the private repos were always ahead of the public open-sourced versions.

I don't know what's the best way to do that.

nonsense: My concern with PL-specific infrastructure code is that it might not be private, but it will leak information about PL infrastructure, which could be a security concern. I think we spoke briefly recently about whether to introduce private repos for Testground infrastructure, and decided not to. I think it would be worth considering this here as well - while these SSH keys are not private, I don't see any benefit in them being in a public repo. There are other such cases - for example our DNS records.

infra/terraform/testground.tf
```diff
@@ -1,130 +1,13 @@
...
-  max_size            = "${var.workers}"
-  desired_capacity    = "${var.workers}"
-
-  tags_as_map = {
-    TG     = "${var.tag}"
-    Name   = "${var.tag}-worker"
-    TGRole = "worker"
-  }
+data "aws_ami" "base" {
+  most_recent = true
```
jimpick: I'm getting this error when I try to run:

```
Error: Error launching source instance: UnsupportedOperation: AMI 'ami-0f37d6c94b5686e2a' with an instance-store root device is not supported for the instance type 'c5.large'.
	status code: 400, request id: e405ffbf-9c25-433a-b1d1-b738358f3b68
```

I suspect the most recent AMI uploaded by Amazon supports the wrong type of root device. There's probably the ability to add an additional filter.

jimpick: I added this to get it to work:

```
filter {
    name   = "root-device-type"
    values = ["ebs"]
  }
```

infra/ansible/README.md
```diff
@@ -0,0 +1,31 @@
...
+## Login to onepassword
+
+Run 'eval $(op signin protocollabs)' to sign into your 1Password account, prior to executing playbooks that need access to secrets, such as `setup.yaml`.
+
+## Useful commands
+
+Perform a quick connectivity test
+
+```bash
+$ ansible all -m ping
```
jimpick: Also needs an `-i inventories/$CLUSTER_NAME` I think...

jimpick:
I got pretty far, but I can't seem to get past this step:

```
< TASK [swarm : Create TestGround overlay control network] >
 ---------------------------------------------------------- 
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||

fatal: [3.120.175.29]: FAILED! => {"changed": false, "msg": "Failed to import the required Python library (Docker SDK for Python: docker (Python >= 2.7) or docker-py (Python 2.6)) on 3.120.175.29's Python /usr/bin/python. Please read module documentation and install in the appropriate location. If the required library is installed, but Ansible is using the wrong Python interpreter, please consult the documentation on ansible_python_interpreter, for example via `pip install docker` or `pip install docker-py` (Python 2.6). The error was: No module named ssl_match_hostname"}
```

I'll investigate deeper if I have time.

jimpick:
I suspect that if we use the public IPs for the Docker Swarm instead of the private IPs, we'll end up having to pay for all the traffic between the machines...

```
$ ssh ubuntu@3.120.175.29 docker node ls
ID                            HOSTNAME            STATUS              AVAILABILITY        MANAGER STATUS      ENGINE VERSION
rtaj6jau6ynb0q0jii0rzykhd *   3.120.175.29        Ready               Active              Leader              19.03.3
jjqdmtn7n3z17qb9s8pv0tj27     3.122.56.232        Ready               Active                                  19.03.3
spsfgq0aw7qyu6ukg3ciakg63     18.185.105.174      Ready               Active                                  19.03.3
zd61bvpvlz1yk8reuye7vsfec     52.57.112.208       Ready               Active                                  19.03.3
```

infra/ansible/files/authorized_keys
lanzafame: IPFS-Cluster has this with its ansible repo, see https://github.com/hsanjuan/ansible-ipfs-cluster for the opensource stuff, and then there is the PL-specific implementation of such, which is private.

lanzafame: Another way to do this is to have an ansible role that slurps up github usernames and then uses the `github.com/<username>.keys` endpoint to gather the public keys. This way the users are just a variable in the host_vars, which makes it easy for other community members to swap them out.

infra/terraform/testground.tf
nonsense: I suspect this `data` call is not returning the correct AMI based on the input instance type. I haven't tested yet with a different instance type by the default - thanks for noticing this, I will make sure it works with ebs, instance store, and all other types.

infra/ansible/README.md
nonsense: I am still used to writing Ansible/Terraform playbooks for specific machines/clusters, where we set up defaults for inventories, but looks like here it won't be possible, since we'd be using these for multiple setups.

This makes me think that the `terraform backend` is also something that would be PL-specific, as we want a backend only for shared infrastructure, such as the production cluster.

nonsense:
> I got pretty far, but I can't seem to get past this step:
> 
> ```
> < TASK [swarm : Create TestGround overlay control network] >
>  ---------------------------------------------------------- 
>         \   ^__^
>          \  (oo)\_______
>             (__)\       )\/\
>                 ||----w |
>                 ||     ||
> 
> fatal: [3.120.175.29]: FAILED! => {"changed": false, "msg": "Failed to import the required Python library (Docker SDK for Python: docker (Python >= 2.7) or docker-py (Python 2.6)) on 3.120.175.29's Python /usr/bin/python. Please read module documentation and install in the appropriate location. If the required library is installed, but Ansible is using the wrong Python interpreter, please consult the documentation on ansible_python_interpreter, for example via `pip install docker` or `pip install docker-py` (Python 2.6). The error was: No module named ssl_match_hostname"}
> ```
> 
> I'll investigate deeper if I have time.

I have no idea what this is about - could be due to the difference instance type, but I doubt it. Let's try to reproduce, I haven't seen it so far.

nonsense:
> I suspect that if we use the public IPs for the Docker Swarm instead of the private IPs, we'll end up having to pay for all the traffic between the machines...
> 
> ```
> $ ssh ubuntu@3.120.175.29 docker node ls
> ID                            HOSTNAME            STATUS              AVAILABILITY        MANAGER STATUS      ENGINE VERSION
> rtaj6jau6ynb0q0jii0rzykhd *   3.120.175.29        Ready               Active              Leader              19.03.3
> jjqdmtn7n3z17qb9s8pv0tj27     3.122.56.232        Ready               Active                                  19.03.3
> spsfgq0aw7qyu6ukg3ciakg63     18.185.105.174      Ready               Active                                  19.03.3
> zd61bvpvlz1yk8reuye7vsfec     52.57.112.208       Ready               Active                                  19.03.3
> ```

I now realized that the hostnames on the machines are confusing. Actually the `docker manager` is advertising its private IP, and also the security groups are blocking all traffic that is not SSH and doesn't come from the private interfaces, so I think docker swarm is setup to be using the private network.

I will double check that, because you are right - if we use the public IPs we'd be paying for traffic I think.

Also hostnames of machines are confusing, so maybe I will set them to the private IPs.

infra/ansible/files/authorized_keys
nonsense: @jimpick @lanzafame I went back on the idea to have static SSH keys in this repo. I still think we should do it for the common infrastructure that the team maintains, but decided that this is not the right place for it.

I pretty much reverted the key configuration as @jimpick had done it in the initial PR.

nonsense:
> I got pretty far, but I can't seem to get past this step:
> 
> ```
> < TASK [swarm : Create TestGround overlay control network] >
>  ---------------------------------------------------------- 
>         \   ^__^
>          \  (oo)\_______
>             (__)\       )\/\
>                 ||----w |
>                 ||     ||
> 
> fatal: [3.120.175.29]: FAILED! => {"changed": false, "msg": "Failed to import the required Python library (Docker SDK for Python: docker (Python >= 2.7) or docker-py (Python 2.6)) on 3.120.175.29's Python /usr/bin/python. Please read module documentation and install in the appropriate location. If the required library is installed, but Ansible is using the wrong Python interpreter, please consult the documentation on ansible_python_interpreter, for example via `pip install docker` or `pip install docker-py` (Python 2.6). The error was: No module named ssl_match_hostname"}
> ```
> 
> I'll investigate deeper if I have time.

I got similar errors when running `ubuntu-disco`. I locked the version of Ubuntu to `ubuntu-bionic` (LTS supported until April 2023), and this is no longer happening.

For some reason on first run of `setup.yaml` I get `filebeat` error:
```
fatal: [3.124.209.15]: FAILED! => {"changed": false, "msg": "Unable to start service filebeat: Job for filebeat.service failed because the control process exited with error code.\nSee \"systemctl status filebeat.service\" and \"journalctl
 -xe\" for details.\n"}
```

When I retry, the playbook passes - I am not sure if this is timing/transient error or something serious, but if you encounter it, one workaround is to just run the playbook again.

infra/docker-swarm/terraform/iam.tf
```diff
@@ -0,0 +1,50 @@
...
+  name        = "${var.cluster}-ecr-policy"
+  description = "A policy giving access to full access to ECR"
+
+  policy = <<EOF
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Action": [
+        "ecr:*"
```
nonsense: I believe all instances need to have access to AWS ECR, so that they download the docker test plan image which would be run. Also `testground` which would probably run on one of these instances needs to have access to AWS ECR to push the built images. Related to: https://github.com/ipfs/testground/pull/213

nonsense:
I am marking this as ready for review, even though there are a few known TODOs that should be addressed soon, but maybe in follow up PRs, as this one is getting big.

I am at the stage of trying to run `testground daemon` and test plans on the machines setup by these scripts. This will involve adding more playbooks, such as deploying the `testground daemon` and configuring it with a local Go proxy and with the local Docker daemon, and AWS APIs.

nonsense:
@jimpick FYI (a few changes since our talk)
- filebeats config and installation fixed - the config needed amendments and i was installing the wrong version (6.x vs 7.x) - now logs are forwarded to our testing setup on elastic.co
- docker swarm init, and docker swarm join are idempotent - i got the config from your commit!

infra/docker-swarm/terraform/sg.tf
```diff
@@ -0,0 +1,29 @@
...
+  # outbound internet access
+  egress {
+    from_port = 0
+
+    to_port     = 0
+    protocol    = "-1"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+}
+
```
raulk: super nit: the trailing newline policy on these files is inconsistent. If TF/Ansible don't require a blank line, let's remove it from everywhere.

infra/docker-swarm/terraform/iam.tf
raulk: Correct. The workers need only "list" and "download" access, i.e. read-only. The daemon needs full access. Happy to keep it on "ecr:*" for now. We can tighten the security later, but please open an issue to remember.

infra/docker-swarm/ansible/roles/testground-services/tasks/main.yaml
```diff
@@ -0,0 +1,53 @@
...
+    force_source: yes
+    state: present
+
+- name: Start Redis service
+  docker_swarm_service:
+    name: testground-redis
+    image: "redis:latest"
+    command: redis-server
+    args:
+      - --notify-keyspace-events $szxK
```
raulk: No need to escape `$`? It's an unfortunate choice by Redis. It represents string commands: https://redis.io/topics/notifications#configuration

infra/docker-swarm/ansible/setup.yaml
```diff
@@ -0,0 +1,77 @@
...
+  # Filebeat
+  - role: filebeat
+    tags: [filebeat]
+  # Log rotation
+  - role: nickhammond.logrotate
+    logrotate_scripts:
+    - name: docker-containers
+      path: /var/lib/docker/containers/*/*.log
+      options:
+        - daily
```
raulk: The rotation may interfere with Filebeat. Have we investigated this?

infra/docker-swarm/README.md
```diff
@@ -0,0 +1,105 @@
...
+- 1. [Terraform](https://www.terraform.io/).
+- 2. [Ansible](https://www.ansible.com/).
+- 3. [op](https://support.1password.com/command-line-getting-started/).
+- 4. [terraform-inventory](https://github.com/adammck/terraform-inventory).
+
+## Set up infrastructure with Terraform
+
+The playbooks set up a "manager" machine, another machine to run Redis, and two worker machines for running tests on.
+
+1. Configure your AWS credentials
```
raulk: Using the AWS cli. Add a link to https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html

infra/docker-swarm/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+2. Pick a cluster name. Cluster names must be unique within the same AWS account.
+
+```
+export CLUSTER_NAME=demo
```
raulk: Is there a command to display all cluster names? If no, can we create it?

infra/docker-swarm/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+1. Run
+
+```
+terraform-inventory -inventory > ../ansible/inventories/$CLUSTER_NAME
+```
+
+## Configure your cluster with Ansible
+
+1. Sign into your 1Password account
```
raulk: This is PL-dependant. Everything else isn't. We need to extract this via some kind of "credentials provider" pluggable mechanism. I'm sure something like that exists in the Ansible universe?

infra/docker-swarm/ansible/setup.yaml
nonsense: This seems to be a non-issue for now. Let's deal with this problem, if it comes up.

infra/docker-swarm/README.md
nonsense: I suggest we do this in a follow up PR, this one is getting rather big already. There are many places where a collision can happen, but the chance of this happening is rather slim.

infra/docker-swarm/ansible/roles/testground-services/tasks/main.yaml
```diff
@@ -0,0 +1,53 @@
...
+      - subnet: 10.1.0.0/16
+
+- name: Pull ipfs/testground image
+  docker_image:
+    name: "ipfs/testground:latest"
+    source: pull
+    force_source: yes
+    state: present
+
+- name: Pull Redis image
```
raulk: I think this pull is redundant. This will pull the image in the master, not in the machine where Redis is scheduled.

infra/docker-swarm/ansible/roles/testground-services/tasks/main.yaml
```diff
@@ -0,0 +1,53 @@
+- name: Create Testground overlay control network
+  docker_network:
+    name: control
+    driver: overlay
+    attachable: yes
+    enable_ipv6: no
+    ipam_config:
+      - subnet: 10.1.0.0/16
+
+- name: Pull ipfs/testground image
```
raulk: What is this pull for?

infra/docker-swarm/README.md
nonsense: Any way we deal with `secrets` will be PL-specific. There are a few options in the PR description, secrets management is not trivial and I don't particularly like any of the listed solutions, so I picked 1Password, because we already use it.

infra/docker-swarm/ansible/setup.yaml
raulk: Actually @nonsense, I disagree with deferring. If this shows up, it'll show up as "logs missing". And it'll send us down a painful debugging trip which can last for hours or days. 10 minutes well spent now investigating this can save us from a lot of pain.

infra/docker-swarm/README.md
raulk: Let's open an issue for it -- or a compilation issue with all the pending TODOs ;-)

infra/docker-swarm/ansible/roles/testground-services/tasks/main.yaml
nonsense: You're right - I was confused because this is generally needed when you do `docker_container`, but here we actually schedule a `docker_swarm_service`, so it should do the pull.

infra/docker-swarm/ansible/roles/testground-services/tasks/main.yaml
nonsense: Same as comment above - confused, because I thought we need it for `sidecar`. Removed.

infra/docker-swarm/README.md
nonsense: Added link.

raulk:
> Why is Redis virtual machine part of the Docker Swarm when we don't want to schedule test plans on it? AFAICT Redis is just a dependency that we can run anywhere we see fit. Is there a requirement for Redis to be part of the swarm? I guess we are just piggy-backing on the docker service discovery, or?

We schedule Redis via Docker Swarm, and we access it via the internal overlay DNS resolver, i.e. we use the hostname `redis`. The alternative is to run it bare, outside a container, and make sure we can still access it with the system DNS resolver. We'd need to set up a systemd service to restart upon failure. Too much pain for no gain.

infra/docker-swarm/terraform/sg.tf
nonsense: Removed.

infra/docker-swarm/ansible/roles/testground-services/tasks/main.yaml
nonsense: This is something I will be reviewing next. Redis config doesn't seem correct right now, but this also depends on the service discovery we have coded in Testground.

nonsense:
@raulk this PR is already rather large, so I merged, and we can address open questions in follow up PRs.

> We schedule Redis via Docker Swarm, and we access it via the internal overlay DNS resolver, i.e. we use the hostname `redis`. The alternative is to run it bare, outside a container, and make sure we can still access it with the system DNS resolver. We'd need to set up a systemd service to restart upon failure. Too much pain for no gain.

I am not convinced this is the best way to do this. Seems like the only benefit is getting the service discovery `for free`. Redis is a critical part of `Testground` (and currently not setup correctly in this PR), so we probably want a bit more hardened setup for it. `backup` and `upgrade` strategy are not clear right now.

raulk:
> @raulk this PR is already rather large, so I merged, and we can address open questions in follow up PRs.

Mind opening a follow-up issue with a TODO list of remnants from here? Thanks!

nonsense:
Follow up PR: https://github.com/ipfs/testground/pull/217

raulk:
> I am not convinced this is the best way to do this. Seems like the only benefit is getting the service discovery for free. Redis is a critical part of Testground (and currently not setup correctly in this PR), so we probably want a bit more hardened setup for it. backup and upgrade strategy are not clear right now.

@jimpick and I discussed this at the outset and settled for running inside Docker Swarm because, besides performance, we see no gain in running it outside at this moment, and it's the simpler approach. Let's reevaluate the decision once we're running production workloads and find Redis to be the contention/unstable piece of the puzzle (if we do), and we determine that running bare metal would fix the issue. Right now we'd be over-architecting.

BTW - we don't need backup or upgrade right now. We've in fact turned off persistence altogether. Upgrading to a new Redis version would be as simple as doing: `docker service update redis`.

PR #211, opened 2019-12-05
Update HOW_WE_WORK.md

daviddias:
So that we don't forget https://github.com/ipfs/testground/pull/203#issuecomment-561343014 :)

### DIFF
--- a/docs/HOW_WE_WORK.md
+++ b/docs/HOW_WE_WORK.md
@@ -1,8 +1,16 @@
 # How we work together
 
+## Weekly Sync
+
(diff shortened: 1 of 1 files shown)


docs/HOW_WE_WORK.md
```diff
@@ -25,3 +33,7 @@ We track our work Kanban style in a [Zenhub board](https://app.zenhub.com/worksp
 - Responsibilities:
   - Project Maintainer and/or Project Architect - Review issues on Inbox, break them down if necessary, move them into Ready when it is the right time. Also, label issues with priority and difficulty.
   - Contributors move issues between the Ready, In Progress and Review/QA Colums. Use help wanted and blocked labels in case they want to flag that work.
+  
+## Contributing Guildeines
```
nonsense: This is already part of the main entry README.md - not sure we should copy-paste a link to it in more places.

daviddias: It makes sense as the README is designed for both users and contributors and this doc is designed for contributors (it reminds contributors that the guildelines are part of how we work).

PR #213, opened 2019-12-06
use aws credentials only if set

nonsense:
We don't have to set AWS credentials when there are other ways to authenticate and get authorisation for a given AWS API.

When `testground` runs as a `daemon` in AWS, the instance itself would have a role, which allows it to query AWS ECR, so we wouldn't have to generate credentials and configure them in the `.env.toml`.

This PR is basically changing the way we create an AWS client, by only using the static credentials if they have been set.

### DIFF
--- a/pkg/aws/ecr.go
+++ b/pkg/aws/ecr.go
@@ -24,8 +24,15 @@ type ecrsvc struct{}
 
 // newService creates a new ECR backend service stub.
 func (*ecrsvc) newService(cfg config.AWSConfig) (*ecr.ECR, error) {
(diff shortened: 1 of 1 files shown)


nonsense:
@jimpick FYI in case you try to run the `daemon` on the cluster we created.

nonsense:
I also added an S3 bucket and redirect to github.com/ipfs/testground as per https://github.com/ipfs/testground/issues/38

This works only for HTTP.

Bonus points if someone wants to modify this for HTTPS (AFAIK we have to go through CloudFront and also get a certificate for testground.ipfs.team , I don't think it is possible only via AWS S3)

PR #214, opened 2019-12-07
rename ready to backlog

daviddias:
Apparently, `ready` often confuses a ton of people as in "PR is ready". Given that `ready` has been used as a synonym for `backlog`, I'm renaming it to the widely accepted term.


daviddias:
Updated Zenhub as well

PR #230, opened 2019-12-10
fix: flush local:exec output

hacdias:
Flushes local:exec output.

License: MIT


nonsense:
Just a sidenote - this is adding `flush`ing all runners, not just the local:exec. After we merged the daemon, every command goes to a testground client, and is handled by the testground daemon/server. The flushing here is needed so that as soon as something is processed on the testground daemon/server side, you send an HTTP chunked content to the testground client.

PR #231, opened 2019-12-10
WIP deployments design blueprint.

raulk:


docs/DEPLOYMENT.md
```diff
@@ -0,0 +1,195 @@
...
+* *Ansible playbooks* to manage the logical infrastructure, such as Docker
+  services, system services, overlay networks, etc. The Ansible scripts are
+  _cloud-leaky_. In theory they could be cloud-agnostic, but in practice various
+  items such as S3 storage, AWS ECR authentication, etc. are cloud-specific.
+    * Users wishing to deploy testground on an alternate provider, will need to
+      patch those elements.
+* _[terraform-inventory](https://github.com/adammck/terraform-inventory)_ for
+  the hand-off between the two.
+
+These assets handle the intantiation of various node types of nodes, and the
```
jimpick: s/intantiation/instantiation/

docs/DEPLOYMENT.md
```diff
@@ -0,0 +1,195 @@
...
+
+     * URL: `dev-<name>.testground.ipfs.team`
+
+     * Availability: Developers don't work 24/7, and neither should their
+       playground be powered on 24/7. Doing so would be wasteful. At the same
+       time, we don't want to destroy and recreate the environment frequently;
+       that takes time, and there's nothing worse than sitting with your arms
+       crossed / context switching for 10 minutes before you can start working.
+
+     * Scaling: the scale is determined manually. The developer uses
```
nonsense: Who would be the user of this? Let's survey developers if this is a workflow that they feel comfortable with, because I am not convinced everyone is going to be happy setting up infrastructure by themselves. This is rather time consuming and takes focus always from application development.

docs/DEPLOYMENT.md
```diff
@@ -0,0 +1,195 @@
...
+  items such as S3 storage, AWS ECR authentication, etc. are cloud-specific.
+    * Users wishing to deploy testground on an alternate provider, will need to
+      patch those elements.
+* _[terraform-inventory](https://github.com/adammck/terraform-inventory)_ for
+  the hand-off between the two.
+
+These assets handle the intantiation of various node types of nodes, and the
+backing infrastructural elements (e.g. machines, VPCs, security groups, etc.):
+
+1. **Testground master**: runs the daemon, watches GitHub, exposes an
```
jimpick: After the big brouhaha in the redis community about the master/slave naming, I think it's best to steer clear of "master" ... I was using "manager" (what Docker Swarm uses) but there are other words.

jimpick:
It's great to see this written down!

nonsense:
@raulk I like this document, it is a good description of many of the moving parts of the `testground project` that we are thinking about. It also features some of the design decisions we are making, that we should test out with developers in PL and iterate on.

Here are my thoughts on a few things that sound good in theory, but I think will be difficult in practice
1. Developer experience - `The developer uses Terraform/Ansible to input the initial scale upon instantiation, and to scale up and down from there onwards.` - this is very time consuming, and developers don't like to deal with infrastructure. I don't see this getting adopted easily, so we should survey developers and gather feedback sooner rather than later.
2. Environments - prod, stg, exp + `n` dev environments seems a bit too many to me even when thinking in the mid-/long- term. I think `dev` environments could be multiplexed on one physical `dev` env, which would reduce complexity, even if it doesn't have as good of isolation as `prod`.

docs/DEPLOYMENT.md
raulk: 👍

raulk:
Merging the current state. Something is better than nothing!

PR #236, opened 2019-12-11
feat: add support for configuration with IPTB & directory sharding test with ext. daemon

hacdias:
This adds support for custom daemon configuration on IPTB (daemon) tests through a function that modifies the current configuration. I decided to use a function to modify the configuration instead of the configuration itself for some reasons:

1. It's what `CreateIpfsInstance` already does.
2. It easily allows to update the configuration of the IPTB-generated daemon with the command `ipfs config replace`.

Also, this enabled me to add one more test: dir sharding with the daemon! And the CIDs match!

### DIFF
--- a/plans/chew-large-datasets/main.go
+++ b/plans/chew-large-datasets/main.go
@@ -59,37 +59,36 @@ func main() {
 		}
 	}
 
+	addRepoOptions := tc.AddRepoOptions()
+
 	if testCoreAPI {
-		apiOpts := tc.InstanceOptions()
-
-		if apiOpts == nil {
-			fmt.Println("Test not implemented against CoreAPI yet")
-		} else {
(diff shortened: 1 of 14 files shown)


plans/chew-large-datasets/main.go
```diff
@@ -59,37 +59,36 @@ func main() {
...
+			AddRepoOptions: addRepoOptions,
+		}, "node")
+
+		ensemble := iptb.NewTestEnsemble(ctx, spec)
+		ensemble.Initialize()
+		defer ensemble.Destroy()
+
+		// In this test suite we agree that the node is tagged as 'node'.
+		node := ensemble.GetNode("node")
+		client := node.Client()
```
daviddias: Can you add a func iptb.spawnDaemon that all the parameters required and returns the node and the client? It would simplify this a lot

hacdias: I can!

PR #239, opened 2019-12-11
feat: check for runner/builder compatibility

### LINKED ISSUE #191: Some runners are incompatible with some builders and fail silently
Trying to run TestGround with:

```
testground -vv run dht/find-peers --builder=exec:go --runner=local:docker --test-param timeout_secs=15
```

results in silent failure, probably because we cannot run a test plan in docker, while asking to run a build with `exec:go`.

hacdias:
personally, I don't like the strategy of just checking for the compatibility on the running phase. However, it was the simplest change I could do to make it work.

Closes #191.

License: MIT

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -112,10 +112,13 @@ func runCommand(c *cli.Context) error {
 			return fmt.Errorf("fatal error from daemon: %s", err)
 		}
 
-		artifactPath, err = client.ProcessBuildResponse(resp)
+		out, err := client.ProcessBuildResponse(resp)
 		if err != nil {
 			return err
 		}
+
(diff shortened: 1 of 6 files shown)


nonsense:
I also think that this should probably be checked in the first possible step, and not when we reach the `runner` phase. As soon as a command is proceed by the `testground daemon` it should have an idea if it is a `valid` command and whether it can run it.

Then again for now, I think this is an improvement, so I am happy to merge it.

hacdias:
@daviddias can I make the tests on a separate PR? i think it's better

daviddias:
> @daviddias can I make the tests on a separate PR? i think it's better

I said yes verbally :)

PR #240, opened 2019-12-11
feat: add (in)compatibility runner/builder test

hacdias:
This tests the compatibility and incompatibility between builder and runner through a simple E2E test where we spin a server and then make requests to it.

For this, I created a placebo test plan which does nothing and it might be useful for future testing.

License: MIT


plans/do-nothing/README.md
```diff
@@ -0,0 +1,3 @@
+# Do Nothing
+
+This is a placebo test for testing purposes.
```
daviddias: oh, call the test placebo then :D

hacdias: Done!

pkg/server/server_test.go
```diff
@@ -0,0 +1,123 @@
...
+		t.Fail()
+	}
+}
+
+func TestCompatibleBuilder(t *testing.T) {
+	api := client.New(getServer())
+
+	req := &client.BuildRequest{
+		Plan:    "placebo",
+		Builder: "exec:go",
```
daviddias: No runner?

hacdias: This is the build request so yes, no runner.

daviddias: So you are building the placebo to then use it with the simple-add runner?

Why call this test `TestCompatibleBuilder` then? Shouldn't it be TestINCompatibleBuilder?

pkg/server/server_test.go
```diff
@@ -0,0 +1,123 @@
...
+	"bytes"
+	"context"
+	"io"
+	"net/http"
+	"testing"
+
+	"github.com/ipfs/testground/pkg/daemon/client"
+)
+
+var srv *Server
```
nonsense: I think it'd be better to not have global vars in tests. We want to be able to run parallel tests. Also the port of the server should be given by the kernel, so just us ":0", and not a hardcoded one.

pkg/server/server_test.go
hacdias: This test tests builder `exec:go` (line 87, 108) with runner `local:exec` so everything is correct.

pkg/server/server_test.go
hacdias: Doooone!

pkg/server/server_test.go
daviddias: image

PR #246, opened 2019-12-11
Delete checked in binary, fix test renaming inside go.mod and *.go files

jimpick:
I'm assuming the binary wasn't supposed to be there.


jimpick:
I'm not sure how to run the test, so I didn't try that.

PR #249, opened 2019-12-12
feat: return non-0 code on failure and streamline server response format

### LINKED ISSUE #241: Meaningful HTTP errors
While working on #240, I noticed that `api.Build` and `api.Run` do not return errors that are not actual network errors.

On the server side, when there is an error, we don't send any response to the client. We should send an adequate HTTP error (500 for internal errors, 40x for client configuration errors) and a message telling what happened.

hacdias:
Streamlines all server responses to one format and correctly handles the errors on the client side. This fixes #241.


pkg/daemon/client/client.go
```diff
@@ -88,72 +88,90 @@ func (c *Client) Run(ctx context.Context, r *RunRequest) (io.ReadCloser, error)
...
+			return mapstructure.Decode(result, resp)
+		},
+	)
+
+	return resp, err
+}
 
-		case "result":
+// ProcessDescribeResponse parses a response from a `describe` call
+func ProcessDescribeResponse(r io.ReadCloser) error {
```
daviddias: so call it `parse` instead of `process`? (This might be an artefact from a direct Portuguese -> English translation)

hacdias: Doone!

PR #262, opened 2019-12-13
Improve Test Plan Specs

daviddias:
I've been wanting to scratch a itch for a while... 

Currently, the plan specs are detached from the manifests, one has to manually update both. This is cumbersome and it basically throws us in the traditional problem of any spec document, that the implementation starts to differ from the spec.

At the same time, our manifest didn't offer an uniform way to describe the different properties and parameters for Builders, Runners and Tests. 

This PR is a proposal to:
- Remove the description of the parameters in the plan specs and instead use the manifest file for that
- Create a uniform way to describe the different Builders, Runners and Tests
- (extra) remove those `seq` numbers for each test case. We really don't need them :) 

Given that this will require a bunch of refactor, I wanted to pass it through you before creating the complete PR. What do you think?

### DIFF
--- a/manifests/chew-datasets.toml
+++ b/manifests/chew-datasets.toml
@@ -1,37 +1,52 @@
 name = "chew-datasets"
 
-# hashicorp/go-getter URLs, so in the future we can support fetching test plans from GitHub.
+# hashicorp/go-getter URLs support.
 source_path = "file:${TESTGROUND_SRCDIR}/plans/chew-datasets"
 
 [defaults]
 builder = "docker:go"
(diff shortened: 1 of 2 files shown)


daviddias:
https://github.com/ipfs/testground/pull/265 reminds me that the number of instances is meaningless when set on the test, this should be specified as a runner parameter instead, with safe defaults for each runner.

raulk:
The chew-datasets plan has been erased, as it was starting an IPFS binary which the docker:go builder no longer supports.

PR #263, opened 2019-12-13
fix: redis race condition when getting bootstrappers

### LINKED ISSUE #252: Redis write/subscribe race
#### Description:

The DHT testplan currently fails randomly, waiting for bootstrappers to appear (in `getBootstrappers`). However, the bootstrappers are clearly registering themselves.

I've noticed a key difference between how we write bootstrapper records versus peer records:

* For peer records, we always _write_ a peer record before we subscribe.
* For bootstrap records, non-bootstrappers will subscribe without writing.

That means, for bootstrap records, we may subscribe to a key that hasn't been written yet. I'm guessing we have some kind of race in our subscription logic for keys that are being written for the first time while another node is concurrently subscribing to the record.

hacdias:
This should fix #252. It also includes the improvements made on `go-libp2p-kad-dht`. We now wait for every peer to be subscribed before starting to write to everyone will receive all bootstrap peers.

Please let me know what you think about the code. I don't think it looks as pretty as it could and I've never used channels before :O

License: MIT


hacdias:
@raulk could you give me some help with this one? I've been looking `sync/subscription.go` for the past ~2 hours and I couldn't manage to fix this. I've noticed you've already fixed some race conditions on that package in the past so you might be better aware of what might be happening or not :o

hacdias:
@raulk @Stebalien I just pushed a proof of concept using streams (which make this work!). however, I haven't changed everything because I notice we're relying on trees of changes and changing to streams would require refactoring the whole sync thingy.

I would like your feedback here!

raulk:
@hacdias that's a great experiment!

I evaluated Redis Streams in https://github.com/ipfs/testground/issues/27 when I moved from Consul to Redis as a backend for the sync service.

The drawback is that you can't set TTL for entries in a stream. My design of the sync service revolved around ZooKeeper's concept of ephemeral nodes: nodes that are alive as long as the writer process/session is alive and pinging. If a node dies, its key will disappear automatically, and the system guarantees that.

You can't achieve that with streams, it's an append-only log, and that's why I had originally discarded the idea.

But armed with a lot more information now, I'm happy to reconsider that decision because the worst thing that could happen is that we'll have stale data in the stream, i.e. addresses, CIDs, multihashes, peer IDs, etc. and when using them, if the process has died the test will functionally error out.

We will want to set a TTL on the entire stream though, and we'll need a "keepalive" supervising procedure (e.g. every 5th node that joins the test case sets a timer to bump up the TTL every 10 seconds). Otherwise we'll collect a bunch of garbage over time, and that's not good.

hacdias:
@raulk @Stebalien I updated this to use streams instead of pubsub. Now the DHT test does not have problems getting bootstrappers. Note: to test this, you must rebuild the sidecar image first because it also uses the `sync` package!

The API stays basically the same, except for `watcher.Close` that does not return any errors.

hacdias:
ping @raulk @Stebalien @nonsense @daviddias

raulk:
Superseded by https://github.com/ipfs/testground/pull/310.

PR #268, opened 2019-12-13
set IPAM for docker swarm service in swarm runner

nonsense:
By default calling a network create, creates a network with 24 bit mask, so we can't schedule more than ~256 containers.

This PR is changing the mask to 16 bits.

We also do that at https://github.com/ipfs/testground/blob/master/infra/docker-swarm/ansible/roles/testground-services/tasks/main.yaml#L7

### DIFF
--- a/pkg/runner/cluster_swarm.go
+++ b/pkg/runner/cluster_swarm.go
@@ -8,6 +8,7 @@ import (
 	"reflect"
 	"time"
 
+	"github.com/docker/docker/api/types/network"
 	"github.com/ipfs/testground/pkg/api"
 	"github.com/ipfs/testground/pkg/aws"
 	"github.com/ipfs/testground/pkg/logging"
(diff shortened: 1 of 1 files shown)


nonsense:
This should address and hopefully close https://github.com/ipfs/testground/issues/266

pkg/runner/cluster_swarm.go
```diff
@@ -142,6 +143,13 @@ func (*ClusterSwarmRunner) Run(input *api.RunInput, ow io.Writer) (*api.RunOutpu
 		Internal:   true,
 		Attachable: true,
 		Scope:      "swarm",
+		IPAM: &network.IPAM{
+			Driver: "default",
+			Config: []network.IPAMConfig{{
+				Subnet:  "192.168.0.0/16",
+				Gateway: "192.168.0.1",
```
Stebalien: FYI, I'm setting permanent network ranges with public and private IP addresses in https://github.com/ipfs/testground/pull/271.

(doesn't block this PR, I'll rebase that one).

PR #270, opened 2019-12-13
sdk/sync: improve test coverage.

raulk:
Relates to #252. This PR introduces two tests to cover:

* subscription before first write (i.e. while the key still doesn't exist): ensures that the subscriber "latches onto" the subtree and that all writes are consequently received.
* write before subscribe: writes all keys, then subscribes. Checks that the sync service correctly replays the state to the subscriber.

These tests are writing 1000 keys, and they're passing iteration after iteration, so something else is at play here, and I think I have an idea -- could be a very specific race between the moment between we subscribe and we acquire the last seq. But that would only make sense if we miss very few items.

### DIFF
--- a/sdk/sync/redis_test.go
+++ b/sdk/sync/redis_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"math/rand"
 	"os/exec"
+	"reflect"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


Stebalien:
> But that would only make sense if we miss very few items.

We're missing exactly one of one.

PR #271, opened 2019-12-13
Sidecar: add method to set IP addresses

Stebalien:
* Make it possible to set IP addresses with the sidecar (docker only, may not work in swarm?).
* Document the sidecar.
* Use a consistent subnet for the control network.
* Test the sidecar.

TODO:

* [x] Testing:
  * [x] Local docker - tested locally but I need to add an automated test.
  * [ ] Docker swarm - I'll need help with this one.

### DIFF
--- a/docs/NETWORKING.md
+++ b/docs/NETWORKING.md
@@ -7,7 +7,7 @@ control and a data network.
 * Test instances communicate with the sync service, and _only_ the sync service,
   over the control network.
 
-The local runner will use your machines local network interfaces.
+The local runner will use your machine's local network interfaces. For now, this runner doesn't support traffic shaping.
 
 ## Control Network
 
--- a/docs/NETWORKING.md
+++ b/docs/NETWORKING.md
@@ -1,0 +1,33 @@
+# Networking
+
+All testground runners _except_ for the local runner have two networks: a
+control and a data network.
+
+* Test instances communicate with each other over the data network.
+* Test instances communicate with the sync service, and _only_ the sync service,
+  over the control network.
+
+The local runner will use your machines local network interfaces.
+
+## Control Network
+
+The "control network" runs on 192.168.0.1/16 and should only be used to
+communicate with the sync service.
+
+After the sidecar is finished [initializing the
+network](https://github.com/ipfs/testground/blob/master/docs/SIDECAR.md#initialization),
+it should be impossible to use this network to communicate with other nodes.
+However, a good test plan should avoid listening on and/or announcing this
+network _anyways_ to ensure that it doesn't interfere with the test.
+
+## Data Network
+
+The "data network", used for all inter-instance communication, will be assigned
+a B block in the IP range 16.0.0.0 - 32.0.0.0. Given the B block X.Y.0.0,
+X.Y.0.1 is always the gateway and shouldn't be used by the test.
+
+The subnet used will be passed to the test instance via the runtime environment
+(as `TestSubnet`).
+
+You can change your IP address (within this range) at any time [using the
+sidecar](https://github.com/ipfs/testground/blob/master/docs/SIDECAR.md#ip-addresses).
--- a/docs/SIDECAR.md
+++ b/docs/SIDECAR.md
@@ -66,7 +66,7 @@ if err != nil {
 
 ### Configure: Create
 
-Once the network is ready, you'll need ot actually _configure_ your network.
+Once the network is ready, you'll need to actually _configure_ your network.
 
 ```go
 config := sync.NetworkConfig{
--- a/docs/SIDECAR.md
+++ b/docs/SIDECAR.md
@@ -1,0 +1,180 @@
+# Sidecar
+
+Before you read this, read the
+[networking](https://github.com/ipfs/testground/blob/master/docs/NETWORKING.md)
+documentation.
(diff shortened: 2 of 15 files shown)


jimpick:
Thank you for the docs!

pkg/runner/common.go
```diff
@@ -0,0 +1,11 @@
...
+
+// Use consistent IP address ranges for both the data and the control subnet.
+// _which_
+var (
+	controlSubnet  = "192.168.0.0/16"
+	controlGateway = "192.168.0.1"
+	// this gives us two public A blocks and a private A block: 8., 9., & 10.
+	dataSubnet  = "8.0.0.0/5"
+	dataGateway = "8.0.0.1"
+)
```
Stebalien: I'm hard-coding these IP ranges. Thoughts? This makes it easier to write tests with consistent IP addresses.

docs/NETWORKING.md
```diff
@@ -0,0 +1,36 @@
...
+
+* Test instances communicate with each other over the data network.
+* Test instances communicate with the sync service, and _only_ the sync service,
+  over the control network.
+  
+The local runner will use your machines local network interfaces.
+
+## Control Network
+
+The "control network" runs on 192.168.0.1/16 and should only be used to
```
nonsense: @Stebalien could you also update `infra/docker-swarm/ansible/roles/testground-services/tasks/main.yaml` with the IPAM, when you are about to merge this?

pkg/sidecar/link.go
```diff
@@ -183,8 +183,32 @@ func (l *NetlinkLink) Shape(shape sync.LinkShape) error {
 
 // Set the links address.
 // NOTE: This won't work in docker; use docker connect/disconnect.
-func (l *NetlinkLink) SetAddr(ip *net.IPNet) error {
-	return l.handle.AddrReplace(l.Link, &netlink.Addr{IPNet: ip})
+func (l *NetlinkLink) AddrAdd(ip *net.IPNet) error {
```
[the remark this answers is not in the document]
nonsense: Could you add this as a comment to all methods to which this applies? This is not obvious to me :) I only see `This won't work in docker` comment, but `None of this code is currently used. It's here for a future runner that doesn't use docker.` will be more helpful.

Stebalien:
This still needs tests but it's ready for review.

@jimpick could you:

* Read the docs and, based on that, augment the test (or create a new test) that assigns IP addresses? (sanity check the docs)
* Test this out on docker swarm to make sure this actually works?

jimpick:
@Stebalien I was already working on a TCP-only (no libp2p) test which transfers data from a client to a server, so I added sidecar support to it to test:

https://github.com/jimpick/testground/blob/jim/shaping-experiment/plans/smlbench2-tcp/main.go

The test without traffic shaping passes, but when I turn on the traffic shaping, the listening process doesn't seem to complete receiving the upload.

I did see the test pass a few times with traffic shaping with a smaller data transfer - I'll have to investigate further tomorrow ... it could just be my go isn't very good. :-)

I only got all together a few hours ago - it should be a good place to start for testing it with IP address assignments or with Docker Swarm.

The docs were really helpful.

Stebalien:
Turns out the IP address assignment won't work because we can't share IP address ranges between distinct subsets. Unfortunately, we _need_ to be able to assign custom IP blocks to test more than 255 nodes.

So, I'm going to start using redis to coordinate IP address assignment.

docs/NETWORKING.md
```diff
@@ -0,0 +1,33 @@
+# Networking
+
+All testground runners _except_ for the local runner have two networks: a
+control and a data network.
+
+* Test instances communicate with each other over the data network.
+* Test instances communicate with the sync service, and _only_ the sync service,
+  over the control network.
+
+The local runner will use your machines local network interfaces.
```
raulk:
```suggestion
The local runner will use your machine's local network interfaces. For now, this runner doesn't support traffic shaping.
```

docs/SIDECAR.md
```diff
@@ -0,0 +1,180 @@
...
+hostname, err := os.Hostname()
+if err != nil {
+    runenv.Abort(err)
+    return
+}
+```
+
+### Configure: Create
+
+Once the network is ready, you'll need ot actually _configure_ your network.
```
raulk:
```suggestion
Once the network is ready, you'll need to actually _configure_ your network.
```

pkg/sidecar/link.go
```diff
@@ -181,10 +181,43 @@ func (l *NetlinkLink) Shape(shape sync.LinkShape) error {
...
+	return l.handle.AddrAdd(l.Link, &netlink.Addr{IPNet: ip})
+}
+
+// AddrDel removes an address from the link.
+//
 // NOTE: This won't work in docker; use docker connect/disconnect.
-func (l *NetlinkLink) SetAddr(ip *net.IPNet) error {
-	return l.handle.AddrReplace(l.Link, &netlink.Addr{IPNet: ip})
+func (l *NetlinkLink) AddrDel(ip *net.IPNet) error {
+	return l.handle.AddrAdd(l.Link, &netlink.Addr{IPNet: ip})
```
raulk: `AddrAdd` here too?

PR #273, opened 2019-12-14
adding dockerhub registry as an option

nonsense:
This PR is adding DockerHub as an option for Docker Registry.

Ideally we shouldn't use this as we will incur data transfer costs, but this is handy for development, until we set on a specific cloud provider (such as AWS, DO, GCP, etc.)

Ideally we should use an internal registry to the provider we pick - DO or GCP or other.

### DIFF
--- a/env-example.toml
+++ b/env-example.toml
@@ -9,6 +9,11 @@ access_key_id = "<aws access key id>"
 secret_access_key = "<aws secret access key>"
 region = "<aws region>"
 
+["dockerhub"]
+repo = "repo to be used for testground"
+username = "username"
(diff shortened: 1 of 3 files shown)


nonsense:
FYI this works also with private registries, so we could use it if we add a docker registry as part of the `testground dependencies` on k8s.

nonsense:
This is a candidate for refactor once we get Testground to the MVP - looks like we don't need multiple functions to call `docker push` in various ways, the two things that differ are:
- URI which is part of the docker image name/tag , by which docker infers if it should push to Docker Hub or to a custom registry
- Authentication which is passed with the options.

PR #274, opened 2019-12-14
Add a few more small tests  and ignores plans/ on Codecov

hacdias:
- Adds a few more small tests.
- Ignores `plans/` on Codecov since they are actual... tests. This will make a bump in the coverage amount because it was counting with those files (I guess... waiting for CI).

**Update**: since go modules, `go test ./...` only tests the current module and not **everything in the directory** so we've not been running all of our tests... https://github.com/golang/go/wiki/Modules#should-i-have-multiple-modules-in-a-single-repository


raulk:
We have disabled Codeconv -- will enable when we do a big testing push in v0.6.

PR #275, opened 2019-12-14
initial k8s runner with a JobSpec

nonsense:
This PR is adding a k8s runner, which creates a `Job` that creates `replica` `Pods` with `restartPolicy` set to `Never`.

There are no guarantees given by k8s that a given `Job` will run exactly `replica` number of `Pods` (especially if some panic very quickly on startup, then the controller might start more than `replica`), so we might need to address this in the future.

This has been tested with docker images hosted on DockerHub registry - there might be permissions problems is we use another registry.

---

TODO:

- [x] Fix compilation error on Windows


pkg/runner/cluster_k8s.go
```diff
@@ -0,0 +1,226 @@
...
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	_ api.Runner = &ClusterK8sRunner{}
+)
+
+func homeDir() string {
+	if home := os.Getenv("HOME"); home != "" {
```
Stebalien: https://golang.org/pkg/os/#UserHomeDir?

pkg/runner/cluster_k8s.go
```diff
@@ -0,0 +1,220 @@
...
+			log.Warnw("transient job error", "job", jobName, "err", err)
+			time.Sleep(300 * time.Millisecond)
+			continue
+		}
+
+		log.Debugw("job status", "job", jobName, "active", job.Status.Active, "succeeded", job.Status.Succeeded, "failed", job.Status.Failed)
+		if job.Status.Active == 0 && (job.Status.Succeeded > 0 || job.Status.Failed > 0) {
+			break
+		}
+		if time.Since(start) > 5*time.Minute {
```
nonsense: This will need changing and is worth having a discussion on it - we have to be careful to not `timeout` any plans, but the `timeout` parameter is dependent on scale - 100 pods can be scheduled in a few seconds, 10k pods will be scheduled in a few minutes.

We don't want to be in a situation where test plans start to error out before the last instances have had a chance to start.

nonsense:
This is far from ready, but since it is already useful from @jimpick I am also up for merging it and iterating on it, as the k8s runner is not on the critical path right now.

PR #276, opened 2019-12-14
remove README chunk on USAGE of Testground on Cloud

nonsense:
I think this part is now outdated, and I'd rather we don't have docs, than have docs that are outdated.

We have multiple ways of running Testground right now, and I am not sure it makes sense documenting them extensively, until we as a team internally agree on an overall direction, for example:

0. Testground client and daemon on local machine + local Docker Swarm instance
1. Testground client and daemon on local machine + local k8s instance
2. Testground client on local machine + Testground daemon on AWS + Docker Swarm on AWS
3. Testground client and daemon on local machine + Docker Swarm on AWS
4. Testground client and daemon on local machine + k8s on AWS EKS
5. Testground client and daemon on local machine + k8s on Digital Ocean
6. Testground client and daemon on local machine + k8s on GKE

~Right now it looks like we will settle for 5 or 6 for the MVP (none of the others will be able to run 1k-10k containers in the near future)~, but each of these setups have their quirks and won't be automated in one command in the near future, so I suggest we document various options separately, in their own Markdown, under `infra`.


nonsense:
@daviddias this goes together with the plan to have @hacdias run the setup based only on docs. The plan is to have a single Markdown file under `infra` which details option 4 or 5.

Even though we have the initial Testground daemon implemented and merged, there is no good reason to run it remotely today, so this step is simplified on purpose.

nonsense:
@daviddias we have documentation under `infra/docker-swarm` currently only for `AWS` as we don't run Docker Swarm anywhere else. Furthermore the docs here are outdated - I think they applied to the playbooks previously, but not anymore - we don't tunnel to the `docker.sock` anymore.

Basically `infra/k8s` should contain docs for running Testground with K8s on a given cloud, and `infra/docker-swarm` should contain docs for running Testground with Docker Swarm.

Usage docs like `Testground automatically loads an `.env.toml` file at root of your source directory.` do not belong on the Infrastructure part I think - `.env.toml` is to be configured for any type of runner.

Ultimately I don't think we miss any of these sentences, so I removed them.

nonsense:
@daviddias after my comment from 6 days ago, actually it turned out that currently devs (@raulk , @jimpick , and future users) are using Docker Swarm over the Docker socket, directly connecting to Docker Swarm, and not running Testground daemon remotely (we have playbooks for that at https://github.com/ipfs/testground/pull/217 which is not yet in `master`).

This is basically option 3 from above. (Currently there is not much value in running a Testground daemon remotely - it only introduces indirection).

Therefore the command for SSH tunnelling is still relevant, and I will move it to an appropriate place, before merging this (after solving the conflicts).

PR #277, opened 2019-12-14
initial tutorial on using Testground with k8s/DO

nonsense:
This is the initial tutorial on using Testground with Kubernetes on Digital Ocean.

Ideally it should be as brief as possible and not verbose, while at the same time providing all the details a developer needs.

I want to avoid having to document knowledge that 99% of the company has, or knowledge that can easily be found with a single Google Search, while staying on point.

Any feedback is very welcome ;)

### DIFF
--- a/infra/k8s/README.md
+++ b/infra/k8s/README.md
@@ -1,0 +1,56 @@
+# Running Testground with Kubernetes on Digitial Ocean backend
+
+## Requirements
+
+1. Digital Ocean account
+2. [Docker Hub](https://hub.docker.com/) account
+3. [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
+4. Docker
+5. [helm](https://github.com/helm/helm)
+
+## Background
+
+This document is explaining how to execute a Testground test plan locally, while Testground targets a Kubernetes cluster on Digital Ocean for execution of all tasks. Images are built locally by Testground via Docker, and pushed to Docker Hub. Redis is to be installed manually on Kubernetes, prior to running any test plans.
+
+## Setup Kubernetes Cluster on Digital Ocean
+
+1. Create a Kubernetes cluster on Digital Ocean, with the necessary number of workers.
+
+Note: There is a limit of 100 pods per worker, so calculate the number of workers you need accordingly.
+
+2. Configure your `kubectl` context and authentication with Digital Ocean. Confirm that `kubectl get nodes` shows the number of nodes in your cluster.
+
+```
+$  kubectl get nodes
+NAME                                                STATUS   ROLES    AGE     VERSION
+gke-standard-cluster-1-default-pool-bce3c7de-0f20   Ready    <none>   4h46m   v1.14.8-gke.17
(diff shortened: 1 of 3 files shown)


infra/k8s/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+This document is explaining how to execute a Testground test plan locally, while Testground targets a Kubernetes cluster on Digital Ocean for execution of all tasks. Images are built locally by Testground via Docker, and pushed to Docker Hub. Redis is to be installed manually on Kubernetes, prior to running any test plans.
+
+## Setup Kubernetes Cluster on Digital Ocean
+
+1. Create a Kubernetes cluster on Digital Ocean, with the necessary number of workers.
+
+Note: There is a limit of 100 pods per worker, so calculate the number of workers you need accordingly.
+
+2. Configure your `kubectl` context and authentication with Digital Ocean. Confirm that `kubectl get nodes` shows the number of nodes in your cluster.
```
hacdias: I'm reviewing this from the POV of a Kubernetes noob (me!): am I supposed to run this in DO or locally?

infra/k8s/README.md
```diff
@@ -0,0 +1,68 @@
...
+gke-standard-cluster-1-default-pool-bce3c7de-gprr   Ready    <none>   4h47m   v1.14.8-gke.17
+gke-standard-cluster-1-default-pool-bce3c7de-sbdv   Ready    <none>   4h46m   v1.14.8-gke.17
+```
+
+## Setup Testground remote dependencies on your cluster
+
+1. Create a `Redis` service on your Kubernetes cluster
+
+```
+helm install redis stable/redis --values redis-values.yaml
```
hacdias: is this run locally or somewhere else?

nonsense:
@hacdias I think it is great that you haven't used Kubernetes for the purpose of this task - if something is not clear, or you can't very quickly figure it out (talking 1 Google Search of smth), then just ask, and let's update the docs.

infra/k8s/README.md
nonsense: Yes. `helm` is a Kubernetes package manager. You run it locally and based on defaults + your specific `redis-values.yaml`, it creates resources on Kubernetes. Will add a bit more info to this step, but ultimately we have to automate it (it is in the `known issues` section).

This is part of the infrastructure that developers should take for granted on not have to deal with at all, but at this stage of the project (`self-serviced Testground`) they have to create it.

infra/k8s/README.md
nonsense: `kubectl` is a control tool for Kubernetes, that you run locally. It has to be configured with any given Kubernetes cluster.

This section is sparse on purpose because configuring `kubectl` with a cluster is mostly different for every single provider - AWS, DO, GCP, etc.

nonsense: @hacdias I added an `Introduction` section with some basic intro to the requirements/tools, let me know if you have any questions, and it you think the descriptions now are enough for people with very little Kubernetes experience :)

PR #281, opened 2019-12-16
daemon refactor

raulk:
* No implicit embedded run mode. Makes running the daemon compulsory.
* Removes the embedded testground run mode when the `[client]` config section is absent, to avoid confusion.
* Introduces sane endpoint defaults, defaulting to `http://localhost:8042`.
* Introduces global `--endpoint` CLI flag to override the endpoint from `.env.toml`.
* Adds a section in the usage doc.
* Wraps the out `Writer` within a flusher.
* Changes default log level to `INFO`.
* Enhances error output in test failures.
* Refactors the cmd integration tests to start an ephemeral daemon.


pkg/tgwriter/tgwriter.go
```diff
@@ -12,7 +13,7 @@ func New(w http.ResponseWriter, log *zap.SugaredLogger) *TgWriter {
 	w.Header().Set("Content-Type", "application/json")
 
 	return &TgWriter{
-		output: w,
+		output: ioutils.NewWriteFlusher(w),
```
nonsense: This is already done at a few other places. If we are going to wrap the `TgWriter` with a `Flusher` by default, we should remove the occurrence at the rest of the places in the codebase.

cmd/daemon.go
```diff
@@ -27,17 +28,17 @@ func daemonCommand(c *cli.Context) error {
...
+		listen = envcfg.Daemon.Listen
+		srv    = server.New(listen)
+	)
 
 	done := make(chan os.Signal, 1)
 	signal.Notify(done, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
 
 	go func() {
-		logging.S().Infow("listen and serve", "addr", srv.Addr)
+		fmt.Printf("daemon listening on addr: %s\n", srv.Addr)
```
nonsense: Why `fmt.Printf` and not logging? We are lacking the timestamp for example, which would be useful when we run this as a service.

Also we are substituting structured logging for non-structured, which brings less value.

docs/USAGE.md
```diff
@@ -54,8 +58,56 @@ In order to build test plans, Testground needs to know where its source director
...
 
+## Starting a testground daemon
+
+---
+
+⚠️ In the past, if the `[client]` section in the `.env.toml` file was not set
+up, we would start an embedded daemon to service the request. To avoid
+confusion, we no longer do that.
+
+---
```
nonsense: We don't need this section. This was the case for barely a week or two, probably nobody even used it that way (as Testground is not really usable right now, so we just generate docs for technical debt that we incurred during iteration.

docs/USAGE.md
```diff
@@ -54,8 +58,56 @@ In order to build test plans, Testground needs to know where its source director
...
+
+Testground has a daemon/client architecture:
+
+* The daemon performs the heavy-lifting. It populates a catalogue of test plans,
+  performs builds, and schedules runs, amongst other things.
+    * The daemon is intended to run in a server setting, but can of course be
+      run locally in a fashion similar to the IPFS daemon/client CLI. 
+    * The daemon exposes an HTTP API to receive client commands.
+* The client is a lightweight CLI tool that sends commands to the daemon via its
+  HTTP API.
```
nonsense: This section is duplicating the `OVERVIEW.md` - if we want to merge this architecture part here, we should remove `OVERVIEW.md`.

nonsense: I think we should just remove it from here, and add architectural docs under other doc, and not under USAGE.md - USAGE.md is our `Getting Started` guide, and should be minimal I think.

pkg/daemon/client/client.go
```diff
@@ -25,6 +25,8 @@ type Client struct {
 
 // New initializes a new API client
 func New(endpoint string) *Client {
+	fmt.Printf("client initialized against addr: %s\n", endpoint)
```
nonsense: We should be consistent with logging, and stop having `fmt.Println` that people can't enable and disable. DEBUG, INFO, WARN, ERROR gives us that flexibility, no? Also this is not structured, in the future we might want to have tests for important log lines and get some of the events they generate.

cmd/daemon.go
raulk: Because I want this to be printed out unconditionally, and we run the logger with `WARN` level by default.

docs/USAGE.md
raulk: Fine with removing it -- I wanted to be nice to our past citizens.

pkg/daemon/client/client.go
nonsense: This particular one should be `DEBUG` in my opinion.

raulk: For the reason I stated above. This is really _pre-initialisation_/prelude logging. Honestly, I hated doing this, but there's no way I can unconditionally print a log statement with zap. Unless we create a custom logging level? I don't have time to look into that; if you do, be my guest ;-)

pkg/daemon/client/client.go
```diff
@@ -25,7 +28,7 @@ type Client struct {
 
 // New initializes a new API client
 func New(endpoint string) *Client {
-	fmt.Printf("client initialized against addr: %s\n", endpoint)
+	logging.S().Infow("client initialized", "addr", endpoint)
```
nonsense: This is a bit ambiguous, because we have many clients:
1. testground client, talking to testground daemon
2. docker client, talking to docker daemon
3. kubectl client, talking to kubectl master

Additional context will be helpful, for example if we enable the caller (pkg, file, line number), or changing to log to: `testground client initialised`.

cmd/daemon.go
nonsense: We both agree that logging can be a religious topic, just sharing my opinion on this religion, and maybe we can reach a common ground where both are happy and logging is sane for a large amount of people :)

pkg/daemon/client/client.go
```diff
@@ -25,6 +28,8 @@ type Client struct {
 
 // New initializes a new API client
 func New(endpoint string) *Client {
+	logging.S().Infow("client initialized", "addr", endpoint)
```
nonsense:
```suggestion
	logging.S().Infow("testground client initialized", "addr", endpoint)
```

pkg/tgwriter/tgwriter.go
hacdias: Was this fixed already? I haven't seen the changes in the code, such as in https://github.com/ipfs/testground/blob/489c147b07706349b2846a5b11824b6b1c82a96a/pkg/server/run.go#L41

docs/USAGE.md
raulk: Over-documenting is a nice problem to have. Let's go that path now, and dry down/consolidate/refine/polish the docs later.

nonsense: OK, fair enough.

docs/USAGE.md
nonsense: This is still there though.

PR #286, opened 2019-12-16
add REDIS_HOST to sidecar

nonsense:
This PR is add REDIS_HOST as an ENV VAR when we setup infrastructure for Docker Swarm backend for Testground. Not sure if we will ever use these playbooks, but better keep them accurate when we can.

Related to https://github.com/ipfs/testground/issues/285

--

Regarding other `runners`:
1. `k8s` - REDIS_HOST will be configured within the resource that adds the sidecar to every machine, currently a `daemonset` which is defined at `infra/k8s/sidecar.yaml`
2. `local` ? If we run `testground` locally on Linux, we should be able to start `sidecar` as well. Who does that? Do we have any such scripts or code?

### DIFF
--- a/infra/docker-swarm/ansible/roles/testground-services/tasks/main.yaml
+++ b/infra/docker-swarm/ansible/roles/testground-services/tasks/main.yaml
@@ -31,6 +31,8 @@
     args:
       - sidecar
       - --runner docker
+    env:
+      REDIS_HOST: "testground-redis"
     networks:
       - control
     placement:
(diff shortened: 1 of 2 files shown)


jimpick:
@nonsense I was hitting the same problem with the local:docker runner ... I added a commit to fix it.

nonsense:
@jimpick thanks! Understood the problem now, and this PR is showing why I was confused - we don't have a single actor that is responsible for starting sidecar right now - it is started either off-band from the Ansible playbooks, or if it is missing it is started by `testground` itself. I'd say one of this is probably redundant, but we can look into it in the future.

Sorry for breaking it, I didn't realise we start sidecar from within `testground`, I was under the impression that testground's deps are managed elsewhere (sidecar, redis, etc.)

PR #302, opened 2019-12-17
feat: use manifest default parameters

### LINKED ISSUE #264: Manifest Parameters not used
Right now, we have test parameters on the manifest with some information, such as name, type, description and default value. However, those are not used anywhere and we need to set the defaults twice (on the manifest and the code itself). It would be much better if we were using that work.

Also, the describe command could also print the parameters...

hacdias:
This closes #264 by using the default parameters from the manifest file instead of having them in-code. In the future, we can also use the type information in the manifest to validate the parameters.

This is almost done but I have a question where I'd like feedback on:
- By removing `ParamD` functions, we only get `Param` functions which return `paramType, ok`. This would require us to check if they're all okay in the code. Should we do that? Or change those functions to not return the `ok` part?

/cc @daviddias @nonsense @raulk @Stebalien


plans/chew-datasets/utils/testconfig.go
```diff
@@ -26,59 +26,58 @@ func GetTestConfig(runenv *runtime.RunEnv, acceptFiles bool, acceptDirs bool) (c
...
-				return cfg, err
-			}
-
-			file, err := CreateRandomFile(runenv, os.TempDir(), int64(n))
-			if err != nil {
-				return nil, err
+		sizes, ok := runenv.StringArrayParam("file-sizes")
+		if ok {
+			for _, size := range sizes {
+				n, err := humanize.ParseBytes(size)
```
raulk: Can we move the logic to parse:
* sizes
* slices of types

to `sdk/runtime`? That way they'll be available to all test plans.

plans/chew-datasets/utils/testconfig.go
```diff
@@ -26,59 +26,58 @@ func GetTestConfig(runenv *runtime.RunEnv, acceptFiles bool, acceptDirs bool) (c
...
+				if err != nil {
+					return cfg, err
+				}
+
+				file, err := CreateRandomFile(runenv, os.TempDir(), int64(n))
```
raulk: Move the random file creation function to `sdk/runtime`? I think `smlbench` also creates random files, so we should refactor both to use a common util function.

plans/chew-datasets/utils/testconfig.go
```diff
@@ -26,59 +26,58 @@ func GetTestConfig(runenv *runtime.RunEnv, acceptFiles bool, acceptDirs bool) (c
...
-			if err != nil {
-				return nil, err
+		if ok {
+			for _, dir := range dirConfigs {
+				n, err := humanize.ParseBytes(dir.Size)
+				if err != nil {
+					return nil, err
+				}
+
+				path, err := CreateRandomDirectory(runenv, os.TempDir(), dir.Depth)
```
raulk: Can also be moved to `sdk/runtime`.

manifests/chew-datasets.toml
```diff
@@ -37,23 +37,40 @@ name = "ipfs-add-defaults"
 instances = { min = 1, max = 1, default = 1 }
 
   [testcases.params]
-  some_param = { type = "int", desc = "some param", unit = "peers" }
+    [testcases.params.file-sizes]
+      type = "string array"
+      desc = "An array of File Sizes to be tested"
+      unit = "SI Units"
+      default = ["1MB", "10MB", "100MB"]
```
raulk: I think this is a TOML array, but below (Δ) you parse this as JSON?

pkg/engine/engine.go
```diff
@@ -282,6 +283,17 @@ func (e *Engine) DoRun(ctx context.Context, testplan string, testcase string, ru
 		return nil, fmt.Errorf("error while generating test run ID: %w", err)
 	}
 
+	// Add the default parameters set on the manifest.
+	for paramName, paramValue := range tcase.Parameters {
+		if _, ok := input.Parameters[paramName]; !ok {
+			data, err := json.Marshal(paramValue.Default)
```
raulk: (Δ)

plans/bitswap-tuning/test/transfer.go
```diff
@@ -29,11 +29,13 @@ var RootCidSubtree = &sync.Subtree{
 // Transfer data from S seeds to L leeches
 func Transfer(runenv *runtime.RunEnv) {
 	// Test Parameters
-	timeout := time.Duration(runenv.IntParamD("timeout_secs", 60)) * time.Second
-	leechCount := runenv.IntParamD("leech_count", 1)
-	passiveCount := runenv.IntParamD("passive_count", 0)
-	requestStagger := time.Duration(runenv.IntParamD("request_stagger", 0)) * time.Millisecond
-	fileSize := runenv.IntParamD("file_size", 200*1024*1024)
+	timeoutRaw, _ := runenv.IntParam("timeout_secs")
```
raulk: Can we keep the old style of one-line value loading, to avoid polluting the local variable namespace?

pkg/engine/engine.go
hacdias: I am marshalling this as a JSON value so it can be used on the environment variables passed to docker so we can get them afterwards using `GetCurrentRunEnv`. This seemed to be the easier way to support this kind of things. If you have a different suggestion, I'll be glad to hear it :D

plans/bitswap-tuning/test/transfer.go
hacdias: We can, and that's what I asked feedback on: should we change this function (and the others) to return only the value, or create a `Must...` function for each that returns only the value?

My suggestion would be: change the definitions to only return the value _and_  create a new function that lets us verify if the value is set!

manifests/chew-datasets.toml
hacdias: See reply below.

plans/bitswap-tuning/test/transfer.go
hacdias: Decided to panic in case there are invalid parameters. A test can't run with invalid parameters anyway.

plans/chew-datasets/utils/testconfig.go
hacdias: Done!

hacdias:
Done @raulk!

sdk/runtime/runenv.go
```diff
@@ -204,93 +205,91 @@ func ParseRunEnv(env []string) (*RunEnv, error) {
...
+	}
+	return v
 }
 
-// StringParamD returns a string parameter, or the default value if not set.
-func (re *RunEnv) StringParamD(name string, def string) string {
-	s, ok := re.StringParam(name)
-	if !ok {
-		return def
+func (re *RunEnv) BytesParam(name string) uint64 {
```
Stebalien: Let's call this `SizeParam`. In go, `BytesX` usually means "returns something as `[]byte`".

sdk/runtime/runenv.go
```diff
@@ -204,93 +205,91 @@ func ParseRunEnv(env []string) (*RunEnv, error) {
...
-// StringArrayParamD returns an array of string parameter or the default one
-// if the parameter was not set or is invalid.
-func (re *RunEnv) StringArrayParamD(name string, def []string) []string {
-	a := []string{}
-	if ok := re.JSONParam(name, &a); !ok {
-		return def
+// BytesArrayParam returns an array of uint64 elements which represent sizes,
+// in bytes. If the response is nil, then there was an error parsing the input.
+// It panics on error.
+func (re *RunEnv) BytesArrayParam(name string) []uint64 {
```
Stebalien: Ditto.

nonsense:
I think this might be breaking many of the tests - there are still many references to `runenv.IntParamD` for example, and this no longer exists - `plans/dht/test/find_peers.go` `provide_stress.go` `find_providers.go`, etc.

nonsense:
At some point we were running the `find-peers` dht test if I remember correctly as part of CI, but it seems we have stopped that. We should add compilation for `production` test plans, those that we think are mostly ready as part of CI.

raulk:
Fixed in #330!

PR #311, opened 2019-12-18
Limit Docker Swarm network names to 63 chars

jimpick:
Most tests will fail because the generated network names are more than 63 characters.

This fix truncates the RunID if there aren't enough characters.

### DIFF
--- a/pkg/runner/cluster_swarm.go
+++ b/pkg/runner/cluster_swarm.go
@@ -117,8 +117,22 @@ func (*ClusterSwarmRunner) Run(ctx context.Context, input *api.RunInput, ow io.W
 		return nil, err
 	}
 
+	const maxNameLength = 63
+	var shortRunID string
+	prefix := fmt.Sprintf("tg-%s-%s-", input.TestPlan.Name, testcase.Name)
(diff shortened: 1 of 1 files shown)


pkg/runner/cluster_swarm.go
```diff
@@ -117,8 +117,22 @@ func (*ClusterSwarmRunner) Run(ctx context.Context, input *api.RunInput, ow io.W
...
+	suffix := "-default"
+	availableChars := maxNameLength - len(prefix) - len(suffix)
+	if availableChars < 4 {
+		return nil, fmt.Errorf("test plan name + test case name too long")
+	}
+	if len(input.RunID) > availableChars {
+		shortRunID = input.RunID[len(input.RunID)-availableChars:]
+	} else {
+		shortRunID = input.RunID
+	}
```
nonsense: It'd be nice if this is extracted into a function that returns a short name that adheres to Docker Swarm's requirements of 63 chars, but I guess for now that works as well, as long as we don't start using it anywhere else :)

Stebalien:
@jimpick could you merge this when you're happy with it?

PR #313, opened 2019-12-18
fix sidecar container and overlay network

nonsense:
Some Ansible improvements and bug fixes we found today:
1. We can't run `sidecar` as a service, because we need to set up the host process namespace, capabilities, etc. - so we run it as a container.
2. Due to 1, we have to create the overlay network on all worker nodes. This is why we run a dummy `sleep` container to propagate the overlay network to all workers. Docker overlay networks are propagated lazily until a service needs them, and here we need the `control` network in the `sidecar` container.
3. Fix on getting the `Leader` IP - this was racy, and worked only if the leader hadn't added any workers.

### DIFF
--- a/infra/docker-swarm/ansible/.gitignore
+++ b/infra/docker-swarm/ansible/.gitignore
@@ -1,3 +1,5 @@
-inventories/
+inventories
 !inventories/group_vars/all.yaml
+roles/external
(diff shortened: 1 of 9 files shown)


nonsense:
Merging this because it is obvious master is broken. If there are any comments, let's address in follow-up PRs.

PR #314, opened 2019-12-18
different address range for every testplan network

nonsense:
This is very hacky and I wouldn't want to merge it, we need to keep track of allocated networks to testplans and get a `free` range when we create a new network.

### DIFF
--- a/pkg/runner/cluster_swarm.go
+++ b/pkg/runner/cluster_swarm.go
@@ -8,7 +8,6 @@ import (
 	"reflect"
 	"time"
 
-	"github.com/docker/docker/api/types/network"
 	"github.com/ipfs/testground/pkg/api"
 	"github.com/ipfs/testground/pkg/aws"
 	"github.com/ipfs/testground/pkg/logging"
@@ -17,6 +16,7 @@ import (
 
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/filters"
+	"github.com/docker/docker/api/types/network"
 	"github.com/docker/docker/api/types/swarm"
 	"github.com/docker/docker/client"
 	"github.com/docker/docker/pkg/stdcopy"
@@ -133,8 +133,20 @@ func (*ClusterSwarmRunner) Run(ctx context.Context, input *api.RunInput, ow io.W
 		return nil, fmt.Errorf("testground-redis service doesn't exist in the swarm cluster; aborting")
 	}
 
+	// We can't create a network for every testplan on the same range,
+	// so we check how many networks we have and decide based on this number
+	networks, err := cli.NetworkList(ctx, types.NetworkListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	currentNetworkId := 10 + len(networks)
+
+	subnet := fmt.Sprintf("10.%d.0.0/16", currentNetworkId)
+	gateway := fmt.Sprintf("10.%d.0.1", currentNetworkId)
+
 	// Create the data network.
-	log.Infow("creating data network", "name", sname)
+	log.Infow("creating data network", "name", sname, "subnet", subnet)
 
 	networkSpec := types.NetworkCreate{
 		Driver:         "overlay",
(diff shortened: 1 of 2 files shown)


nonsense:
On a second thought, this is actually better than `master` so probably worth merging and iterating on improvements.

pkg/runner/cluster_swarm.go
```diff
@@ -219,7 +231,7 @@ func (*ClusterSwarmRunner) Run(ctx context.Context, input *api.RunInput, ow io.W
 				},
 			},
 			Placement: &swarm.Placement{
-				MaxReplicas: 10000,
+				MaxReplicas: 100,
```
nonsense: Changing this to 100 - originally I thought this applies to the `service` as a whole, but this is a limit per worker.

```
d21soxnh35lk        tg-dht-find-peers-4bd9c685-da5e-45d3-a93d-805d15d584b2   replicated          30/500 (max 10000 per node)   909427826938.dkr.ecr.eu-central-1.amazonaws.com/testground-dht:41fcc08e-4fb0-4d56-b11b-ac2c39d5edba
```

Having killed a few workers due to 100% CPU, I suggest we stay at 100 containers as a start before we know who much resources each container needs.

pkg/runner/cluster_swarm.go
```diff
@@ -133,8 +133,20 @@ func (*ClusterSwarmRunner) Run(ctx context.Context, input *api.RunInput, ow io.W
 		return nil, fmt.Errorf("testground-redis service doesn't exist in the swarm cluster; aborting")
 	}
 
+	// We can't create a network for every testplan on the same range,
+	// so we check how many networks we have and decide based on this number
```
Stebalien: * Can we choose based on how many _testground_ networks we have?
* Can we allocate from `16.0.0.0/4`? That gives us 16.0.0.0 -> 31.0.0.0.
  * Public IP addresses.
  * We're not going to run out after 256 runs.

jimpick:
It doesn't even really make sense to me that Docker shouldn't be able to support overlapping subnets on completely separated overlay networks, but that's what they've done.

* https://github.com/moby/moby/issues/26912
* https://github.com/docker/libnetwork/issues/2246#issuecomment-410777061

jimpick:
I applied this and it enabled my scheduled Docker Swarm services for my test case to run again. :-)

It conflicts with PR #271 which tries to set the data subnet to "8.0.0.0/5" ... but I couldn't get that to start running in Docker Swarm.

pkg/runner/cluster_swarm.go
nonsense: @raulk @Stebalien given that we run only one testplan at a time, do we actually need multiple testplan networks? If we start enforcing the `one concurrent testplan run`, and make sure we clean up after the testplan, we could simplify this.

nonsense: @Stebalien I am confused about what you're asking - `16.0.0.0 -> 31.0.0.0` - isn't this public space? Nevertheless pushed a commit with it.

Stebalien: > @raulk @Stebalien given that we run only one testplan at a time, do we actually need multiple testplan networks? If we start enforcing the one concurrent testplan run, and make sure we clean up after the testplan, we could simplify this.

Well, I think the goal was to be able to run multiple at the same time.

> @Stebalien I am confused about what you're asking - 16.0.0.0 -> 31.0.0.0 - isn't this public space? Nevertheless pushed a commit with it.

Yes. Libp2p may choose to behave differently when it doesn't have a publicly routable IP address. Ideally, every instance would have two interfaces with two IP addresses (and would be able to turn them on/off with the sidecar).

Stebalien:
> It conflicts with PR #271 which tries to set the data subnet to "8.0.0.0/5" ... but I couldn't get that to start running in Docker Swarm.

That PR is blocked on something like this PR. That PR also does some network fiddling (disconnect/reconnect) that worked in my local docker swarm tests (manually) but needs more testing on cloud infrastructure.

Stebalien:
I'm going to merge this as "strictly better" than what we have (so I can unblock #271).

PR #317, opened 2019-12-20
Kubernetes cluster setup with kops and Flannel on AWS

nonsense:
This PR contains the playbooks to setup a Kubernetes cluster with `kops` and create two networks on it with Flannel - a default Kubernetes network (100.96.0.0/11 as default pod CIDR range used by `kops`), and a secondary overlay network (10.0.0.0/8).

This PR is the result of our effort to evaluate Kubernetes for an orchestrator for Testground - https://github.com/ipfs/testground/issues/295


jimpick:
I followed the instructions and now I have a Kubernetes cluster to play with!

(I initially used the wrong version of kops, but that's my bad, since it's documented in the README)

nonsense:
Since the docs are good enough for @jimpick to run, I think we can merge this and iterate on it, so that we have a clear picture of the `infra` part in terms of whats the configuration is for:
- Docker Swarm on AWS
- Kubernetes on AWS

We will have to converge to an orchestrator at some point, and then we can remove all the ununsed playbooks.

@jimpick I am not sure what the implications are for increasing the max number of pods per VM (which by kops is 110), but since you already changed it, and since our networking generally allows for at least 250 (if we use a `SubnetLen` of 24 per host), maybe we can bump that number here and try it out?

nonsense:
cc @lanzafame adding you as well, as I know you have Kubernetes experience and would love to hear your feedback.

jimpick:
@nonsense I think it probably makes sense to stay with the default pods per node (110). We could maybe document how to increase it somewhere. It worked for my particular basic-tcp/upload test because I took extra care to ensure that only one upload would occur at once across the entire cluster, so it was pretty light on the CPU. Even then, I eventually ended up using a smaller instance (m5a.large) with 2 vCPUs, with about 140 pods per node. I think most tests are probably going to require much more CPU and so running so many pods won't be feasible.

To update, do `kops edit cluster` and add the following entry:

```
spec:
  kubelet:
    maxPods: 200
```

PR #323, opened 2020-01-07
README.md for kops/weave/flannel/cnigenie

nonsense:
Here is a README which goes through a setup of Kubernetes cluster with `kops` and Flannel and Weave for networks.

It takes about 10-15min to go over it and have a working cluster.

There are a few known issues that would be good to have a second pair of eyes on, in case I am doing something stupid. ~I have tried to dynamically attach/detach network interfaces with `weave` to pods, but they are not accessible.~ I believe this must be possible because Kubernetes / CNI-Genie / Weave CNI must be setting up those interfaces somehow correctly when the annotations are present.

---

Here is a demo of this setup in case you find this useful:
1. Download https://drive.google.com/open?id=1p8RqDv7tyqJH0V6uOpLEtBvaa_fU2zjg
2. `asciinema play -s 3 ./demo2.cast`


nonsense:
@lanzafame adding you as well in case you have some insight into the `known issues` part.

nonsense:
Known issues have been addressed by disabling the `weave-npc` as the weave network policy controller is picking up the PodIP when adding rules to `ipset` (used by `iptables`), and this is not OK when the PodIP is set to an IP within the flannel network.

nonsense:
Merging this as this is pretty much the setup we used with @Stebalien to try `sidecar` with `k8s`.

The current known issues are:
1. When using the `host network`, for some reason there is a missing `ip route` to the Kubernetes Services CIDR. I opened an issue/question for this one at https://github.com/coreos/flannel/issues/1243 - not sure if this is a bug within `kops` or `flannel`, or we are misconfiguring something on our side. For now we add these routes manually (or via an init container after Flannel is installed).

2. We can't allocate IPs because the Weave CNI doesn't support that. We have to submit a PR or fork it, or use the `weave` cmdline utility. Technically it is possible to allocate a specific IP with Weave, but not through the current version of the Weave CNI.

---

Will follow up with additional PR, where we improve docs, and add code to support sidecar k8s runner.

PR #324, opened 2020-01-07
codecov set informational flag to true

nonsense:
From the docs, pointed out by @raulk

_informational_
Use Codecov in informational mode. Default is false. If true is specified the resulting status will pass no matter what the coverage is or what other settings are specified. Informational mode is great to use if you want to expose codecov information to other developers in your pull request without necessarily gating PRs on that information.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -8,6 +8,7 @@ coverage:
         # project coverage decrease by more than 5%:
         target: auto
         threshold: 5%
+        informational: true
     patch:
       default:
         # Be tolerant on slight code coverage diff on PRs to limit


nonsense:
Offtopic, but related to CI: As a next step we should review `golangci-lint run` - I had to restart 2 PRs already because it can't finish in 60s or 80s and throws errors. Not sure why, as locally it runs within a second.

nonsense:
I guess this config is not right, as I am still getting red builds, for example https://github.com/ipfs/testground/pull/275 (rebased on `master`)

PR #330, opened 2020-01-09
add dht test plan compilation to tests

nonsense:
This PR is adding compilation of `dht test` as part of CI. We should be doing this for every test that we consider ready and maintained, at least for now, when tests are somewhat part of the `testground` codebase.

It is also fixing the code that gets the runtime config as we no longer have methods with default parameters - they are now taken from the `toml` config.

### DIFF
--- a/manifests/dht.toml
+++ b/manifests/dht.toml
@@ -39,7 +39,7 @@ instances = { min = 16, max = 10000, default = 16 }
   n_bootstrap   = { type = "int", desc = "number of bootstrap nodes", unit = "int", default = "1" }
   bucket_size   = { type = "int", desc = "routing table bucket size", unit = "peers", default = "2" }
   n_find_peers  = { type = "int", desc = "number of peers to find", unit = "peers", default = "1" }
-  timeout_secs  = { type = "int", desc = "test timeout", unit = "seconds", default = "30" }
+  timeout_secs  = { type = "int", desc = "test timeout", unit = "seconds", default = "60" }
 
 # seq 1
 [[testcases]]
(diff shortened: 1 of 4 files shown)


manifests/dht.toml
```diff
@@ -39,21 +39,25 @@ instances = { min = 16, max = 10000, default = 16 }
   n_bootstrap   = { type = "int", desc = "number of bootstrap nodes", unit = "int", default = "1" }
   bucket_size   = { type = "int", desc = "routing table bucket size", unit = "peers", default = "2" }
   n_find_peers  = { type = "int", desc = "number of peers to find", unit = "peers", default = "1" }
-  timeout_secs  = { type = "int", desc = "test timeout", unit = "seconds", default = "30" }
```
nonsense: I assume the timeout in the code was the source of truth.

manifests/dht.toml
```diff
@@ -39,21 +39,25 @@ instances = { min = 16, max = 10000, default = 16 }
...
+  timeout_secs  = { type = "int", desc = "test timeout", unit = "seconds", default = "60" }
 
 # seq 1
 [[testcases]]
 name = "find-providers"
 instances = { min = 16, max = 250, default = 16 }
 
   [testcases.params]
```
nonsense: Looks like params in `dht.toml` `find-providers` diverged from the params in the code. Not sure how correct they are now, but I tried to align them.

nonsense:
Merging this as `dht test` is broken on master and I think there won't be any contention here. If there is, I will be happy to revert, or address comments in follow up PR.

PR #347, opened 2020-01-16
use cfg.Network instead of netId

nonsense:
@Stebalien I haven't verified this one, I am focusing on the rest of the issues - probably worth running it, before we merge it (unless you are confident enough that this is a bug).

---

TODO:
- [ ] confirm this is actually the desired behaviour


Stebalien:
It was a bug, but I forgot one additional fix.

Stebalien:
(also, CI runs this)

PR #349, opened 2020-01-16
replace Digital Ocean README with AWS kops README

nonsense:
This PR should be merged after https://github.com/ipfs/testground/pull/336

It is replacing the main README.md (which had docs on how to setup a cluster on Digital Ocean) with a README.md on how to setup a cluster on AWS with `kops`, since we decided a while ago to focus on self-managed k8s clusters due to flexibility with underlying CNI plugins.

---

I am also removing the AWS Terraform playbooks for the EKS setup, as this is also a managed Kubernetes environment.

I am also removing the setup consisting of two Flannel networks, as it looks like Weave is a better fit for now. If we encounter problems with Weave in the future, we can always refer back to the two Flannel networks setup if necessary.

---

TODO:
- [ ] 1. Go over the README.md with someone else from the team - @hacdias or @raulk or @Stebalien just so that we are on the same page regarding where we are at with infrastructure, and to confirm there are no missing steps.
- [ ] 2. Once 1 is done, maybe do a demo for @dirkmc , @aschmahmann and other people, and gather more feedback.

---

cc @jimpick you might also be interested in this!


infra/k8s/README.md
```diff
@@ -1,39 +1,97 @@
...
+.
+├── README.md
+└── kops-weave                     # Kubernetes resources for setting up networking with Weave and Flannel
+```
+
+## Introduction
+
+Kubernetes Operations (kops) is a tool which helps to create, destroy, upgrade and maintain production-grade Kubernetes clusters from the command line. We use it to create a k8s cluster on AWS.
+
+We use CoreOS Flannel for networking on Kubernetes - for the default Kubernetes network
```
Stebalien: Let's talk about this in terms of testground. I.e., which network do we use for the control network, which for the data network, etc.

infra/k8s/README.md
```diff
@@ -1,39 +1,97 @@
...
-pool-jyub1rb95-snl5   Ready    <none>   60m   v1.16.2
-pool-jyub1rb95-snl6   Ready    <none>   60m   v1.16.2
+kubectl apply -f ./weave.yml
 ```
 
+10. Destroy the cluster when you're done working on it
+
+```
+kops delete cluster $NAME --yes
+```
```
Stebalien: Can we create a script for all this?

Stebalien: I.e., a "cluster-create.sh" and a "cluster-delete.sh"?

nonsense: Yes, we should automate it further. The reason I haven't done so far is, because we need to wait for given states on Kubernetes before running commands (for example wait for master to come up as `NotReady` before installing `flannel`, wait for `flannel` to be installed, before installing `genie`, etc.)

Still this needs to be simpler.

PR #352, opened 2020-01-17
Traffic shaping for bitswap plan

dirkmc:
Allows client to set latency and bandwidth as test parameters

### DIFF
--- a/plans/bitswap-tuning/test/transfer.go
+++ b/plans/bitswap-tuning/test/transfer.go
@@ -36,18 +36,8 @@ func Transfer(runenv *runtime.RunEnv) {
 	fileSize := runenv.IntParam("file_size")
 
 	/// --- Set up
-
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
-
-	node, err := utils.CreateNode(ctx, runenv)
-	if err != nil {
-		runenv.Abort(err)
-		return
-	}
-	defer node.Close()
-
-	runenv.Message("I am %s with addrs: %v", node.Host.ID(), node.Host.Addrs())
 
 	watcher, writer := sync.MustWatcherWriter(runenv)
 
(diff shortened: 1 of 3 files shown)


plans/bitswap-tuning/utils/net.go
```diff
@@ -0,0 +1,48 @@
...
+
+	// TODO: just put the hostname inside the runenv?
+	hostname, err := os.Hostname()
+	if err != nil {
+		return err
+	}
+
+	latency := time.Duration(runenv.IntParam("latency_ms")) * time.Millisecond
+	bandwidth := runenv.IntParam("bandwidth_mb")
+	writer.Write(sync.NetworkSubtree(hostname), &sync.NetworkConfig{
```
raulk: At some point, we'll probably want some API sugar for the network operations, in the same vein as `WaitNetworkInitialized()`.

plans/bitswap-tuning/test/transfer.go
```diff
@@ -170,11 +185,12 @@ func Transfer(runenv *runtime.RunEnv) {
 	runenv.OK()
 }
 
-func setupSeed(ctx context.Context, node *utils.Node, fileSize int) (cid.Cid, error) {
+func setupSeed(ctx context.Context, node *utils.Node, fileSize int, runenv *runtime.RunEnv) (cid.Cid, error) {
```
raulk: Do we intend to use the `runenv` in this function going forward?

plans/bitswap-tuning/utils/net.go
```diff
@@ -0,0 +1,48 @@
...
+	if !runenv.TestSidecar {
+		return nil
+	}
+
+	// Wait for the network to be initialized.
+	if err := sync.WaitNetworkInitialized(ctx, runenv, watcher); err != nil {
+		return err
+	}
+
+	// TODO: just put the hostname inside the runenv?
```
raulk: We can certainly do that with the local runners, but not sure how this would work with k8s. @nonsense?

plans/bitswap-tuning/test/transfer.go
dirkmc: Good catch, I'll remove in the next PR I do

plans/bitswap-tuning/utils/net.go
dirkmc: By the way I copied this from the DHT testplan, as with a bunch of the code in this utils package. It would be good to put it all in a central location somewhere

nonsense: This will be problematic in case we run concurrently the same testplan in multiple namespaces in the future. `pod` names are unique within a `namespace`, but we have `sidecar` running per host right now. I wouldn't worry too much about it, but maybe change the comment to say `unique testplan id`, rather than `hostname`, because as far as I understand we just need a unique testplan identifier here, so that we have a unique subtree, and we use `hostname` as such.

dirkmc: Addressed in https://github.com/ipfs/testground/pull/353

plans/bitswap-tuning/test/transfer.go
dirkmc: Addressed in https://github.com/ipfs/testground/pull/353

PR #366, opened 2020-01-20
changing weave network to 16.0.0.0/4

nonsense:
Changing the `data` network to 16.0.0.0/4 as requested by @Stebalien 

TODO:
- [ ] The k8s runner also needs to be adjusted.

### DIFF
--- a/infra/k8s/kops-weave/weave.yml
+++ b/infra/k8s/kops-weave/weave.yml
@@ -186,6 +186,8 @@ items:
                   value: --log-level=debug
                 - name: EXPECT_NPC
                   value: "0"
+                - name: IPALLOC_RANGE
+                  value: "16.0.0.0/4"
                 - name: HOSTNAME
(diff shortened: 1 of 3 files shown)


Stebalien:
Can we allocate a _specific_ B block for every test case (data network only)? Alternatively, can we get weave to allocate all new machines into a specific range then _move_ these machines to to the appropriate range via the sidecar?

I'd like tests instances to own an entire B block so they can freely re-assign IP addresses from within this block without conflicting.

---

Note: if we can only run one test at a time anyways with our current setup, simply allocating `16.0.0.0/16` is enough.

nonsense:
@Stebalien we have an issue for that - https://github.com/ipfs/testground/issues/357

We should keep track of used networks and decide how to pick a next free B block - either by keeping track on `testground`, or by querying `weave`, or by some other method.

PR #383, opened 2020-01-21
add dummy daemonset

nonsense:
Apparently we must have at least one pod , so that we get an IP address from the default k8s networking (Flannel), so that interface `cni0` is created.

We used this `cni0` interface later in the `initContainer` in `weave.yml` to add a route to all workers so that they can reach the services Kubernetes network (`redis-master`, etc.).

### DIFF
--- a/infra/k8s/README.md
+++ b/infra/k8s/README.md
@@ -79,13 +79,19 @@ kubectl apply -f ./flannel.yml
 kubectl apply -f ./genie-plugin.yaml
 ```
 
-9. Install Weave
+9. Install Dummy daemonset - we need a container on every worker node so that interface `cni0` is created, and Weave's initContainer can add a route to the Services CIDR
+
+```
+kubectl apply -f ./dummy.yaml
+```
+
+10. Install Weave
 
 ```
 kubectl apply -f ./weave.yml
 ```
 
-10. Destroy the cluster when you're done working on it
+11. Destroy the cluster when you're done working on it
 
 ```
 kops delete cluster $NAME --yes
--- a/infra/k8s/README.md
+++ b/infra/k8s/README.md
@@ -128,7 +128,7 @@ testground --vv run network/ping-pong \
     --runner=cluster:k8s \
     --build-cfg bypass_cache=true \
     --build-cfg push_registry=true \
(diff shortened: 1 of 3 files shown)


nonsense:
This is a hack, but I don't really know how to fix this any other way right now. This is needed because of:

```
          - name: iproute-add
            image: busybox
            securityContext:
              privileged: true
            command:
            - sh
            - -ac
            - >
              export GW=$(ip route | grep cni0 | awk '{print $7'}) &&
              echo $GW &&
              ip route &&
              ip route add 100.64.0.0/16 via $GW &&
              ip route || true
```

The issue on Flannel github has no responses so far... `https://github.com/coreos/flannel/issues/1243`.

infra/k8s/README.md
```diff
@@ -79,13 +79,19 @@ kubectl apply -f ./flannel.yml
 kubectl apply -f ./genie-plugin.yaml
 ```
 
-9. Install Weave
+9. Install Dummy daemonset - we need a container on every worker node so that interface `cni0` is created, and Weave's initContainer can add a route to the Services CIDR
```
Stebalien: Let's expand the explanation here. I'm still not sure why this is necessary.

nonsense: If you look at the script from above, you will see that we use `cni0` interface to detect what gateway to use for the Services CIDR (located at 100.64.0.0/16). `cni0` does not exist before you've allocated at least one pod on a given host, so this is what this daemonset is doing.

We could remove the daemonset after that, because it is no longer needed.

nonsense:
@Stebalien I wasn't very clear with this PR, it should have been a draft one probably - I don't want us to merge this hack, but knowing that a few people are testing the setup today, I also want it to work for them.

Ultimately I guess the problem is the way I detect the GW interface though `ip route`s and `cni0` - I will see how we can detect the IP of Flannel without that, which should make it possible to add the route even before there have been any pods on a given host.

infra/k8s/README.md
Stebalien: Ah, now I see. Could we just use the sidecar?

Stebalien: Oh, we can't because that uses host routing. Makes sense.

PR #384, opened 2020-01-21
update redis endpoint from `redis-master` to `redis-headless`

nonsense:
Something is wrong with the current k8s-sidecar runner. Network for `redis-master` (the control network) was not reachable in the `dht/find-peers` test when running the k8s-sidecar runner from `master`.
Resolution:
This is expected actually, because we do not have route for 100.64 services CIDR, we only have route to pod IPs.

---

This is just a heads-up for @dirkmc that you have to use `nonsens3/testground` (which has routes removal part disabled), in case you want to try `sidecar` with k8s today.

Will write follow up when I understand why `redis-master` is not reachable when we remove the routes - this should not be the case.

### DIFF
--- a/infra/k8s/sidecar.yaml
+++ b/infra/k8s/sidecar.yaml
@@ -22,6 +22,7 @@ spec:
       containers:
       - name: testground-sidecar
         image: ipfs/testground:edge
+        imagePullPolicy: Always
         command: ["testground"]
         args: ["--vv", "sidecar", "--runner", "k8s"]
         securityContext:
@@ -30,13 +31,13 @@ spec:
           privileged: true
         env:
         - name: REDIS_HOST
-          value: "redis-master"
+          value: "redis-headless"
         resources:
           limits:
-            memory: 200Mi
+            memory: 1024Mi
           requests:
-            cpu: 100m
-            memory: 200Mi
+            cpu: 200m
+            memory: 1024Mi
         volumeMounts:
           - name: dockersock
             mountPath: "/var/run/docker.sock"
--- a/pkg/runner/cluster_k8s.go
+++ b/pkg/runner/cluster_k8s.go
@@ -30,6 +30,11 @@ var (
 	_ api.Runner = &ClusterK8sRunner{}
 )
 
+func init() {
+	// Avoid collisions in picking up subnets
+	rand.Seed(time.Now().UnixNano())
+}
+
 func homeDir() string {
 	home, _ := os.UserHomeDir()
 	return home
--- a/pkg/runner/cluster_k8s.go
+++ b/pkg/runner/cluster_k8s.go
@@ -103,6 +103,11 @@ func (*ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow io.Wri
 		TestSidecar:        true,
 	}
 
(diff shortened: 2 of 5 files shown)


Stebalien:
> This is just a heads-up for @dirkmc that you have to use nonsens3/testground (which has routes removal part disabled), in case you want to try sidecar with k8s today.

This will make any bitswap tests useless unless we explicitly listen on the data network and avoid listening on the control network (which we can and probably should do anyways).

nonsense:
@Stebalien reverted the change to the sidecar, we still block all routes when initialising the network. The bug was not in the sidecar runner, but rather the fact that there was no route to DNS, and to `redis-master` from testplan.

I shouldn't have changed `redis-headless` to `redis-master`.

pkg/runner/cluster_k8s.go
```diff
@@ -98,7 +99,8 @@ func (*ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow io.Wri
 	}
 
 	var err error
-	_, runenv.TestSubnet, err = net.ParseCIDR("10.33.10.0/24")
+	b := 1 + rand.Intn(200)
```
nonsense: This has to be fixed in a follow up PR.

It is not exactly clear at what point Weave is releasing IP addresses, so even if we remove containers, if we do a second testplan run in the same subnet, we will probably be getting errors right now.

nonsense: Still this is better than just hardcoding a single subnet of length `/24`.

nonsense:
@jimpick you were right - `network/ping-pong` wouldn't work with current `master` and we need this PR so that it works.

pkg/runner/cluster_k8s.go
Stebalien: Does weave ever release it? If not, I'd say this is strictly worse as things will kind of work till they suddenly start failing.

Is there any way to query weave?

(regardless, we need to at least leave a comment).

nonsense: @Stebalien I agree we should fix this, but unless we have some kind of a work-around, we cannot make progress on any of the other issues, unless we fix this right now.

nonsense: Ultimately there is an issue for this, as currently we don't have a way of selecting the next available data network. We don't really store networks in-use anywhere, and we should do that. I believe on the Docker Swarm runner we query `docker networks ls` and check how many networks are present there - we either need something similar for `weave` or `testground` should be keeping track.

nonsense: Added another issue specifically for this: https://github.com/ipfs/testground/issues/392

infra/k8s/sidecar.yaml
```diff
@@ -30,13 +31,13 @@ spec:
           privileged: true
         env:
         - name: REDIS_HOST
-          value: "redis-master"
+          value: "redis-headless"
```
nonsense: This change still assumes that we are going to resolve `redis-headless` only once, and cache the IP. Once we drop the `default` route, we won't be able to resolve it again, until we fix https://github.com/ipfs/testground/issues/390

PR #393, opened 2020-01-22
traces for network/ping-pong

nonsense:
This was helpful when encountering the issue described at https://github.com/ipfs/testground/issues/391

### DIFF
--- a/plans/network/main.go
+++ b/plans/network/main.go
@@ -27,7 +27,7 @@ func main() {
 		return
 	}
 
-	logging.S().Debug("before sync.MustWatcherWriter")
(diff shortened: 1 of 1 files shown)


nonsense:
The logger is not configured properly apparently. `network/ping-pong` sometimes blocks on
`before sync.MustWatcherWriter` or on `before barrier` and it is not clear what the underlying issue is:
1. bug in sidecar in not detecting the container (https://github.com/ipfs/testground/issues/391)
2. some interim failure with `redis` or DNS - there are no error logs on `redis` side though.

Stebalien:
>  logger is not configured properly apparently

Can we fix this or file an issue?

PR #394, opened 2020-01-22
add dns route before removing default route

nonsense:


pkg/sidecar/k8s_instance.go
```diff
@@ -178,6 +182,20 @@ func (d *K8sInstanceManager) manageContainer(ctx context.Context, container *doc
 		}
 
 		if route.Dst == nil && route.Src == nil {
+			// if default route, get the gw and add a route for DNS
+			dnsRoute := route
+			dnsRoute.Src = nil
+			dnsRoute.Dst = &net.IPNet{
+				IP:   net.IPv4(100, 64, 0, 10),
```
Stebalien: Constant?

pkg/sidecar/k8s_instance.go
```diff
@@ -178,6 +182,20 @@ func (d *K8sInstanceManager) manageContainer(ctx context.Context, container *doc
 		}
 
 		if route.Dst == nil && route.Src == nil {
+			// if default route, get the gw and add a route for DNS
```
Stebalien: Any reason we're not doing the same thing we're doing with redis? That is, get the route to the DNS server (regardless of the local gateway)? (i.e., `RouteGet(...)`).

nonsense: No reason, what you suggest should work as well. I just wanted to make sure that the route is added before we remove the `default` route. That might be a candidate for refactor.

PR #397, opened 2020-01-22
improve IP address assignment on k8s

Stebalien:
partial fix for https://github.com/ipfs/testground/issues/392

### DIFF
--- a/infra/k8s/kops-weave/weave.yml
+++ b/infra/k8s/kops-weave/weave.yml
@@ -187,7 +187,7 @@ items:
                 - name: EXPECT_NPC
                   value: "0"
                 - name: IPALLOC_RANGE
-                  value: "10.0.0.0/8"
+                  value: "16.0.0.0/4"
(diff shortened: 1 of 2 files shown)


Stebalien:
This should be a stupid fix for https://github.com/ipfs/testground/pull/366. We _should_ be querying weave, but this will likely "mostly work".

infra/k8s/kops-weave/weave.yml
```diff
@@ -172,7 +172,7 @@ items:
             - sh
             - -ac
             - >
-              export GW=$(ip route | grep cni0 | awk '{print $7'}) &&
```
nonsense: OMG, not sure how this even worked, but I did verify the routes were there by inspecting the logs with `kubectl logs weave-id -c iproute-add`

Stebalien: Because it "works" in some shells.

PR #400, opened 2020-01-23
Introducing compositions

### LINKED ISSUE #365: Test compositions

raulk:
Fixes #365, #361, #362, #363, #364.

---

Kinda feel bad for not having unit tests for all of this -- but life is what it is. Writing unit tests for this stuff is not easy, and deserves a larger initiative than this PR.

---

This PR introduces *Compositions*. Compositions are declarative files that describe what to build and run. No more cumbersome CLI flags -- kiss them goodbye. 

Compositions are TOML files, see `compositions/` for an example file (I can add more).

The major innovation is the concept of *Groups*. A *Group* is a cohort of instances that are built against a concrete set of upstream dependencies, and can specify their own test parameters.

With this feature, it's possible to launch a test case with, e.g. 200 instances, out of which:

* 100 instances are DHT clients, built against commit X, Y, Z of go-libp2p-kad-dht, go-libp2p, and go-libp2p-autonat.
* 40% of instances are DHT servers, built against the latest releases of all upstreams.
* the remaining 10% are bootstrappers, built against another set of dependencies.

Compositions are now accepted by the `build` and `run` commands, both of which have now been splintered into two variants (subcommands):

* `testground <build/run> single`: accepts previous CLI switches, runs a single group, creates a synthetic singleton composition, and sends it to the daemon.
* `testground <build/run> composition -f composition.toml`: accepts a composition file, along with two optional flags:
  * `--write-artifacts`: if set, it updates the composition file with the generated build artifacts. This is useful to bypass the build step when we intend to run the same composition multiple times (n iterations).
  * `--ignore-artifacts`: if set, it ignores all build artifacts specified in the composition file, and triggers a build for each group.
  * both flags can be combined to "refresh" the composition file, if need be.

The daemon RPCs have been simplified to only deal with Compositions.

## Other minor changes introduced by this PR

* Rename Server to Daemon, for consistency.
* Hoist the `client` package to `pkg/client`.
* Simplify the gorilla mux handlers, using middleware.
* Removal of `util` package (unidiomatic in Go), moving its functions to `conv` and `docker` (IIRC).
* Creation of `Engine` and `TestCensus` interfaces; useful in the near future for lifecycles.
* Renamed some conversion functions for clarity (now in the `conv` package).
* Removal of singleton package-level Engine instance. We now create it in the daemon init logic, and inject it to all gorilla/mux handlers.
* Removed the `cluster_nomad.go` placeholder runner.

## TODO

- [ ] A lot more testing.
- [ ] Update README and docs.

## Try it out

```sh
$ ./testground run composition -f compositions/find-peers.example.toml
```

Play with the composition file! Change the runner, builder, groups, number of instances, test parameters.


dirkmc:
In general it looks good to me.

It might be simpler not have a global instance count, instead specifying an exact number of instances in each group (eg 10, 45, 45)

raulk:
@dirkmc My rationale for going this way is that I can see folks wanting to gradually increase the instance count in their test case. Start with 100, if all goes well, dial up to 200, 300, 400, etc. With this approach, you'd only need to change the value in a single place.

dirkmc:
Makes sense 👍

compositions/find-peers.example.toml
```diff
@@ -0,0 +1,46 @@
...
+
+[[groups]]
+id = "bootstrappers"
+instances = { count = 10 }
+
+  [groups.build]
+  dependencies = [
+      { module = "github.com/libp2p/go-libp2p-kad-dht", version = "995fee9e5345fdd7c151a5fe871252262db4e788"},
+      { module = "github.com/libp2p/go-libp2p", version = "76944c4fc848530530f6be36fb22b70431ca506c"},
+  ]
```
Stebalien: I'm not sure if overriding dependencies is enough in the general as interfaces might have changed (also, we might want to test against javascript).

However, for our current use-case, I think it'll be fine.

---

Alternatives:

* Use a different module? That is, separate test plans from test instance runners.
* Allow build tags. That's the simplest for now.

raulk: I dig the build tags idea; that should allow us to shim API changes elegantly. There are some edge cases where folks would need to be mindful about making a commit introducing the build tags _before_ making any changes to the actual codebase, so that the old behaviour can be shimmed and compared against the new behaviour, but that's a fair tradeoff.

pkg/engine/engine.go
```diff
@@ -161,53 +174,98 @@ func (e *Engine) DoBuild(ctx context.Context, testplan string, builder string, i
...
+	)
+
+	// obtain an explicitly cancellable context so we can stop build jobs if
+	// something fails.
+	ctx, cancel = context.WithCancel(ctx)
+	defer cancel()
+
+	// Trigger a build for each group, and wait until all of them are done.
+	for i, grp := range comp.Groups {
+		ii, grpp := i, grp // captures
```
nonsense: No need to introduce new names, it is a common pattern to just do `i := i` when you need to capture the var.

pkg/runner/cluster_k8s.go
```diff
@@ -10,13 +10,17 @@ import (
 	"os"
```
nonsense: FYI this whole file is revamped at https://github.com/ipfs/testground/pull/433 - I think the fact that we got rid of `job` will also make it easier to start various compositions/images, and not have to manage m jobs of n pods.

nonsense:
Overall LGTM, given the time constraint we are under, I suggest we merge and integrate with the rest of the PR, so that we have a bit of time to review outstanding issues, such as the hanging of the sidecar, which is hard to do with many open PRs.

pkg/engine/engine.go
raulk: Done.

PR #402, opened 2020-01-23
fix k8s_instance.go runner to remove routes

nonsense:
I had royally fucked up the `routes` adding and removal, probably it was working by coincidence, due to the small cluster that these were initially developed on.

I am not happy with the current state of `manageContainer`, but at least this is working now:

Routes on a `testplan` container start as:
```
default via 100.96.5.1 dev eth0
100.96.0.0/16 via 100.96.5.1 dev eth0
100.96.5.0/24 dev eth0 proto kernel scope link src 100.96.5.253
```

and become the following after we attach the `data` network:
```
default via 20.0.0.0 dev eth1
20.0.0.0 dev eth1 scope link
24.171.0.0/16 dev eth1 proto kernel scope link src 24.171.64.0
100.64.0.10 via 100.96.5.1 dev eth0
100.96.4.3 via 100.96.5.1 dev eth0
```

You can see that `100.96.0.0/16` is restricted to just redis at `100.96.4.3`, and we also have route to the DNS at `100.64.0.10`.

This is working fine from a node that is different to the redis node (100.96.4.1 vs 100.96.5.1) as well as on the node which hosts both redis and containers:

```
default via 100.96.4.1 dev eth0
100.96.0.0/16 via 100.96.4.1 dev eth0
100.96.4.0/24 dev eth0 proto kernel scope link src 100.96.4.198
```

becomes
```
default via 31.0.0.0 dev eth1
24.171.0.0/16 dev eth1 proto kernel scope link src 24.171.2.2
31.0.0.0 dev eth1 scope link
100.64.0.10 via 100.96.4.1 dev eth0
100.96.4.3 via 100.96.4.1 dev eth0
```

---

TODO in follow-up PRs:
- [ ] Ortogonal to this, we might want to add some affinity restrictions, so that `redis` is hosted on a node all by itself, and not sharing the node with 100 other testplan containers. (or we could just bump up redis' resource requirements essentially making sure that not much else fits on its machine).
- [ ] Add test so that we are sure that testplans are not reachable on the control network (100.96.0.0 currently), issue is already added by @Stebalien at https://github.com/ipfs/testground/issues/348

### DIFF
--- a/pkg/sidecar/k8s_instance.go
+++ b/pkg/sidecar/k8s_instance.go
@@ -142,49 +142,63 @@ func (d *K8sInstanceManager) manageContainer(ctx context.Context, container *doc
 	}
 
 	// Get the routes to redis. We need to keep these.
-	redisRoutes, err := netlinkHandle.RouteGet(d.redis)
+	redisRoute, err := getRedisRoute(netlinkHandle, d.redis)
 	if err != nil {
-		return nil, fmt.Errorf("failed to resolve route to redis: %w", err)
+		return nil, fmt.Errorf("cant get route to redis: %s", err)
 	}
-
-	if len(redisRoutes) != 1 {
-		return nil, fmt.Errorf("expected to get only one route to redis, but got %v", len(redisRoutes))
-	}
-
-	redisRoute := redisRoutes[0]
+	logging.S().Debugw("got redis route", "route.Src", redisRoute.Src, "route.Dst", redisRoute.Dst, "gw", redisRoute.Gw.String(), "container", container.ID)
 
 	controlLinkRoutes, err := netlinkHandle.RouteList(controlLink, netlink.FAMILY_ALL)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/sidecar/k8s_instance.go
```diff
@@ -142,49 +142,63 @@ func (d *K8sInstanceManager) manageContainer(ctx context.Context, container *doc
...
-	redisRoute := redisRoutes[0]
+	logging.S().Debugw("got redis route", "route.Src", redisRoute.Src, "route.Dst", redisRoute.Dst, "gw", redisRoute.Gw.String(), "container", container.ID)
 
 	controlLinkRoutes, err := netlinkHandle.RouteList(controlLink, netlink.FAMILY_ALL)
 	if err != nil {
 		return nil, fmt.Errorf("failed to list routes for control link %s", controlLink.Attrs().Name)
 	}
 
-	gw := redisRoute.Gw
+	ip := redisRoute.Dst.IP
```
Stebalien: Can we rename this to `redisIp`?

Issue #399, opened 2020-01-23
fix global logger in testplans


nonsense:
For some reason logs from the global logger are not visible when running `testplan --vv`, see https://github.com/ipfs/testground/pull/393/files

This is probably some misconfiguration, but it'd be good to be able to enable trace logs on testplans so that we can debug them more easily until they become stable, and until testground itself is stable.

Stebalien:
Is this when writing to a file (`log_file`) or otherwise?

nonsense:
Logs emitted with `logging.S().Debug` are not send to `stdout` for some reason - hence me replacing with `runenv.Message` at https://github.com/ipfs/testground/pull/393/files

I haven't tried redirecting to a file.

Stebalien:
Hm. It looks like logging is a bit broken.

Stebalien:
I've found two issues:

1. We use the test runner's config to set the level (`log_level`), not "-vv". Not sure what we should do here.
2. We don't set this on k8s.

nonsense:
This has been completely revamped by Raul, so closing it.

PR #407, opened 2020-01-23
fix: set log level in cluster k8s

Stebalien:
relates to #399


nonsense:
Thanks, I am sure I had that in one of the PR, but probably got misplaced :/

Issue #377, opened 2020-01-21
generating, storing, and collecting event/stats dumps


raulk:
This epic includes everything related to:

1. Runners consolidating the stdout output from test instances (local:exec, local:docker already handled in #342; cluster:k8s missing).
2. Allowing test instances to emit files, e.g. dumps from the DHT event log, into a known location such that they can be collected as an aggregate for mining.
3. With a single command, collecting all assets from a given distributed test run into a single archive: consolidated stdout + all files.

raulk:
This v0.1 epic is done, barring #372 which will be tackled in v0.2.

PR #409, opened 2020-01-24
persist assets and logs

Stebalien:
* Mount an assets directory in test containers for persisting assets.
* Use the sidecar to write stderr/stdout to the each test instance's asset directory.

Currently, we put all this in the "work directory" under "results". I.e., `~/.testground/results/$testplan/$runid/$container_hostname/`.

At the moment, this is configured for local docker only. On K8S, we should mount the same S3 bucket on each host then mount that S3 bucket inside the sidecar and each test instance.

Note: I've also removed the log_file option from the docker runner. That's no longer useful given that we store test results.

part of #377

### DIFF
--- a/pkg/dockermanager/dockermanager.go
+++ b/pkg/dockermanager/dockermanager.go
@@ -3,6 +3,7 @@ package dockermanager
 import (
 	"context"
 	"fmt"
+	"io"
 	"time"
 
 	"github.com/docker/docker/api/types"
@@ -47,6 +48,15 @@ type Container struct {
 // Inspect inspects this docker container.
 func (dc *Container) Inspect(ctx context.Context) (types.ContainerJSON, error) {
 	return dc.Manager.ContainerInspect(ctx, dc.ID)
+}
+
+// Logs returns the logs for this docker container.
+func (dc *Container) Logs(ctx context.Context) (io.ReadCloser, error) {
+	return dc.Manager.ContainerLogs(ctx, dc.ID, types.ContainerLogsOptions{
+		ShowStderr: true,
+		ShowStdout: true,
+		Follow:     true,
+	})
 }
 
 // IsOnline returns whether or not the container is online.
--- a/pkg/runner/eventmanager.go
+++ b/pkg/runner/eventmanager.go
@@ -1,7 +1,6 @@
 package runner
 
 import (
-	"bufio"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -10,6 +9,8 @@ import (
 	"time"
 
 	"github.com/ipfs/testground/sdk/runtime"
+
+	"go.uber.org/zap/zapcore"
 )
 
 type eventType int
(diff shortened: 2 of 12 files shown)


pkg/sidecar/sidecar.go
```diff
@@ -57,68 +61,99 @@ func Run(runnerName string) error {
...
-		}
+		g, ctx := errgroup.WithContext(ctx)
 
-		/* TODO: Initialize all networks to the "down" state.
-		for _, n := range instance.Network.ListActive() {
-			instance.Network.ConfigureNetwork(&sync.NetworkConfig{
-				Network: n,
-				Enable:  false,
+		// Network configuration loop.
+		g.Go(func() error {
```
Stebalien: Note: this code hasn't changed. It's just in a goroutine now.

pkg/sidecar/sidecar.go
```diff
@@ -57,68 +61,99 @@ func Run(runnerName string) error {
...
-			Enable:  true,
-		})
-		if err != nil {
-			return err
-		}
+		g, ctx := errgroup.WithContext(ctx)
 
-		/* TODO: Initialize all networks to the "down" state.
-		for _, n := range instance.Network.ListActive() {
-			instance.Network.ConfigureNetwork(&sync.NetworkConfig{
```
Stebalien: Well, I removed this commented out code, but that's it.

pkg/sidecar/sidecar.go
```diff
@@ -57,68 +61,99 @@ func Run(runnerName string) error {
...
+						)
+					}
 				}
 			}
-		}
+			return nil
+		})
 
-		return nil
+		if resultPath != "" {
```
Stebalien: This is the new code in this file.

pkg/runner/eventmanager.go
```diff
@@ -89,12 +90,50 @@ func (c *EventManager) Manage(id string, stdout, stderr io.ReadCloser) {
...
-		for scanner.Scan() {
-			c.msg(idx, id, time.Now(), Error, scanner.Text())
-		}
-		if err := scanner.Err(); err != nil {
-			c.msg(idx, id, time.Now(), Error, "stderr error: "+err.Error())
+
+		decoder := json.NewDecoder(stderr)
+		for {
+			event := make(map[string]json.RawMessage, 16)
+			if err := decoder.Decode(&event); err != nil {
```
Stebalien: Unfortunately, I can't find a _nice_ way to parse log events with arbitrary fields.

pkg/dockermanager/dockermanager.go
```diff
@@ -49,6 +50,15 @@ func (dc *Container) Inspect(ctx context.Context) (types.ContainerJSON, error) {
 	return dc.Manager.ContainerInspect(ctx, dc.ID)
 }
 
+// Logs returns the logs for this docker container.
+func (dc *Container) Logs(ctx context.Context) (io.ReadCloser, error) {
+	return dc.Manager.ContainerLogs(ctx, dc.ID, types.ContainerLogsOptions{
```
raulk: Why not replace `os.Stdout` and `os.Stderr` with teed I/O, such that it would duplicate the output to both the real stdout and stderr handles + `stdout.json` and `stderr.json` files, under the designated assets directory? The problem with this approach is that it won't work for the `local:exec` runner, no?

pkg/runner/local_docker.go
```diff
@@ -400,6 +435,10 @@ func ensureSidecarContainer(ctx context.Context, cli *client.Client, log *zap.Su
 				Type:   mount.TypeBind,
 				Source: "/var/run/docker.sock", // TODO: don't hardcode this.
 				Target: "/var/run/docker.sock",
+			}, {
```
nonsense: I missed that during review - we should be consistent with the way we start the `sidecar` and update it in every runner when changing - currently we have Docker Swarm playbooks, and Kubernetes yaml specs in addition to this. Any other places we start the sidecar?

cc @raulk @Stebalien

PR #415, opened 2020-01-27
merge k8s setup steps together

nonsense:
TODO:
- [x] add `yaml` config for the cluster, so that we can add extra configs to the kubelet:

```
  kubelet:
    anonymousAuth: false
    maxPods: 200
    allowedUnsafeSysctls:
    - net.core.somaxconn
```

### DIFF
--- a/infra/k8s/README.md
+++ b/infra/k8s/README.md
@@ -59,36 +59,25 @@ kops create cluster \
   --master-size c5.2xlarge \
   --node-size c5.2xlarge \
   --node-count 8 \
-  --networking cni \
+  --networking flannel \
   --name $NAME \
   --yes
 ```
 
-5. Wait for `master` node to appear in `kubectl get nodes` - it will be in `NotReady` state as we haven't installed any Networking CNI yet.
-```
-watch 'kubectl get nodes -o wide'
-```
-
-6. Install Flannel
-
-```
-kubectl apply -f ./infra/k8s/kops-weave/flannel.yml
-```
-
-7. Wait for all nodes to appear in `kubectl get nodes` with `Ready` state, and for all pods in `kube-system` namespace to be `Running`.
+5. Wait for all nodes to appear in `kubectl get nodes` with `Ready` state, and for all pods in `kube-system` namespace to be `Running`.
 ```
 watch 'kubectl get nodes -o wide'
 
 kubectl -n kube-system get pods -o wide
 ```
 
-8. Install CNI-Genie, Weave and Dummy daemonset - we need a container on every worker node so that interface `cni0` is created, and Weave's initContainer can add a route to the Services CIDR
+6. Install CNI-Genie, Weave and Dummy daemonset - we need a container on every worker node so that interface `cni0` is created, and Weave's initContainer can add a route to the Services CIDR
 
 ```
 kubectl apply -f ./infra/k8s/kops-weave/genie-plugin.yaml -f ./infra/k8s/kops-weave/dummy.yml -f ./infra/k8s/kops-weave/weave.yml
 ```
 
-9. Destroy the cluster when you're done working on it
+7. Destroy the cluster when you're done working on it
 
 ```
 kops delete cluster $NAME --yes
--- a/infra/k8s/README.md
+++ b/infra/k8s/README.md
@@ -50,7 +50,7 @@ export KOPS_STATE_STORE=s3://kops-backend-bucket
 export ZONES=eu-central-1a
 ```
 
-4. Create The cluster
+4. Generate the cluster spec
 
 ```
 kops create cluster \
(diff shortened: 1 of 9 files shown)


infra/k8s/README.md
```diff
@@ -56,9 +56,9 @@ export ZONES=eu-central-1a
 kops create cluster \
   --zones $ZONES \
   --master-zones $ZONES \
-  --master-size m4.xlarge \
-  --node-size m4.xlarge \
-  --node-count 2 \
+  --master-size c5.2xlarge \
```
nonsense: Looks like our setup is CPU intensive, so it makes more sense to run `c5` instances, rather than `m4` or `m5` for now.

infra/k8s/kops-weave/weave.yml
```diff
@@ -172,6 +172,7 @@ items:
             - sh
             - -ac
             - >
+              sleep 20 &&
```
nonsense: `weave` daemonset is currently dependent on the `dummy` daemonset, which is not great - we should not have such dependencies between daemonsets in general, but I don't know how to avoid that one, because the `flannel` interface is lazily created, only when there is a pod on the host.

So for now we add a `sleep 20`, which will delay the `initContainer` on `weave`, hopefully enough so that `dummy` has started - when we start both daemonsets at the same time.

I will add a command as part of the setup so that we can easily confirm that routes on every host are setup correctly during Kubernetes startup.

manifests/dht.toml
```diff
@@ -42,7 +42,7 @@ instances = { min = 16, max = 10000, default = 16 }
   n_bootstrap   = { type = "int", desc = "number of bootstrap nodes", unit = "int", default = 1 }
   bucket_size   = { type = "int", desc = "routing table bucket size", unit = "peers", default = 2 }
   n_find_peers  = { type = "int", desc = "number of peers to find", unit = "peers", default = 1 }
-  timeout_secs  = { type = "int", desc = "test timeout", unit = "seconds", default = 600 }
+  timeout_secs  = { type = "int", desc = "test timeout", unit = "seconds", default = 300 }
```
nonsense: 10min. timeout seems to be too long. 5min is fine for 1000 instances. I suggest we get 1000 setup working correctly and then we can see how to adjust this based on underlying infrastructure.

infra/k8s/tools/README.md
```diff
@@ -0,0 +1,9 @@
+# Tools and commands that are helpful during running Testground on Kubernetes
+
+1. `confirm_weave.sh`
```
nonsense: @raulk we might want to move this to a command, such as `testground validate-cluster` or something like this, which does a set of checks that the cluster is setup correctly:
- confirms that `sidecar` is running on every host.
- confirms that `redis` is running
- confirms that routes on hosts are setup correctly

`kops` has a similar tool:

```
➜  testground git:(less-steps) kops validate cluster simple.k8s.local
Validating cluster simple.k8s.local

INSTANCE GROUPS
NAME                    ROLE    MACHINETYPE     MIN     MAX     SUBNETS
master-eu-central-1a    Master  c5.2xlarge      1       1       eu-central-1a
nodes                   Node    c5.2xlarge      14      14      eu-central-1a

NODE STATUS
NAME                                            ROLE    READY
ip-172-20-34-37.eu-central-1.compute.internal   node    True
ip-172-20-39-154.eu-central-1.compute.internal  master  True
ip-172-20-39-244.eu-central-1.compute.internal  node    True
ip-172-20-45-108.eu-central-1.compute.internal  node    True
ip-172-20-48-175.eu-central-1.compute.internal  node    True
ip-172-20-52-192.eu-central-1.compute.internal  node    True
ip-172-20-53-4.eu-central-1.compute.internal    node    True
ip-172-20-58-197.eu-central-1.compute.internal  node    True
ip-172-20-58-89.eu-central-1.compute.internal   node    True
ip-172-20-59-73.eu-central-1.compute.internal   node    True
ip-172-20-60-196.eu-central-1.compute.internal  node    True
ip-172-20-61-105.eu-central-1.compute.internal  node    True
ip-172-20-61-144.eu-central-1.compute.internal  node    True
ip-172-20-61-26.eu-central-1.compute.internal   node    True
ip-172-20-63-3.eu-central-1.compute.internal    node    True

Your cluster simple.k8s.local is ready
```

infra/k8s/kops-weave/weave.yml
nonsense: Actually it is not `weave` that is dependent on the DNS route, but rather `sidecar`. So I moved the `initContainer` to `sidecar` and this simplifies things.

infra/k8s/sidecar.yaml
```diff
@@ -19,6 +19,21 @@ spec:
...
+        command:
+        - sh
+        - -ac
+        - >
+          sleep 20 &&
+          export GW=$(ip route | grep cni0 | awk '{print $7}') &&
+          echo $GW &&
+          ip route &&
+          ip route add 100.64.0.0/16 via $GW &&
+          ip route || true
```
nonsense: It'd be nice for this command to detect if the `route` is already there, so that we can restart the `sidecar` daemonset.

pkg/runner/cluster_k8s.go
```diff
@@ -243,7 +243,7 @@ func (*ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow io.Wri
 		if job.Status.Active == 0 && (job.Status.Succeeded > 0 || job.Status.Failed > 0) {
 			break
 		}
-		if time.Since(start) > 5*time.Minute {
+		if time.Since(start) > 10*time.Minute {
```
Stebalien: Ideally, we wouldn't have hard-coded timeouts like this. But we can tackle that later.

infra/k8s/sidecar.yaml
gmasgras: Might not need to use the dummy container or sleep here, see https://github.com/ipfs/testground/issues/416#issuecomment-579060016

nonsense: I will try to use `--networking flannel` and see if the interface is created. If `cni0` is indeed created, then we don't need a dummy container or sleep, otherwise we still need them until we come up with another solution.

nonsense: @gmasgras when starting `kops` with `--networking flannel`, we still need to create a pod on every host, so that we get the `cni0` interface and then use it to add the DNS route (100.64.0.0/16).

We should get rid of the sleep, and adding pooling in this cmd, so that it is more consistent, and not time-based.

PR #418, opened 2020-01-27
sdk examples

coryschwartz:
I wrote this primarily as a way for studying the SDK, and might be useful to future plan-writers or tests like the placebo test is already used.

Each file represents some small feature of the plan SDK.

### DIFF
--- a/plans/example/failure.go
+++ b/plans/example/failure.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"errors"
+
 	"github.com/ipfs/testground/sdk/runtime"
 )
 
--- a/plans/example/failure.go
+++ b/plans/example/failure.go
@@ -1,0 +1,12 @@
+package main
+
+import (
+	"errors"
+	"github.com/ipfs/testground/sdk/runtime"
+)
+
+// ExampleFailure always fails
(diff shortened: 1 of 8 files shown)


plans/example/failure.go
```diff
@@ -0,0 +1,12 @@
+package main
+
+import (
+	"errors"
+	"github.com/ipfs/testground/sdk/runtime"
```
Stebalien: nit: there should be a newline between these.

plans/example/go.mod
```diff
@@ -0,0 +1,10 @@
+module github.com/ipfs/testround/plans/example
+
+go 1.13
+
+require (
```
Stebalien: We add replace directives for the SDK to use the versions in this repo:

```
replace github.com/ipfs/testground/sdk/runtime => ../../sdk/runtime
replace github.com/ipfs/testground/sdk/sync => ../../sdk/sync
```

plans/example/main.go
```diff
@@ -0,0 +1,31 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"github.com/ipfs/testground/sdk/runtime"
```
Stebalien: nit: newline

PR #430, opened 2020-01-29
chore: do not fail on negative cov difference

### LINKED ISSUE #321: Shall we disable codecov from being run in CI on every commit?
I think this is a discussion that we should have as a team:

**Do we want codecov to be run as part of CI and to mark the state of Testground as broken when it fails?**

Screenshot 2019-12-23 at 16 28 49

I personally find the current CI setup problematic, because:
1. It looks like the project has a broken master branch, which is not a good development practice. (see screenshot)
2. PRs appear as they are failing CI, and a reviewer has to check individually all PRs that they are marked to review, to determine if an `optional` (codecov) check is failing, or a `required` check is failing.

I think that code coverage is useful in finding untested parts of our codebase, but I don't find the actual numeric measurement very valuable. Therefore I find that tying our CI success/failure with the numeric measurement is actually counterproductive.

Another point worth mentioning is that it seems we are still in the MVP phase of the project - we have a high level overview of what we want to build and we have an idea of where we want Testground to go in the future, but we haven't locked the tools we will need to achieve that. Do we need `cluster_swarm.go` to be covered with tests, if we might drop Docker Swarm support in a month? Do we need `cluster_k8s.go` to be covered with tests, if we might drop Kubernetes support in favour of Nomad runner in another month? I think it should be up to code reviewers to determine the answers to these questions during code reviews, rather than have an automated system telling us that we are a few % short of the necessary code coverage.

Interested in your opinions @daviddias @hacdias @jimpick @raulk @Stebalien

hacdias:
Apparently, the informational flag on Codecov is not working as we expect it to, i.e., not to fail when there are negative changes in the coverage.

My idea is to just change the threshold to 100%. This param is:

> Allow the coverage to drop by X%, and posting a success status.

Closes #321.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -9,7 +9,7 @@ coverage:
         # Commits pushed to master should not make the overall
         # project coverage decrease by more than 5%:
         target: auto
-        threshold: 5%
+        threshold: 100%
     patch:
       default:
         # https://docs.codecov.io/docs/commit-status#section-informational
@@ -22,4 +22,4 @@ coverage:
         # extension:
         # https://github.com/codecov/browser-extension
         target: auto
-        threshold: 5%
+        threshold: 100%


nonsense:
@hacdias thanks, this looks like it will do the trick - we keep the informational metric on the README.md, but we never fail PRs.

Issue #374, opened 2020-01-21
all runners: inject TEST_OUTPUT_FILE_PATH env var with location to write files


raulk:
This value should be set by the runner. Why? Each runner will have a different strategy. k8s will want to ship logs to a networked volume, e.g. S3, whereas the local runner will be fine storing assets in the local filesystem.

_Note: looks like #409 implemented some of this._

What is missing is the following:
- [x] local:exec runner.
- [x] cluster:k8s runner (sync with @nonsense).
- [ ] rename `TestArtifacts` to `TestAssetsDir`.

hacdias:
Just PRed for the first and last points (#431).

@nonsense how'd you this for kubernetes? I presume there should be some kind of mounting on the `jobRequest`?

nonsense:
@hacdias I opened a draft PR at https://github.com/ipfs/testground/pull/433 , which works on small number of machines, but needs to be load tested and ran with at least 1k containers on 15-20 VMs.

raulk:
This was closed in https://github.com/ipfs/testground/pull/467.

PR #431, opened 2020-01-29
feat: add test output file path

hacdias:
This finishes two TODOs from #374:

- local:exec runner.
- rename `TestArtifacts` to `TestAssetsDir`.

License: MIT

### DIFF
--- a/pkg/runner/local_docker.go
+++ b/pkg/runner/local_docker.go
@@ -147,8 +147,8 @@ func (*LocalDockerRunner) Run(ctx context.Context, input *api.RunInput, ow io.Wr
 	}
 
 	// Ensure the work dir.
-	workDir := filepath.Join(input.EnvConfig.WorkDir(), "results")
-	if err := os.MkdirAll(workDir, 0777); err != nil {
+	workDir, err := getWorkDir(input)
+	if err != nil {
 		return nil, err
 	}
 
@@ -163,9 +163,8 @@ func (*LocalDockerRunner) Run(ctx context.Context, input *api.RunInput, ow io.Wr
 		return nil, err
 	}
 
-	// Create the run output directory and write the runenv.
-	runDir := filepath.Join(workDir, input.TestPlan.Name, input.RunID)
-	if err := os.MkdirAll(runDir, 0777); err != nil {
+	runDir, err := getRunDir(input)
+	if err != nil {
 		return nil, err
 	}
 	if f, err := os.Create(filepath.Join(runDir, "env.json")); err == nil {
--- a/pkg/runner/local_docker.go
+++ b/pkg/runner/local_docker.go
@@ -208,8 +208,8 @@ func (*LocalDockerRunner) Run(ctx context.Context, input *api.RunInput, ow io.Wr
 				"testground.runid":    input.RunID,
 			},
 		}
-		assetsDir := filepath.Join(runDir, name)
-		err = os.Mkdir(assetsDir, 0777)
+		hostAssetsDir := filepath.Join(runDir, name)
+		err = os.Mkdir(hostAssetsDir, 0777)
 		if err != nil {
 			break
 		}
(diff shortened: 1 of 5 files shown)


hacdias:
@raulk @Stebalien will make the k8s on a different PR.

hacdias:
Actually, I noticed there's something *wrong* with this PR.

hacdias:
Fixed now!

nonsense:
Related to https://github.com/ipfs/testground/pull/433 - I will do the rename there when this gets merged.

pkg/runner/local_docker.go
```diff
@@ -107,7 +107,7 @@ func (*LocalDockerRunner) Run(ctx context.Context, input *api.RunInput, ow io.Wr
 		TestInstanceCount:  input.Instances,
 		TestInstanceParams: input.Parameters,
 		TestSidecar:        true,
-		TestArtifacts:      "/artifacts",
+		TestAssetsDir:      "/artifacts",
```
nonsense: Maybe change the mounted dir as well, so that we are consistent?

pkg/runner/local_docker.go
```diff
@@ -209,17 +209,17 @@ func (*LocalDockerRunner) Run(ctx context.Context, input *api.RunInput, ow io.Wr
 				"testground.runid":    input.RunID,
 			},
 		}
-		artifactDir := filepath.Join(runDir, name)
-		err = os.Mkdir(artifactDir, 0777)
+		assetsDir := filepath.Join(runDir, name)
```
nonsense: nitpick: Maybe rename to `hostAssetsDir`, so that it is clear that this is the directory on the host, and not inside the container.

pkg/runner/local_exec.go
```diff
@@ -81,6 +83,12 @@ func (*LocalExecutableRunner) Run(ctx context.Context, input *api.RunInput, ow i
 
 	testcase := plan.TestCases[seq]
 
+	// Create the assets directory
+	assetsDir := filepath.Join(input.EnvConfig.WorkDir(), input.TestPlan.Name, input.RunID, "assets")
```
nonsense: Same comment as above - `assetsDir` -> `hostAssetsDir` in case you agree that it is nice to be explicit that this is the host directory, just so that we are consistent with the other runner.

sdk/runtime/files.go
```diff
@@ -64,7 +64,7 @@ func (re *RunEnv) CreateRandomDirectory(directoryPath string, depth uint) (strin
 //
 // Test artifacts will be saved when the test terminates and available for
```
nonsense: Maybe change the godoc as well to be consistent.

pkg/runner/local_docker.go
Stebalien: note: "assets" are usually inputs, not outputs. The outputs are usually called "artifacts".

pkg/runner/local_exec.go
Stebalien: We should also have one per test instance.

pkg/runner/local_docker.go
hacdias: I agree artifacts would be better because it's the **output** and not the **input**. We should come up with a decision (see #374). What do you think @raulk?

pkg/runner/local_exec.go
hacdias: @Stebalien ugh. I did that and forgot to push because I put it on a different branch... pushed!

hacdias:
@Stebalien @nonsense I have added one directory per instance on the `local:exec`. I kept the 'assets' name.

@raulk said:

>  This is ambiguous. An artifact in testground is a build artifact, that is, a Docker image or an executable. Unless we want to rename those, let’s call these “assets”.

So I just kept assets here.

pkg/runner/local_docker.go
raulk: I don’t think there’s an industry norm here. We already use the name “artifact” for build outputs: Docker images and executables, and I think that name is pretty straightforward and clear. Naming these “artifacts” too would be ambiguous. That’s why I proposed “assets”, which can be input assets (e.g. fixtures and configs), or output assets (e.g. log files, traces, dumps, snapshots of network topology, etc.)

raulk: Basically I don’t want things being consumed or produced from test runs to have two names (artifacts and assets, like proposed). I’d rather use the term _input/output asset_ to refer to a resource handled by a test case.

raulk:
Superseded by #467.

PR #433, opened 2020-01-29
s3fs daemon ; n k8s pods instead of k8s job controller

nonsense:
This PR is adding:
1. s3fs daemonset so that we mount an S3 bucket on every host in a k8s cluster
2. volume mounts to each pod, so that each pod has access to a directory inside the shared S3 bucket
3. changing k8s `job` to `n` `pods`, because it seems that job doesn't fit our use case very well
- a job would restart a pod if it fails, because it cares about n `successful` runs - not exactly what we need with testground
- we don't have a unique identifier per pod, so it is hard to mount an `artifacts` subdirectory for every testplan and have isolation
4. terraform plans so that we have infrastructure as code for the shared s3 bucket, as well as Dockerfile for the s3fs daemon - heavily influenced from https://devops-bordeaux.net/blog/mounting-s3-bucket-in-kubernetes-pod/
5. support in cluster_k8s for artifacts

---

OPEN QUESTIONS FOR FUTURE PRS:
- [ ] 1. What retention rate do we want on the shared S3 bucket that will keep artifacts/assets from all users of Testground?
- [ ] 2. What do we do with `stdout/stderr` emitted from containers running on Kubernetes? Currently this PR is tailing the last 2 lines (as we might have 1k or 10k containers) and returning it back to the `testground daemon`. I don't see a reason to forward these to the `sidecar`. This should be intuitive and ideally similar/same on every runner. https://github.com/ipfs/testground/issues/439
- [ ] 3. What do we do with `normalised results` emitted from containers running on Kubernetes? Currently this PR is tailing the last 2 lines (as we might have 1k or 10k containers) and returning it back to the `testground daemon`. I don't see a reason to forward these to the `sidecar`. This should be intuitive and ideally similar/same on every runner. https://github.com/ipfs/testground/issues/439

---

TODO:
- [x] load test `n` `pods` vs `job`
- [x] load test s3fs when having 1000 containers writing to the same s3 bucket
- [x] cleanup testplan
- [x] automate adding of AWS credentials for the `assets-s3-bucket` - we could store these on S3 or on 1Password, but users shouldn't have to manually edit the `yaml` file to add them.
- [x] ~which vault shall we load the AWS credentials for `assets-s3-bucket`, so that everyone who wants to use Testground can? for example @coryschwartz doesn't have access to `Test Infra`.~
- [ ] add `nonsens3/s3fs` docker image as part of `ipfs`
---

NICE TO HAVE:
- [x] proper pool of connections, not roundrobin of `n` clients.


nonsense:
This is becoming rather big, so opening for review already. Please check the TODO list so we have no duplicates.

infra/k8s/sidecar.yaml
```diff
@@ -36,7 +36,8 @@ spec:
           ip route || true
       containers:
       - name: testground-sidecar
-        image: ipfs/testground:edge
+        #image: ipfs/testground:edge
```
nonsense: I'm afraid we have introduced a regression in the `sidecar` - a commit from a few days ago is much more stable for me, otherwise it seems like `sidecar` is either losing connection to `docker.sock` or deadlocking (we have an issue for that). Not sure what is going on, and I haven't had time to investigate further just yet.

pkg/runner/cluster_k8s.go
```diff
@@ -146,67 +148,156 @@ func (*ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow io.Wri
...
 	// Define k8s client configuration
 	config := defaultKubernetesConfig()
-	k8scfg, err := clientcmd.BuildConfigFromFlags("", config.KubeConfigPath)
-	if err != nil {
-		return nil, fmt.Errorf("could not start k8s client from config: %v", err)
-	}
 
-	// Create the clientset
-	clientset, err := kubernetes.NewForConfig(k8scfg)
+	workers := 20
```
nonsense: Scheduling all pods to Kubernetes and then pooling for state is faster when we use a few clients concurrently.

infra/k8s/kops-weave/s3bucket.yml
```diff
@@ -0,0 +1,46 @@
+apiVersion: apps/v1
+kind: DaemonSet
```
nonsense: Actually this can replace the `dummy` daemonset, so I will probably unite them in a future PR.

pkg/runner/client_pool.go
```diff
@@ -0,0 +1,56 @@
...
+			return nil, fmt.Errorf("could not create k8s clientset: %v", err)
+		}
+
+		pool.availableC <- k8sClientset
+	}
+
+	return pool, nil
+}
+
+func (p *pool) Get(ctx context.Context) (*kubernetes.Clientset, error) {
```
nonsense: better names for these might be `Acquire` and `Release`.

infra/k8s/kops-weave/s3bucket.yml
nonsense: Already removed `dummy` daemonset from the README, `s3bucket` now does the function to create an `cni0` interface to be detected by the `sidecar`.

pkg/runner/client_pool.go
hacdias: I'm a bit confused about how 'Acquire' and 'Release' might be equivalent. If I understood correctly, this a connection pool and this get is used to acquire a connection from the pool. If so, I'd rename it acquire as it is a bit more explicit.

nonsense: OK, will rename.

The idea is that we have `n` connections, and just pick one that is `free` whenever we need a client, as this is a slow HTTP client and blocks for quite a while when we make a request to the Kubernetes API.

nonsense:
@hacdias you should have all required access to try this now :) thanks for review! :)

go.sum
```diff
@@ -24,6 +24,8 @@ github.com/aead/siphash v1.0.1/go.mod h1:Nywa3cDsYNNK3gaciGTWPwHt0wlpNV15vwmswBA
 github.com/aws/aws-sdk-go v1.15.78/go.mod h1:E3/ieXAlvM0XWO57iftYVDLLvQ824smPP3ATZkfNZeM=
 github.com/aws/aws-sdk-go v1.25.32 h1:GhqlDvuPXnlW46VoKvfLZkJj5IA6jGLO+/TUPCJSYOY=
```
raulk: nit: go mod tidy

infra/k8s/kops-weave/s3bucket.yml
```diff
@@ -0,0 +1,55 @@
...
+  template:
+    metadata:
+      annotations:
+        cni: "flannel"
+      labels:
+        name: s3bucket
+    spec:
+      containers:
+        - name: s3fs
+          image: nonsens3/s3fs
```
raulk: Is this a fork you maintain, @nonsense?

infra/k8s/sidecar.yaml
raulk: Can we instead depend on [the past tag of the `ipfs/testground` image](https://hub.docker.com/r/ipfs/testground/builds) that was well-behaved?

pkg/runner/client_pool.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"k8s.io/client-go/kubernetes"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+type pool struct {
+	sync.Mutex
+
+	availableC chan *kubernetes.Clientset
```
raulk: Not draining this channel on close is fine for now, because we have no concept of graceful shutdown across components, but at one point we'll want to do that.

sdk/runtime/runenv.go
```diff
@@ -30,6 +30,7 @@ const (
 	EnvTestInstanceCount  = "TEST_INSTANCE_COUNT"
 	EnvTestInstanceRole   = "TEST_INSTANCE_ROLE"
 	EnvTestInstanceParams = "TEST_INSTANCE_PARAMS"
+	EnvTestArtifacts      = "TEST_ARTIFACTS"
```
raulk: Let's rename in accordance with #439 ;-)

pkg/runner/cluster_k8s.go
raulk: Great discovery!

pkg/runner/cluster_k8s.go
```diff
@@ -273,4 +243,146 @@ func (*ClusterK8sRunner) CompatibleBuilders() []string {
...
+func monitorTestplanRunState(ctx context.Context, pool *pool, log *zap.SugaredLogger, runID string, k8sNamespace string, replicas int) error {
+	client, err := pool.Acquire(ctx)
+	if err != nil {
+		return err
+	}
+	defer pool.Release(client)
+
+	start := time.Now()
+	for {
+		if time.Since(start) > 10*time.Minute {
```
raulk: I think we'll want to extract these timing assumptions to runner config variables.

infra/k8s/kops-weave/s3bucket.yml
nonsense: There is no official version for this one, I have committed here the Dockerfile, but yeah - we should move that to the `ipfs` docker repo.

sdk/runtime/runenv.go
nonsense: I will leave that for a future PR, because there are multiple occurrences of the word artifacts, and it appears there is no consensus on the name.

go.sum
nonsense: done

infra/k8s/kops-weave/s3bucket.yml
nonsense: Added a TODO, will move to an issue and we can fix later.

PR #437, opened 2020-01-30
Only build Docker logs on linux

aschmahmann:
`dockerLogs` is a private struct that is only used in the linux host, it's file should be linux build only as is the rest of the sidecar package. Since we have git commit hooks that run the linter this is required to make commits on Windows succeed.

@Stebalien should I enable the linter for the Windows CI to prevent this in the future or is the cost not worth it?


Stebalien:
@aschmahmann no reason not to enable it on windows.

PR #440, opened 2020-01-31
stdout and stderr logs from k8s

nonsense:
1. `structured logs / output` ends up at `~/.testground/results` (at the `WorkDir`).
2. `assets/artifacts` end up at `s3://assets-s3-bucket/` and can be merged with the `logs` with:

```
aws s3 cp s3://assets-s3-bucket/<RUN_ID> ~/.testground/results/<RUN_ID> --recursive
```

### DIFF
--- a/pkg/runner/cluster_k8s.go
+++ b/pkg/runner/cluster_k8s.go
@@ -1,8 +1,8 @@
 package runner
 
 import (
-	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -204,6 +204,33 @@ func (*ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow io.Wri
 
 	var gg errgroup.Group
 
+	// TODO: hack - this is the daemon, and not the user's local machine
+	workDir := filepath.Join(input.EnvConfig.WorkDir(), "results")
+	if err := os.MkdirAll(workDir, 0777); err != nil {
+		return nil, err
+	}
+
+	// TODO: hack - create the run output directory and write the runenv.
+	runDir := filepath.Join(workDir, input.RunID)
+	if err := os.MkdirAll(runDir, 0777); err != nil {
+		return nil, err
+	}
+	if f, err := os.Create(filepath.Join(runDir, "env.json")); err == nil {
+		encoder := json.NewEncoder(f)
+		encoder.SetIndent("", "  ")
+		encoder.SetEscapeHTML(false)
+		err1 := encoder.Encode(template)
+		err2 := f.Close()
+		if err1 != nil {
+			return nil, err1
+		}
+		if err2 != nil {
+			return nil, err2
+		}
+	} else {
+		return nil, err
+	}
+
 	for _, g := range input.Groups {
 
 		for i := 0; i < g.Instances; i++ {
@@ -221,9 +248,30 @@ func (*ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow io.Wri
 
 				podName := fmt.Sprintf("%s-%s-%s-%d", jobName, input.RunID, g.ID, i)
 
-				logs := getPodLogs(client, podName)
+				// Log monitor.
+				path := filepath.Join(
+					runDir,
+					podName,
+				)
+				err = os.MkdirAll(path, 0777)
+				if err != nil {
+					return err
+				}
 
-				fmt.Print(logs)
+				f, err := os.Create(filepath.Join(path, "stdoutanderr.json"))
+				if err != nil {
+					return err
+				}
+
+				err = getPodLogs(client, podName, f)
+				if err != nil {
+					return err
+				}
+
+				err = f.Close()
+				if err != nil {
+					return err
+				}
 				return nil
 			})
 		}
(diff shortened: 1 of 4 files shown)


pkg/runner/cluster_k8s.go
```diff
@@ -204,6 +204,33 @@ func (*ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow io.Wri
...
 	var gg errgroup.Group
 
+	// TODO: hack - this is the daemon, and not the user's local machine
+	workDir := filepath.Join(input.EnvConfig.WorkDir(), "results")
+	if err := os.MkdirAll(workDir, 0777); err != nil {
+		return nil, err
+	}
+
+	// TODO: hack - create the run output directory and write the runenv.
+	runDir := filepath.Join(workDir, input.RunID)
```
nonsense: @Stebalien note that this is different to the `local-docker` runner, which also adds the plan name. Not sure which one we prefer, but better be consistent.

pkg/runner/cluster_k8s.go
```diff
@@ -221,9 +248,30 @@ func (*ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow io.Wri
...
+					runDir,
+					podName,
+				)
+				err = os.MkdirAll(path, 0777)
+				if err != nil {
+					return err
+				}
 
-				fmt.Print(logs)
+				f, err := os.Create(filepath.Join(path, "stdoutanderr.json"))
```
nonsense: I am not aware of a way to split `stdout` and `stderr` on k8s - I think there is no such feature.

Stebalien: We could just let the sidecar do this.

Stebalien: But can we not talk to docker directly?

pkg/runner/cluster_k8s.go
```diff
@@ -204,6 +204,33 @@ func (*ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow io.Wri
...
+	if err := os.MkdirAll(workDir, 0777); err != nil {
+		return nil, err
+	}
+
+	// TODO: hack - create the run output directory and write the runenv.
+	runDir := filepath.Join(workDir, input.RunID)
+	if err := os.MkdirAll(runDir, 0777); err != nil {
+		return nil, err
+	}
+	if f, err := os.Create(filepath.Join(runDir, "env.json")); err == nil {
```
Stebalien: Let's make a helper function for this.

pkg/runner/cluster_k8s.go
```diff
@@ -204,6 +204,33 @@ func (*ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow io.Wri
 
 	var gg errgroup.Group
 
+	// TODO: hack - this is the daemon, and not the user's local machine
```
Stebalien: Put this in S3?

pkg/runner/cluster_k8s.go
Stebalien: If we're going to put the results of each test in s3, I wouldn't use the plan name. I only used the plan name to make browsing `~/.testground/results/` easier.

Ideally, `testground run ...` would spit out the result location for the test.

pkg/runner/cluster_k8s.go
```diff
@@ -221,9 +248,30 @@ func (*ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow io.Wri
...
 
-				fmt.Print(logs)
+				f, err := os.Create(filepath.Join(path, "stdoutanderr.json"))
+				if err != nil {
+					return err
+				}
+
+				err = getPodLogs(client, podName, f)
+				if err != nil {
+					return err
```
Stebalien: close file on error.

pkg/runner/cluster_k8s.go
nonsense: `sidecar` is on the remote cluster. This is on the `daemon` or on the `client` side, which currently is on the devs machine. I don't see any benefit in the logs going from individual containers to the sidecar container and then somehow again to the testground daemon and client.

What is the benefit of `sidecar` managing `logs` at all? This only introduces complexity.

nonsense: > But can we not talk to docker directly?

Not sure why we want to go one-level deeper and avoid using the Kubernetes APIs, when this is how we manage to workloads - we schedule `pods` with the Kubernetes API, we monitor their state with the Kubernetes API, why would we have to investigate what Kubernetes is doing under the hood and attach to individual containers for logs?

pkg/runner/cluster_k8s.go
nonsense: We could, however as far as I understood @raulk , we want the `structured output` from tests to be read and returned right away to users. Not sure what is the point of storing this on S3, and then adding more logic to process it on `testground` by reading from S3.

S3 makes sense for assets, which are not structured, and will potentially be much bigger.

pkg/runner/cluster_k8s.go
nonsense: Sure, this PR is just trying to flesh out the `flow` of `logs`, and when we agree on that, we can refactor this.

nonsense:
> We could just let the sidecar do this.

I don't understand why stdout and stderr from all testplan instances on a given host have to pass through the sidecar. What's the benefit of doing that, in comparison with just using the Kubernetes API directly with each instance and fetching them on the developer's machine?

nonsense:
I imagine this will conflict with @raulk 's changes, but just opening it up, as it is ready and working - we could further improve by propagating logs from `daemon` to `client`, but for now I assume these are running on the same machine, which is the case for v0.1

Feel free to close if the outputs/logs gets solved in a better way in any other PR.

pkg/runner/cluster_k8s.go
Stebalien: My point was that the sidecar talks directly to the docker daemons on each of these nodes and _can_ distinguish between STDOUT and STDERR. Really, we don't want to send this back to the daemon anyways. If we're just dropping stderr/stdout to a directory in S3, the sidecar code can (could?) already do this.

However, if we're going going to start writing logs to a file, we should be able to just collect stdout (should only include stats) with the daemon and sidestep this.

pkg/runner/cluster_k8s.go
Stebalien: There are different kinds of "structured output". We have stdout which is status messages, metrics etc. Then we have _logs_. Collecting all those logs and streaming them back to the user is likely going to overwhelm something somewhere.

Stebalien:
Note: I really don't care where, what, or how we do this as long as:

1. It's easy to write structured logs somewhere.
2. I can easily collect these structured logs.
3. This feature lands ASAP.

pkg/runner/cluster_k8s.go
nonsense: @Stebalien if we talk to the k8s API, then after a test is detected to be complete, we can get the logs, and only then remove the pods. If we talk directly to the docker daemon from the `sidecar`, I am not sure how we can synchronise the fact that testground will be removing the pods after the test plan completes - how do we get guarantee that all logs have been streamed to the sidecar from all instances?

That synchronisation is easier on Kubernetes, because we control the testplan flow there.

Stebalien: Hm. _That's_ a good point. Given that, I'd really consider redirecting stderr to a file within the container.

nonsense:
Superceded by https://github.com/ipfs/testground/pull/467

PR #443, opened 2020-01-31
compositions: follow-up fixes

raulk:
* Validate compositions appropriately based on context: is it a build or a run?
* Migrate `builder` and `runner` flags to normal string flags. The previous logic leaked, as runners and builders are maintained and operated on the daemon side, and the client should not instantiate an engine.
* Adjust `cmd/itests` to use the `single` variant of build and run commands; also to pass in the instance count.
* Mutex to avoid concurrent initialisations of the `local:docker` runner.

---

Note: `TestSidecar` times out, but it also did in master previous to merging compositions.

### DIFF
--- a/cmd/all.go
+++ b/cmd/all.go
@@ -38,7 +38,7 @@ var Flags = []cli.Flag{
 }
 
 func setupClient(c *cli.Context) (*client.Client, error) {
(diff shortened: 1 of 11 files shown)


raulk:
Nothing controversial here. Just fixes. Merged.

PR #444, opened 2020-02-02
fix: use the auto-detected local:docker socket location

Stebalien:
E.g., this might be set via DOCKER_HOST.

Note: If we're using a TCP socket, we still guess /var/run/docker.sock. In that case, we might be running on another machine. We could probably get smarter about this, but this is likely good enough.


nonsense:
Right - I've never ran the builder on another machine, but this makes sense. LGTM.

Issue #416, opened 2020-01-27
k8s cluster autoscaling doesn't work


nonsense:
There is a problem when autoscaling a Testground Kubernetes cluster, due to the dependencies between the various daemonsets.

1. We need to have networking daemonset installed first - `kube-flannel-ds-amd64`
2. We need to have at least one pod on the new host attached to the flannel network - done with the `dummy` daemonset.
3. Only then we can run the `weave-net` daemonset, which detects the `cni0` interface created in 2. and adds a route to DNS on every host.

---

Currently this is not working, because Kubernetes is starting every daemonset at the same time, and there are no explicit dependencies between them. We should probably address this with `initContainers` on each that wait for its dependencies, for example with: https://github.com/yogeshlonkar/pod-dependency-init-container

---

This problem is a result of https://github.com/ipfs/testground/blob/master/infra/k8s/kops-weave/weave.yml#L171 - the `initContainer` that we run prior to setting up the `weave` container. 

We need to run this `job` once on every host, and so far I've added it as an `initContainer` to `weave`, which is rather wrong, but does the trick.

This job is necessary to be run before we start `sidecar`, as `sidecar` is connecting to the host network and needs access to the cluster DNS.

Related: https://github.com/kubernetes/kubernetes/issues/64623

nonsense:
Actually moving the `iproute` `initContainer` to `sidecar.yml` is solving most of these.

The script that the `initContainer` runs needs to be adjusted to detect if the DNS route is already present (in case we want to restart the `sidecar` daemonset).

gmasgras:
It should also be possible to pass in `--networking flannel` into `kops create cluster` so that the flannel is automatically installed when the nodes first boot up.

nonsense:
@gmasgras thanks for the suggestion, I completely forgot about that, and this removed 1-2 steps from the process!

PR #446, opened 2020-02-03
additional addons

coryschwartz:
Some things can't be specified on the CLI

These are a few things testers might want access to, but aren't currently included
Add prometheus metrics
Add metrics server for kubernetes horizonal pod autoscaling
Add configuration needed for horizontal pod autoscaling

Not included in this patch is the cluster auto-scaler.

When this is complete, I will try to include the weave installation with working auto-scaling, taking the problems already noted in https://github.com/ipfs/testground/issues/416

### DIFF
--- a/infra/k8s/cluster.yaml
+++ b/infra/k8s/cluster.yaml
@@ -1,0 +1,105 @@
+apiVersion: kops.k8s.io/v1alpha2
+kind: Cluster
+metadata:
+  creationTimestamp: null
+  name: testground.cory.systems
+spec:
+  addons:
+  - manifest: metrics-server
+  - manifest: prometheus-operator
+  - manifest: kubernetes-dashboard
+  api:
+    dns: {}
+  authorization:
+    rbac: {}
+  channel: stable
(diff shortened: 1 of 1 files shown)


nonsense:
@coryschwartz thanks for your contributions :)

1. We definitely need some cluster-wide monitoring - prometheus seems like a good fit. It would be great if users of `testground` know how much resources their testplans are using, and can see CPU/memory/disk/network usage.

2. Not sure if we need cluster auto-scaler during the `self-serviced` phase. This is up to @raulk and team to decide, but I think that would make the system even more complex, and would personally suggest postponing until it is a problem - right now we assume that developers will know what kind of cluster size they will need - the calculation depending on the testplan they work on is straight-forward. If this is too big or too small, they are 1 command away from resizing and waiting for 2min for machines to come up or disappear. (provided the `sidecar` `initContainer` is fixed as you mentioned #416)

coryschwartz:
I'm going to cancel this one in the mean time, thanks for keeping me on the right track @nonsense

PR #448, opened 2020-02-03
fix parsing of TEST_INSTANCE_PARAMS

nonsense:
Fixes: https://github.com/ipfs/testground/issues/447


nonsense:
Merging this because cluster:k8s is not working at all currently, and this is fixing it. If we want to refactor the way we handle runtime envs, let's do it in a follow-up PR.

PR #451, opened 2020-02-03
improved usage docs

nonsense:
This PR is fixing a bit the USAGE docs, which are linked from the main page of Testground - https://github.com/ipfs/testground/

1. I remove the reference to `SPEC.md`, which is the first thing that we tell users in the `USAGE` section. The `SPEC.md` is rather outdated, so it is very misleading to tell that to users in their first step. Furthermore `SPEC.md` is linked from the main page, right below the link to `USAGE` - no need to add the same link at multiple high-profile places.

2. Removed the reference to the automatic daemon that we used to start - we haven't done that for a while, so probably no-one cares (unless someone runs a commit from 2 months ago, which I doubt).

3. Remove the sample `run` command which is in the section that explains the Testground environment - it is misleading because it says to run a test on `cluster:swarm`, and we haven't spoken about any environments at that time. Furthermore below we have detailed explanation on how to run test plans via local:docker and outside of Testground.

4. Small fixes to links and rendering of COMPOSITIONS.md

5. Remove of `WIP` at the top of the project space. We've been working on this project for a while, it should be in a usable state now.

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -1,10 +1,5 @@
 # Usage
 
-We kindly ask you to read through the [SPEC](./SPEC.md) first and give this
-project a run first in your local machine. It is a fast moving project at the
-moment, and it might require some tinkering and experimentation to compensate
-for the lack of documentation.
-
 ## Setup
 
 Ensure that you are running go 1.13 or later (for gomod support):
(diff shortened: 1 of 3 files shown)


docs/USAGE.md
```diff
@@ -106,9 +79,9 @@ The client CLI will also expect to find the daemon at `http://localhost:8042`.
 To configure a custom endpoint address, refer to the `[client]` settings on the
 [env-example.toml](../env-example.toml) file at the root of this repo.
 
-## Running the tests locally with Testground
+## Running test plans locally with Testground
```
raulk: What's actually run is a test case, though. Suggest fixing it here and in the prose.

docs/USAGE.md
```diff
@@ -155,9 +128,9 @@ You should see a bunch of logs that describe the steps of the test, from:
 
 
 
-## Running a test outside of Testground orchestrator
+## Running a test plan outside of the Testground orchestrator
 
-You must have a redis instance running locally. Install it for your runtime follow instruction at https://redis.io/download.
+You must have a Redis instance running locally. Install it for your runtime follow instructions at https://redis.io/download.
```
raulk:
```suggestion
You must have a Redis instance running locally. Install it for your runtime following instructions at https://redis.io/download.
```

docs/USAGE.md
```diff
@@ -167,7 +140,7 @@ Then run it locally with
 93801:M 03 Oct 2019 14:42:52.430 * Ready to accept connections
 ```
 
-Then move into the folder that has the plan and test you want to run locally. Execute it by sessting the TEST_CASE & TEST_CASE_SEQ env variables
+Then move into the folder that has the plan and test you want to run locally. Execute it by setting the TEST_CASE & TEST_CASE_SEQ environment variables:
```
raulk: Might as well backtick these.

```suggestion
Then move into the folder that has the plan and test you want to run locally. Execute it by setting the `TEST_CASE` & `TEST_CASE_SEQ` environment variables:
```

docs/USAGE.md
nonsense: I am just being a bit explicit with saying `test plan`, so that it is clear we are referring to the plans, and not for example to Testground's tests.

PR #452, opened 2020-02-03
use shorter runid

nonsense:
I think it is nicer to use a shorter runid for test plans, because then we don't have gigantic folder name, pod names, etc.

12 characters with an alphabet of 26+10 is enough entropy IMO.

---

Also adding the `RunID` as part of pod name, so that we can run multiple testplans and also use `keep_service=true`.


nonsense:
Probably worth doing this for `build` ids as well, just so that we are consistent?

Stebalien:
Yes.

PR #456, opened 2020-02-03
feat: use contexts for the sync service

Stebalien:
This patch also simplifies several parts of the sync service given that we can now just cancel contexts to, e.g., cancel subscriptions.


sdk/sync/common.go
```diff
@@ -69,12 +70,12 @@ func redisClient(runenv *runtime.RunEnv) (client *redis.Client, err error) {
...
 
 	// PING redis to make sure we're alive.
-	return client, client.Ping().Err()
+	return client, client.WithContext(ctx).Ping().Err()
 }
 
 // MustWatcherWriter proxies to WatcherWriter, panicking if an error occurs.
-func MustWatcherWriter(runenv *runtime.RunEnv) (*Watcher, *Writer) {
-	watcher, writer, err := WatcherWriter(runenv)
+func MustWatcherWriter(ctx context.Context, runenv *runtime.RunEnv) (*Watcher, *Writer) {
```
Stebalien: note: this context is only used for _connecting_ to redis. Every request gets its own context.

sdk/sync/common.go
```diff
@@ -43,7 +44,7 @@ func redisClient(runenv *runtime.RunEnv) (client *redis.Client, err error) {
 		// Fall back to attempting to use `host.docker.internal` which
 		// is only available in macOS and Windows.
 		for _, h := range []string{RedisHostname, HostHostname} {
-			if addrs, err := net.LookupHost(h); err == nil && len(addrs) > 0 {
+			if addrs, err := net.DefaultResolver.LookupHost(ctx, h); err == nil && len(addrs) > 0 {
```
Stebalien: Probably wasn't the issue but can't hurt to fix.

sdk/sync/subscription.go
```diff
@@ -52,8 +35,7 @@ func (s *subscription) process() {
 
 	startSeq, err := s.client.XLen(key).Result()
 	if err != nil {
-		s.connCh <- -1
-		s.result = fmt.Errorf("failed to fetch current length of stream: %w", err)
+		s.w.re.SLogger().Errorf("failed to fetch current length of stream: %w", err)
```
Stebalien: Before: We'd learn about errors from `Close` and log them.
Now: We just log the errors.

The "correct" way to do this is probably to return an error channel from `Subscribe`. Is that worth it?

sdk/sync/subscription.go
```diff
@@ -64,27 +46,17 @@ func (s *subscription) process() {
...
-	s.connCh <- id
-
 	args := &redis.XReadArgs{
 		Streams: []string{key, "0"},
 		Block:   0,
 	}
 
 	var last redis.XMessage
-	for !s.isClosed() {
+	for {
```
Stebalien: The connection _should_ obey the context that we've set on the client. We shouldn't need to play this close dance.

sdk/sync/watch.go
```diff
@@ -47,57 +44,45 @@ func NewWatcher(runenv *runtime.RunEnv) (w *Watcher, err error) {
...
+// occurs.
+func (w *Watcher) Subscribe(ctx context.Context, subtree *Subtree, ch interface{}) (context.CancelFunc, error) {
+	ctx, cancel := context.WithCancel(ctx)
+
+	if err := ctx.Err(); err != nil {
+		return cancel, err
+	}
+	if err := w.client.Context().Err(); err != nil {
+		return cancel, err
+	}
```
Stebalien: Just some nice sanity checks.

sdk/sync/watch.go
```diff
@@ -47,57 +44,45 @@ func NewWatcher(runenv *runtime.RunEnv) (w *Watcher, err error) {
...
 // returned errCh.
 //
-// The user can cancel the subscription by calling the returned cancelFn. The
-// subscription will die if an internal error occurs, in which case the cancelFn
-// should also be called.
-func (w *Watcher) Subscribe(subtree *Subtree, ch interface{}) (func() error, error) {
+// The user can cancel the subscription by calling the returned cancelFn or by
+// canceling the passed context. The subscription will die if an internal error
+// occurs.
+func (w *Watcher) Subscribe(ctx context.Context, subtree *Subtree, ch interface{}) (context.CancelFunc, error) {
```
Stebalien: I'm still returning a cancel function to make it clear that the context _must_ be canceled.

sdk/sync/watch.go
```diff
@@ -61,43 +65,25 @@ func (w *Watcher) Subscribe(subtree *Subtree, ch interface{}) (func() error, err
...
-
-		delete(w.subtrees[subtree], sub)
-		if len(w.subtrees[subtree]) == 0 {
-			delete(w.subtrees, subtree)
-		}
+	w.subs.Add(1)
+	go func() {
+		defer w.subs.Done()
+		sub.process()
+	}()
```
Stebalien: The subscriptions will automatically shutdown when the context is canceled. No need to track them.

pkg/sidecar/sidecar.go
```diff
@@ -92,22 +92,16 @@ func Run(runnerName string, resultPath string) error {
 			// Now let the test case tell us how to configure the network.
 			subtree := sync.NetworkSubtree(instance.Hostname)
 			networkChanges := make(chan *sync.NetworkConfig, 16)
-			closeSub, err := instance.Watcher.Subscribe(subtree, networkChanges)
-			if err != nil {
+			if _, err := instance.Watcher.Subscribe(ctx, subtree, networkChanges); err != nil {
```
nonsense: Shouldn't we have a `closeSub` in a `defer` statement?

Stebalien: Doesn't matter now. The subscription will be canceled when the context is canceled.

We only return that function (a) for convenience and (b) to indicate to the user that the subscription needs to be released somehow.

Thoughts? Would it be cleaner to remove it? Should we just always call it?

nonsense: I only raised this, because I saw we used to call it - previously cancel was called when this function returns, and now it is called when the context is canceled and I am not sure if this is the same time, but it looks like it is the same time, so as you said - doesn't matter.

nonsense:
LGTM. I also tested this and we don't appear to have broken anything, and it is mostly equivalent to before.

sdk/sync/watch.go
nonsense: Not sure about this pattern... in generally usage of a package/function should be more or less clear. Those functions that require the user to do something else (call `Close()`, or cancel a context), and mostly always documented.

So I am not sure we should add a return value, but you do have a point this is more explicit.

nonsense: What I am trying to say is that we are adding a return value, and we STILL can't know if the user will call it, so we might as well not return it an expect correct usage...

Stebalien: SGTM. I'll just go ahead and remove it.

Stebalien:
Hm. We're not running the sync service tests in CI.

sdk/sync/subscription.go
```diff
@@ -52,39 +35,50 @@ func (s *subscription) process() {
...
+			// no need to unblock anything.
+		}
+	}()
 
-	// store the conn ID in the channel.
-	s.connCh <- id
+	defer func() {
+		close(done)
+		<-closed
+	}()
```
Stebalien: @nonsense thoughts on how to write this better? This works, it's just a little messy (cleaner than requiring a separate stop method, IMO, but still messy.

nonsense:
@Stebalien sorry for the conflicts. I was under the impression we had merged this! Will review the new changes now.

plans/bitswap-tuning/test/transfer.go
```diff
@@ -93,37 +93,38 @@ func Transfer(runenv *runtime.RunEnv) error {
...
+		if _, err = writer.Write(ctx, RootCidSubtree, &rootCid); err != nil {
 			return fmt.Errorf("Failed to get Redis Sync RootCidSubtree %w", err)
 		}
 	} else if isLeech {
 		// Get the root CID from a seed
 		rootCidCh := make(chan *cid.Cid, 1)
-		cancelRootCidSub, err := watcher.Subscribe(RootCidSubtree, rootCidCh)
-		if err != nil {
+		sctx, cancelRootCidSub := context.WithCancel(ctx)
+		if err := watcher.Subscribe(sctx, RootCidSubtree, rootCidCh); err != nil {
```
nonsense: We are not calling `cancelRootCidSub` on `error`?  I think it is good to always call cancel, and if we add it as a defer at one place it would be easier to reason about it.

plans/dht/test/common.go
```diff
@@ -389,23 +388,24 @@ func Bootstrap(ctx context.Context, runenv *runtime.RunEnv, watcher *sync.Watche
 
 // get all bootstrap peers.
 func getBootstrappers(ctx context.Context, runenv *runtime.RunEnv, watcher *sync.Watcher, opts *SetupOpts) ([]peer.AddrInfo, error) {
+	// cancel the sub
+	ctx, cancel := context.WithCancel(ctx)
```
nonsense: Here for example we call `defer cancel()` right after creating the context - not consistent with how we use the context above with `watcher.Subscribe`.

sdk/sync/subscription.go
```diff
@@ -19,29 +18,13 @@ type subscription struct {
...
-}
-
-func (s *subscription) isClosed() bool {
-	select {
-	case <-s.closeCh:
-		return true
-	default:
-		return false
-	}
+	cancel context.CancelFunc
```
nonsense: This doesn't appear to be used.

nonsense:
Adding @raulk as well as original author of this code. The code in this service is a bit complex for my taste, and I am still trying to understand all the inner workings of this service and how the various contexts and cancels fit together.

raulk:
@nonsense you should've seen it before we switched to redis pubsub 😆

plans/dht/test/common.go
raulk: Should be functionally equivalent, but yeah, we should defer the cancel immediately above to avoid a goroutine leak: https://golang.org/pkg/context/#WithCancel.

plans/dht/test/common.go
```diff
@@ -389,23 +388,24 @@ func Bootstrap(ctx context.Context, runenv *runtime.RunEnv, watcher *sync.Watche
...
-	defer cancelSub()
 
 	// TODO: remove this if it becomes too much coordination effort.
 	peers := make([]peer.AddrInfo, opts.NBootstrap)
 	// Grab list of other peers that are available for this run.
 	for i := 0; i < opts.NBootstrap; i++ {
-		select {
-		case ai := <-peerCh:
-			peers[i] = *ai
-		case <-ctx.Done():
```
raulk: Nice.

sdk/sync/common.go
raulk: Would be good to note that in the godocs.

sdk/sync/common.go
```diff
@@ -83,13 +84,13 @@ func MustWatcherWriter(runenv *runtime.RunEnv) (*Watcher, *Writer) {
 
 // WatcherWriter creates a Watcher and a Writer object associated with this test
 // run's sync tree.
-func WatcherWriter(runenv *runtime.RunEnv) (*Watcher, *Writer, error) {
-	watcher, err := NewWatcher(runenv)
+func WatcherWriter(ctx context.Context, runenv *runtime.RunEnv) (*Watcher, *Writer, error) {
```
raulk: Same here; let's document that this context doesn't get stored and it's only used while connecting to the backend.

sdk/sync/common.go
```diff
@@ -28,7 +29,7 @@ const (
 //
 // TODO: source redis URL from environment variables. The Redis host and port
 // will be wired in by Nomad/Swarm.
-func redisClient(runenv *runtime.RunEnv) (client *redis.Client, err error) {
+func redisClient(ctx context.Context, runenv *runtime.RunEnv) (client *redis.Client, err error) {
```
raulk: Let's make clear that this context is only used for resolving the hostname; it in no way governs the lifecycle of the redis client. It's pretty confusing.

Stebalien:
Let's get https://github.com/ipfs/testground/pull/499 merged first. This needs to be tested in CI.

sdk/sync/subscription.go
Stebalien: It doesn't but I've fixed that.

Stebalien:
> My only concern is that the context passed to the constructors {Must,New}{Watcher/Writer/WatcherWriter} and redisClient is very misleading, as it doesn't govern the lifecycle of the components that are bring constructed. Not sure how to best warn about that. I guess in godocs? Can't think of a different API shape that would be ergonomic enough to avoid the confusion in the first place.

Note: this is _mostly_ confusing because of how we abuse contexts. Contexts are supposed to govern request lifetime, not service lifetime.

I've documented it now.

PR #457, opened 2020-02-03
feat(sidecar): stop waiting for workers to finish

Stebalien:
1. Log an error every 5 seconds when waiting.
2. Panic if we wait too long.

We don't want to just move on because we have a goroutine hanging.

### DIFF
--- a/pkg/dockermanager/dockermanager.go
+++ b/pkg/dockermanager/dockermanager.go
@@ -13,6 +13,9 @@ import (
 	"github.com/ipfs/testground/pkg/logging"
 )
 
+const workerShutdownTimeout = 1 * time.Minute
+const workerShutdownTick = 5 * time.Second
+
 // Manager is a convenient wrapper around the docker client.
 type Manager struct {
 	logging.Logging
@@ -123,7 +126,24 @@ func (dm *Manager) Manage(
 		if m, ok := managers[container]; ok {
 			m.cancel()
 			delete(managers, container)
-			<-m.done
+
+			timeout := time.NewTimer(workerShutdownTimeout)
+			defer timeout.Stop()
+
+			ticker := time.NewTicker(workerShutdownTick)
+			defer ticker.Stop()
+
+			for {
+				select {
+				case <-m.done:
+					return
+				case <-timeout.C:
+					dm.S().Panicw("timed out waiting for container worker to stop", "container", container)
+					return
+				case <-ticker.C:
+					dm.S().Errorw("waiting for container worker to stop", "container", container)
+				}
+			}
 		}
 	}
 	start := func(container string) {


nonsense:
> There is a container that can't be stopped for some reason.
> The sidecar panics after 50 seconds

The sidecar should not panic if there is a misbehaving container in my opinion. The sidecar is needed for many testplans, and is a service required for the operation of testground. It's handling of various containers should be independent from one another.

> perhaps the daemon starts a new sidecar while the misbehaving container is still running.

Not sure what you mean with `daemon`, but if you mean `testground daemon` - it is not the `testground daemon` that is setting up the environment for testground - `testground daemon` expect its dependencies to be available, and these are: `redis` and `sidecar`.

If the `sidecar` panics, it should be restart automatically - since it is in a DaemonSet or on the local machine, due to the configuration of the container.

> It's not clear to me whether the misbehaving container is still being managed still has network access or can interfere with future tests.

Misbehaving containers should not interfere with future tests.

---

I've also had similar thoughts when the sidecar hangs from one run to another, and these are my opinions on how it should work.

pkg/dockermanager/dockermanager.go
```diff
@@ -13,6 +13,8 @@ import (
 	"github.com/ipfs/testground/pkg/logging"
 )
 
+const instanceWorkerShutdownStallTick = 5 * time.Second
```
raulk: That's a mouthful :-)

Stebalien:
This issue isn't about misbehaving tests, it's about the sidecar misbehaving. That is, we've canceled the container's sidecar worker (not a part of the container) but the worker refuses to stop.

Stebalien:
@raulk done.

pkg/dockermanager/dockermanager.go
```diff
@@ -123,7 +126,24 @@ func (dm *Manager) Manage(
...
+
+			ticker := time.NewTicker(workerShutdownTick)
+			defer ticker.Stop()
+
+			for {
+				select {
+				case <-m.done:
+					return
+				case <-timeout.C:
+					dm.S().Panicw("timed out waiting for container worker to stop", "container", container)
```
nonsense: I'm not sure why we would ever panic the sidecar on a problem with a container.

If `docker` (or k8s api) goes berserk and doesn't remove a container, we should just flood `sidecar` with errors, but not panic it. One container should not be bringing the sidecar on a given host down.

nonsense: To put more context here, if we run test plans overnight and there is a problem with Docker for some reason, I don't want to see `sidecar` having restarted 200 times, because we somehow managed to break another part of the system - I'd like to see 200 errors and have to figure out what has gone wrong.

With respect to the hanging go-routines - just close them after a timeout for containers that are not responding if this ever happens, and log an error.

Stebalien: This is a problem in the sidecar, not containers. We'll only hit this if the sidecar is stuck doing something and isn't returning from the worker. (see https://github.com/ipfs/testground/pull/456 for a case where that can happen).

I can't close hanging goroutines. Go doesn't let me do that.

Basically, if the sidecar is stuck, there's a bug somewhere in the sidecar and we need to do _something_. Crashing is about the best thing we can do.

nonsense: OK, then I misunderstood what this is about!

PR #458, opened 2020-02-04
sidecar installation - wait for redis

nonsense:
This PR is adding a check requiring redis to be running before we start sidecar.

It makes it possible to install all testground services at the same time without them crashing.


nonsense:
We don't necessarily have to do that, because the sidecar will panic if Redis is not available, but I think it is nice to prevent going into a crash backoff loop. Also makes it easier to detect if sidecar failed afterwards due to an error we have to fix.

PR #459, opened 2020-02-04
improved installation process

nonsense:
This PR is adding an `install.sh` script which goes through all the steps from creating the cluster to the installation of Testground dependencies.

If you have a `cluster.yaml` file, you only have to set the ENV variables (ideally in your .rc file) and have a one-command to start and later destroy the cluster.

---

Also added a note on how to `resize` the cluster in case you figure out that you need more or less machines than originally planned.

---

On my end it takes 6:30min for the cluster to be ready:
```
Testground cluster is ready

Execution time was 383 seconds
```

### DIFF
--- a/infra/k8s/README.md
+++ b/infra/k8s/README.md
@@ -19,7 +19,7 @@ We use Weave for the `data` plane on Testground - a secondary overlay network th
 
 `kops` uses 100.96.0.0/11 for pod CIDR range, so this is what we use for the `control` network.
 
-Weave by default uses 10.32.0.0/11 as CIDR, so this is the CIDR for the Testground `data` network. The `sidecar` is responsible for setting up the `data` network for every testplan instance.
+We configure Weave to use 16.0.0.0/4 as CIDR (we want to test `libp2p` nodes with IPs in public ranges), so this is the CIDR for the Testground `data` network. The `sidecar` is responsible for setting up the `data` network for every testplan instance.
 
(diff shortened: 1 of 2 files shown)


infra/k8s/README.md
```diff
@@ -81,63 +85,23 @@ kops create cluster \
...
 helm repo add stable https://kubernetes-charts.storage.googleapis.com/
 helm repo update
 ```
 
-2. Create a `Redis` service on your Kubernetes cluster
-
+7. Apply cluster setup and install Testground dependencies
 ```
-helm install redis stable/redis --values ./infra/k8s/redis-values.yaml
+./install.sh $NAME $CLUSTER_SPEC $PUBKEY $WORKER_NODES
```
nonsense: Note that this file is not idempotent. Then again it is fairly straight-forward. In case it has to become any more elaborate than it already is, we might have to consider other tools than `bash`.

infra/k8s/README.md
```diff
@@ -81,63 +85,24 @@ kops create cluster \
...
-1. Set up Helm
+6. Set up Helm and add the `stable` Helm Charts repository
 
 ```
 helm repo add stable https://kubernetes-charts.storage.googleapis.com/
 helm repo update
 ```
 
-2. Create a `Redis` service on your Kubernetes cluster
+## Apply the cluster configuration and create cloud resources and install Testground dependencies
```
nonsense: When you've gone through the steps above once, if you've recorded the cluster spec (`cluster.yaml`) and your env variables, next time you can just call `install.sh`.

PR #463, opened 2020-02-04
feat/collect cmd

nonsense:
This PR is building on top of https://github.com/ipfs/testground/pull/438 and adding zip functionality to the daemon for the `cluster:k8s`.


raulk:
I've changed the base branch of this to point to `feat/test-outputs`, and I've merged the latter into here. I'll work on the `local:exec` and `local:docker` collection methods here.

raulk:
Ok, `local:exec` and `local:docker` collection is implemented.

pkg/runner/local_docker.go
```diff
@@ -307,7 +307,7 @@ func (r *LocalDockerRunner) Run(ctx context.Context, input *api.RunInput, ow io.
 		return nil, pretty.Wait()
```
aschmahmann: We should learn about the runID here too. For example:

```suggestion
              return &api.RunOutput{RunID: input.RunID}, pretty.Wait()
```

PR #469, opened 2020-02-05
[draft] IPFS E2E Plan

coryschwartz:
Starting work for https://github.com/ipfs/testground/issues/465

Comments welcome, if you have time.


plans/ipfs-e2e/test/testcase.go
```diff
@@ -0,0 +1,108 @@
...
+
+// This test runs for each file size tested.
+// 1. Leechers enter the "not leaching" state.
+// 2. Seeders wait for all leachers to enter the "not leaching" state.
+// 3. Now that all the leachers are ready, seeders transition to the "seeding" state
+// 4. The leechers wait until both seeders have transitioned to seeding before they proceed with
+// leeching.
+// You might be wondering... why not just start leeching as soon as one seeder is available? I want to
+// prevent a race condition where one seeder starts, and the other fails to start because there are
+// not enough leechers still in the "not seeding" state".
```
coryschwartz:
```suggestion
// not enough leechers still in the "not leeching" state".
```

dirkmc:
Looks good so far - I think I will change the bitswap tests to have a structure more like this, with a test case struct that holds all the variables for a single case.

A couple of gotchas to watch out for when running this kind of test, with local:docker I had to
- generate seed data sequentially for each seed
  if I generated the seed data on every instance at the same time my computer would melt down and the instances would fail to communicate to each other through Redis in weird ways
- clean up memory at the end of each run
  I do several runs of each test case, and test several file sizes, so my instances were running out of memory

raulk:
We have removed the IPTB SDK module as it was contrary to the spirit of Testground. The philosophy of Testground is to hit the code under test directly, rather than by puppeteering daemons.

PR #471, opened 2020-02-05
use folder with index name ; sidecar should not logs to /outputs ; toNet should not panic

nonsense:
This PR is:
- [x] 1. changing `cluster_k8s` runner to log into directory with the index of the current instance from the group and not the pod name
- [x] 2. fixing bug with parsing empty subnets - sidecar is receiving all kinds of containers, not all are going to need it services
- [x] 3. removing the TEST_OUTPUTS env var from the sidecar config so that it doesn't try to log in /outputs directory

### DIFF
--- a/sdk/runtime/runenv.go
+++ b/sdk/runtime/runenv.go
@@ -171,6 +171,7 @@ func toBool(s string) bool {
 	return v
 }
 
+// toNet might parse any input, so it is possible to get an error and nil return value
 func toNet(s string) *IPNet {
 	_, ipnet, err := net.ParseCIDR(s)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


sdk/runtime/runenv.go
```diff
@@ -171,8 +171,12 @@ func toBool(s string) bool {
 	return v
 }
 
+// toNet might parse any input, so it is possible to get an error and nil return value
 func toNet(s string) *IPNet {
-	_, ipnet, _ := net.ParseCIDR(s)
+	_, ipnet, err := net.ParseCIDR(s)
```
raulk: Could maybe get rid of `toNet` and use `IPNet#UnmarshalJSON` instead, but that would not return a nil *IPNet?

nonsense: Let's revisit this after the release, as we might need a slightly bigger refactor of the env vars in these methods.

PR #482, opened 2020-02-05
Update docker and x/sys for Windows support

aschmahmann:
x/sys and docker not both playing it fast and loose with dependencies has been causing us problems, particularly on Windows. Perhaps this issue is related to golang breaking x/sys on Windows (and then unbreaking it, which is a breaking a change), or maybe this will be an ongoing issue.

The fix: Updated the docker version and the x/sys replace directive

How it was achieved:
1) Look through docker repo for an issue/PR related to x/sys https://github.com/moby/moby/pull/40312
2) Find the commit as it exists in the tree https://github.com/moby/moby/commit/de30f30e905488112a6c0c20fa30ca93fc1b22d3 and then update the two deps


aschmahmann:
This was fixed by #489

PR #483, opened 2020-02-06
fix: don't hide output from go commands

Stebalien:


pkg/build/golang/docker.go
```diff
@@ -192,7 +192,7 @@ func (b *DockerGoBuilder) Build(ctx context.Context, in *api.BuildInput, output
 	// Write replace directives.
 	cmd := exec.CommandContext(ctx, "go", append([]string{"mod", "edit"}, replaces...)...)
 	cmd.Dir = plandst
-	_, err = cmd.CombinedOutput()
+	err = cmd.Run()
```
coryschwartz: This runs successfully, but if the goal is not to hide the command output, the output is still hidden like this.

To display the output you'd have to attach a buffer to stdout and stderr, and then later print out those buffers.

I think the use of CombinedOutput is actually correct, just don't assign the bytes buffer to _.

```suggestion
	output, err := cmd.CombinedOutput()
```

pkg/build/golang/docker.go
```diff
@@ -192,7 +192,7 @@ func (b *DockerGoBuilder) Build(ctx context.Context, in *api.BuildInput, output
...
 	if err != nil {
 		return nil, fmt.Errorf("unable to add replace directives to go.mod; %w", err)
```
coryschwartz:
```suggestion
		return nil, fmt.Errorf("unable to add replace directives to go.mod; %w", err, output)
```

coryschwartz:
Actually, I was wrong in my previous comments, the err returned by cmd.Run is this one:
https://golang.org/src/os/exec/exec.go?s=13315:13903#L445

The error *does* contain the stderr, but only the process state is returned in ExitError.Error(). If convident the error is always an ExitError, you could print the stderr from there.

pkg/build/golang/docker.go
Stebalien: Ah, you're right. Go sends the output to /dev/null by default.

Stebalien:
@coryschwartz I've changed it to _actually_ send the output to stderr/stdout.

PR #484, opened 2020-02-06
More useful comment for NetworkSubtree

coryschwartz:
I think there was a bad copy-paste for this comment :)


nonsense:
@coryschwartz could you run `go mod tidy` and send another commit please, so that CI works?

coryschwartz:
Was failing for the windows build, which was fixed earlier today in https://github.com/ipfs/testground/issues/490 so I pulled in go.mod and go.sum from master.

PR #501, opened 2020-02-07
fix(make): fail on non-zero status

Stebalien:
Apparently, find doesn't return a non-zero status when a command exits with a non-zero status.


Stebalien:
Oops. Well, the tests were _running_ at least.

Stebalien:
Status: currently rebased on #501. We should merge that first (or just merge this one and close that one).

PR #502, opened 2020-02-07
feat: distinguish between runenv and run params

Stebalien:
The RunEnv is for the testcase and needs an initialized logger, etc. The run params are just data.

This prevents us from, e.g., calling functions that try to access uninitialized fields. This _also_ lets test cases (e.g., the redis test cases) safely construct fully initialized run environments.

### DIFF
--- a/sdk/sync/redis_test.go
+++ b/sdk/sync/redis_test.go
@@ -363,7 +363,7 @@ func randomRunEnv() *runtime.RunEnv {
 
 	_, subnet, _ := net.ParseCIDR("127.1.0.1/16")
 
-	return &runtime.RunEnv{
+	return runtime.NewRunEnv(runtime.RunParams{
 		TestPlan:           fmt.Sprintf("testplan-%d", rand.Uint32()),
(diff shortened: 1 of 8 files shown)


sdk/sync/redis_test.go
```diff
@@ -363,7 +363,7 @@ func randomRunEnv() *runtime.RunEnv {
 
 	_, subnet, _ := net.ParseCIDR("127.1.0.1/16")
 
-	return &runtime.RunEnv{
+	return runtime.NewRunEnv(runtime.RunParams{
```
Stebalien: This is the bugfix. `RunEnv` was uninitialized.

Stebalien: Fixes the bug found in https://github.com/ipfs/testground/pull/501.

Stebalien:
> We can then make the sidecar use RunParams for its managed instances, instead of a fully fledged RunEnv.

I still need RunEnv, unfortunately, because I need to talk to the sync service. But I can definitely remove that hack.

PR #503, opened 2020-02-07
fix: remove a runenv hack in the sidecar

Stebalien:
Instead of parsing the environment, modifying it, then re-serializing it; parse it without constructing a runenv, modify it, then construct a runenv.

fixes https://github.com/ipfs/testground/pull/502#pullrequestreview-355481906

### DIFF
--- a/pkg/sidecar/docker_instance.go
+++ b/pkg/sidecar/docker_instance.go
@@ -8,8 +8,6 @@ import (
 	"net"
 	"os"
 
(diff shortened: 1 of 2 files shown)


nonsense:
We still have this hack on `k8s` side. Shouldn't it be updated as well?

nonsense:
Possibly related, but not sure: https://github.com/ipfs/testground/issues/514

PR #507, opened 2020-02-10
testground clean-up / reset script

nonsense:
We've spoken that it'd be nice to have a script that resets the Testground environment, so here is an initial attempt at it.

This is removing all pods created by Testground, and restarting Sidecar.

I don't think we have to restart Redis or s3bucket for now, as we've never seen them fail for now.

### DIFF
--- a/infra/k8s/clean.sh
+++ b/infra/k8s/clean.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+START_TIME=`date +%s`
+
+echo "Resetting Testground environment..."
+echo
+
+kubectl delete pods -l app=testground --force --grace-period=0
+kubectl delete pods -l name=testground-sidecar --force --grace-period=0
+echo
+
+sleep 5
+
+echo "Wait for Sidecar to be Ready..."
+echo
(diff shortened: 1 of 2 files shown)


raulk:
This is part of the healthcheck / repair epic.

infra/k8s/clean.sh
```diff
@@ -0,0 +1,28 @@
...
+set -o errexit
+set -o pipefail
+set -o nounset
+
+START_TIME=`date +%s`
+
+echo "Resetting Testground environment..."
+echo
+
+kubectl delete pods -l app=testground --force --grace-period=0
```
nonsense: @hacdias that could be the initial health/repair functionality for Kubernetes:
1. Count the number of worker nodes in the cluster at the time of the call to the command (see `maxPods` in cluster_k8s.go)
2. Delete all pods labeled with `app=testground`
3. Delete all sidecar pods
4. Wait for pods with label `app=testground` to be 0 (all states, including Terminated)
5. Wait for worker nodes number of sidecar pods to appear.

---

This doesn't account for the race of somebody running another `testground run` command, while we are terminating, but I'd say this functionality is not something we need right now :)

nonsense:
Closing this, as this will be moved to the `testground terminate` and healthcheck commands (continuation of https://github.com/ipfs/testground/pull/557)

PR #510, opened 2020-02-10
handle canceled context in cluster_k8s.go

nonsense:
Fixes https://github.com/ipfs/testground/issues/474

### DIFF
--- a/pkg/runner/cluster_k8s.go
+++ b/pkg/runner/cluster_k8s.go
@@ -187,10 +187,7 @@ func (*ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow io.Wri
 				if cfg.KeepService {
 					return
 				}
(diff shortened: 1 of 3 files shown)


pkg/runner/cluster_k8s.go
```diff
@@ -187,10 +187,7 @@ func (*ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow io.Wri
 				if cfg.KeepService {
 					return
 				}
-				client, err := pool.Acquire(ctx)
-				if err != nil {
-					log.Errorw("couldn't get client from pool", "pod", podName, "err", err)
-				}
+				client := pool.Acquire()
```
nonsense: If the main context gets canceled, I still want to get a client and do clean-up. The context in the pool is generally not needed.

PR #513, opened 2020-02-10
configurable pod resource requirements

nonsense:


nonsense:
Probably this should be a testplan config, rather than a global runner config, @raulk what do you think?

PR #515, opened 2020-02-10
fix: remove a runenv hack in the sidecar

nonsense:
Maybe fixes https://github.com/ipfs/testground/issues/514 but I haven't had a chance to test yet.


nonsense:
Sorry, I just glanced at it and thought it might be related, didn't think it through!

PR #516, opened 2020-02-10
buffer barrier channel

Stebalien:
Otherwise, we'll leak a goroutine if the user has walked away.

Also, make sure canceling the context actually cancels the barrier (test).

### DIFF
--- a/sdk/sync/redis_test.go
+++ b/sdk/sync/redis_test.go
@@ -140,6 +140,35 @@ func TestBarrier(t *testing.T) {
 	}
 }
 
+func TestBarrierCancel(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	close := ensureRedis(t)
+	defer close()
+
+	runenv := randomRunEnv()
+
+	watcher, err := NewWatcher(ctx, runenv)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer watcher.Close()
+
+	state := State("yoda")
+	ch := watcher.Barrier(ctx, state, 10)
+	cancel()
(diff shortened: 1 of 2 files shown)


sdk/sync/redis_test.go
```diff
@@ -140,6 +140,32 @@ func TestBarrier(t *testing.T) {
...
+	select {
+	case err := <-ch:
+		if err != context.Canceled {
+			t.Errorf("expected context canceled, got %s", err)
+		}
+	case <-time.After(3 * time.Second):
+		t.Error("expected a cancel")
+		return
+	}
+}
```
Stebalien: This _doesn't_ test the change in this PR. I just added this test to make sure barrier cancellation actually works.

Stebalien: Ok, so this _is_ actually broken. It looks like redis treats context like a suggestion, more than anything else.

Stebalien: Nevermind, my test is broken.

sdk/sync/redis_test.go
```diff
@@ -140,6 +140,35 @@ func TestBarrier(t *testing.T) {
...
+
+	state := State("yoda")
+	ch := watcher.Barrier(ctx, state, 10)
+	cancel()
+	select {
+	case err := <-ch:
+		if err == nil {
+			t.Errorf("expected an error")
+		}
+	case <-time.After(3 * time.Second):
```
nonsense: Not sure why this timeout is after 3 seconds, when the context is set to 5 seconds timeout.

nonsense: Oh now I get it, you actually manually call cancel, so this is not at all related with the timeout from above.

PR #527, opened 2020-02-11
Update USAGE.md

FabioMartins91:
Added note to install Docker

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -87,6 +87,8 @@ docker pull ipfs/testground:latest
 
 or
 
+Please, ensure before running the command bellow, make sure you install Docker on your machine 
+
 ```bash
 make docker-ipfs-testground
 ```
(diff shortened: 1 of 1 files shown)


docs/USAGE.md
```diff
@@ -87,6 +87,8 @@ docker pull ipfs/testground:latest
 
 or
 
+Please, ensure before running the command bellow, make sure you install Docker on your machine
```
nonsense: That's a good point. `Testground` doesn't really need `Docker`, but if you want to use the `sidecar`, or the docker-related builder or runners, you have to have it.

I suggest we just add Docker as a requirement at the top, alongside Go, and not in the middle of the docs.

nonsense: Also typo: `Please, before running the command below, make sure you install Docker on your machine.`

FabioMartins91: I agree! On it ;)

PR #531, opened 2020-02-11
Create 2020-02-11

FabioMartins91:
Added 2020-02-11 Weekly Sync Notes


nonsense:
@FabioMartins91 FYI usually `Assignments` is filled in, but this week it is clear we are using the board.

@daviddias has made a really nice checklist of stuff we should cover as part of this call: https://gist.github.com/daviddias/ac0ee538254217b4fe82c20719d6e2f2 - this might need revision, as previously this call consistent of 4-5 people, and now we are close to 10.

PR #533, opened 2020-02-11
Fix some minor issues with the S3 collection logic

Stebalien:
* Use continuations instead of startAfter
* Use contexts.
* Actually download sequentially.


nonsense:
LGTM. However both approaches (master and this PR) are rather slow. It takes almost 2 min with both approaches to download run `9ccc44f28b15`, whereas downloading with `aws s3 cp --recursive` takes 10 seconds.

Maybe we should refactor this to first download and then zip, rather than be forced to download sequentially.

PR #534, opened 2020-02-12
Example cluster.yaml

coryschwartz:
This is a cluster.yaml which has been edited to enable cluster monitoring with Prometheus

This might fix: https://github.com/ipfs/testground/issues/228

The idea here is to include a cluster.yaml rather than generating it with kops create and making small edits, as currently prescribed in the README. This version includes edits necessary to add host monitoring and enables testground users to observe how the cluster nodes are performing.

### DIFF
--- a/infra/k8s/README.md
+++ b/infra/k8s/README.md
@@ -68,9 +68,10 @@ aws s3api create-bucket \
 You might want to add them to your `rc` file (`.zshrc`, `.bashrc`, etc.)
 
 ```
-export KOPS_STATE_STORE=s3://kops-backend-bucket
+export NAME=<desired kubernetes cluster name>
+export KOPS_STATE_STORE=s3://<kops state s3 bucket>
+export ZONE=<aws region>
 export WORKER_NODES=4
-export CLUSTER_SPEC=~/cluster.yaml
 export PUBKEY=~/.ssh/testground_rsa.pub
 
 # details for S3 bucket to be used for assets
@@ -92,7 +93,7 @@ helm repo update
 ## Install the kuberntes cluster
 
 ```
-./install.sh <cluster name> <aws zone> <template> <ssh public key> <number of worker nodes>
+./install.sh  <template file>
 ```
 
 for example, to create a monitored cluster in eu-central-1a with five kubernetes workers:
--- a/infra/k8s/README.md
+++ b/infra/k8s/README.md
@@ -212,6 +212,24 @@ kubectl get pods -o wide
 kubectl logs <pod-id, e.g. tg-dht-c95b5>
 ```
 
+5. Check on the monitoring infrastructure (it runs in the monitoring namespace)
+```
+kubectl get pods --namespace monitoring
+```
+
+6. Get access to teh redis shell
+```
+kubectl port-forward svc/redis-master 6379:6379 &
+redis-cli -h localhost -p 6379
+```
+
+7. Get access to the kubernetes dashboard 
+```
+kubectl proxy
+```
+and then, direct your browser to `http://localhost:8001/ui`
+
+
 
 ## Use a Kubernetes context for another cluster
 
--- a/infra/k8s/README.md
+++ b/infra/k8s/README.md
@@ -68,8 +68,6 @@ aws s3api create-bucket \
 You might want to add them to your `rc` file (`.zshrc`, `.bashrc`, etc.)
 
 ```
-export NAME=my-first-cluster-kops.k8s.local
-export ZONES=eu-central-1a
 export KOPS_STATE_STORE=s3://kops-backend-bucket
 export WORKER_NODES=4
 export CLUSTER_SPEC=~/cluster.yaml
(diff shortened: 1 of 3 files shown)


nonsense:
@coryschwartz this looks like a good start. I agree that if we need more changes to the cluster.yaml, we better provide a ready one with placeholders and use it as a template, rather than ask users to manually edit bits and pieces.

However running this particular `cluster.yaml` locally changes the behaviour and `install.sh` now fails at:

```
Got        1 ready nodes
Got        1 ready nodes
Got        2 ready nodes
Got        5 ready nodes
Cluster nodes are Ready

Add secret for S3 bucket

secret/assets-s3-bucket created
Install Weave, CNI-Genie, s3bucket DaemonSet, Sidecar Daemonset...

clusterrole.rbac.authorization.k8s.io/genie-plugin created
clusterrolebinding.rbac.authorization.k8s.io/genie-plugin created
serviceaccount/genie-plugin created
configmap/genie-config created
daemonset.apps/genie-plugin created
daemonset.apps/genie-network-admission-controller created
service/genie-network-admission-controller created
serviceaccount/weave-net created
clusterrole.rbac.authorization.k8s.io/weave-net created
clusterrolebinding.rbac.authorization.k8s.io/weave-net created
role.rbac.authorization.k8s.io/weave-net created
rolebinding.rbac.authorization.k8s.io/weave-net created
daemonset.apps/weave-net created
daemonset.apps/s3bucket created
daemonset.apps/testground-sidecar created
Install Redis...

Error: could not get apiVersions from Kubernetes: unable to retrieve the complete list of server APIs: metrics.k8s.io/v1beta1: the server is currently unable to handle the request
```

We might need to add another wait for the Kubernetes APIs to be ready, before we run the `Install Redis` step. I am not sure if this was always racy, or now the metrics add-ons are changing the behaviour.

infra/k8s/cluster.yaml.example_with_monitoring
```diff
@@ -0,0 +1,108 @@
...
+kind: Cluster
+metadata:
+  creationTimestamp: null
+  name: my-first-cluster-kops.k8s.local
+spec:
+  addons:
+  - manifest: metrics-server
+  - manifest: prometheus-operator
+  - manifest: kubernetes-dashboard
+  api:
```
nonsense: on my end, `kops` generates:

```
  api:
    loadBalancer:
      type: Public
```

Why are we changing it to `dns: {}` and what does that mean?

infra/k8s/cluster.yaml.example_with_monitoring
```diff
@@ -0,0 +1,108 @@
...
+  - cpuRequest: 100m
+    etcdMembers:
+    - instanceGroup: master-eu-central-1a
+      name: a
+    memoryRequest: 100Mi
+    name: events
+  iam:
+    allowContainerRegistry: true
+    legacy: false
+  kubeAPIServer:
```
nonsense: This seems independent from the `metrics` part - it'd be nice to have a bit more documentation on the changes to the default `kops` cluster manifest.

infra/k8s/cluster.yaml.example_with_monitoring
```diff
@@ -0,0 +1,108 @@
...
+    runtimeConfig:
+      autoscaling/v2beta1: "true"
+  kubeControllerManager:
+    horizontalPodAutoscalerUseRestClients: true
```
nonsense: This seems independent from the `metrics` part as well.

nonsense:
@coryschwartz there are many changes done to the default `cluster.yaml` - can we split them (`monitoring` and `auto-scaling`) and add docs on what we are modifying.

infra/k8s/cluster.yaml.example_with_monitoring
```diff
@@ -0,0 +1,108 @@
...
+    anonymousAuth: false
+    authenticationTokenWebhook: true
+    authorizationMode: Webhook
+    enableCustomMetrics: true
+    maxPods: 200
+    allowedUnsafeSysctls:
+      - net.core.somaxconn
+  kubernetesApiAccess:
+  - 0.0.0.0/0
+  kubernetesVersion: 1.15.7
```
nonsense: We've been using the following so far:
```
  kubernetesVersion: 1.17.0
```

coryschwartz:
@nonsense Hopefully these last commits address the comments you gave.

I don't want to include it in this PR because I think it muddies it it up even more, but I might also want to turn the weave setup into a kubernetes addon as well.

infra/k8s/cluster.yaml.example_with_monitoring
nonsense: Looks like you reverted this, so I am resolving this conversation.

nonsense:
@coryschwartz the master node doesn't come up for me with this spec, and I end up at the:
`Got        0 ready nodes` stage.

The Kubernetes version seems to still be `1.15.7` when `kops` supports `1.17.2` - is there a reason we are rolling this back?

Which version of `kops` are you using locally? On my end I have `Version 1.17.0-alpha.1 (git-501baf7e5)`.

Have you tried this setup with other instance type, but `m3.medium` and `t2.medium` ?

coryschwartz:
I'm on kops 1.15, I think that explains the kubernetes version difference. Strange, what prevents the master node from coming up? 

I currently just have the machineType hard-coded, not templated, but I can template the machine type as well.

nonsense:
@coryschwartz I remember we had issues with earlier versions of `kops`, maybe while installing Weave. I remember @jimpick ran into a problem with `kops` and then noticed that the rest of us use the `v1.17.0-alpha` version, and this fixed it for him.

I suggest we use the same version, because there are already other people on the team that use this (Adin, Dirk, Henrique, etc.) so we should not spend time debugging various issues with backward-incompatibility with `kops`.

I don't see a good reason for us to roll back k8s version from 1.17 to 1.15 (knowing that we had issues with earlier versions), so could you please install a newer version of `kops` and try to get the metrics addons work with the newer versions, that we are already using? If this creates any problems, we can decide how to move on.

In the `https://github.com/ipfs/testground/tree/master/infra/k8s` README we currently specify `kops. >= 1.17.0-alpha.1` - we could restrict this even further if necessary, seeing that `1.17.0-alpha.4` is now out.

nonsense:
Great stuff - works with the current `yaml` - I will try with the latest version of kops `1.17.2` later!

nonsense:
I see we have `dashboard` pod started - we should document how to access it.

nonsense:
Replacing Kubernetes version with 1.17.2 results in:
```
Error: could not get apiVersions from Kubernetes: unable to retrieve the complete list of server APIs: metrics.k8s.io/v1beta1: the server is currently unable to handle the request
```

---

This worked with 1.15 for me, so either a transient error, or something changed between versions.

coryschwartz:
I think to change the kubernetes version you'll need to change the images used also --that's probably what's happening when you try to upgrade to 1.17.2

coryschwartz:
I'll update the doc to explain how access the dash.

nonsense:
> I think to change the kubernetes version you'll need to change the images used also --that's probably what's happening when you try to upgrade to 1.17.2

Actually kops had generated images referencing 1.15, even when I had specified 1.17.2 for the Kubernetes version, so I am not sure about that.

infra/k8s/README.md
```diff
@@ -169,6 +169,24 @@ kubectl get pods -o wide
 kubectl logs <pod-id, e.g. tg-dht-c95b5>
 ```
 
+5. Check on the monitoring infrastructure (it runs in the monitoring namespace)
+```
+kubectl get pods --namespace monitoring
+```
+
+6. Get access to teh redis shell
```
nonsense: typo

nonsense:
@coryschwartz if this is ready, could you please rebase to master, and we can try it tomorrow and merge?

PR #540, opened 2020-02-12
--collect and --collect-file flags

nonsense:
Fixes: https://github.com/ipfs/testground/issues/520

---

Cross posting from another PR:

It currently takes almost 2 min to download assets with `--collect` for run `9ccc44f28b15` (~2k small files), whereas downloading with `aws s3 cp --recursive` takes 10 seconds.

Maybe we should refactor this to first download and then zip, rather than be forced to download sequentially. WDYT ?

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -38,9 +38,13 @@ var RunCommand = cli.Command{
 					Name:  "ignore-artifacts, i",
 					Usage: "Ignores any build artifacts present in the composition file.",
 				},
-				cli.StringFlag{
-					Name:  "collect-into, o",
+				cli.BoolFlag{
+					Name:  "collect",
 					Usage: "Collect assets at the end of the run phase.",
+				},
+				cli.StringFlag{
+					Name:  "collect-file, o",
+					Usage: "Destination for the assets if --collect is set",
 				},
 			},
 		},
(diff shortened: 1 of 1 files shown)


nonsense:
@raulk it seems not possible to have one flag be used as a bool and as a string, or as a string that could be missing (`--collect` and `--collect filename.zip`)... I couldn't find any documents on support for this, so I added two flags, resulting in:

```
$ testground run composition -f composition.toml --collect
>> collects into <run_id>.zip
$ testground run composition -f composition.toml --collect -o foo.zip
>> collects into foo.zip
$ testground run composition -f composition.toml --collect --collect-file foo.zip
>> collects into foo.zip
```

raulk:
@nonsense no need to do that. You can check if a flag is set or not with: https://pkg.go.dev/gopkg.in/urfave/cli.v1?tab=doc#Context.IsSet.

* If the flag is set, but the value is "", we'd save as `<run_id>.zip`.
* If the flag is set, but the value is != "", we'd save as whatever the user has specified.
* If the flag is not set, we don't collect.

nonsense:
> @nonsense no need to do that. You can check if a flag is set or not with: https://pkg.go.dev/gopkg.in/urfave/cli.v1?tab=doc#Context.IsSet.
> 
> * If the flag is set, but the value is "", we'd save as `<run_id>.zip`.
> * If the flag is set, but the value is != "", we'd save as whatever the user has specified.
> * If the flag is not set, we don't collect.

I think that would make the command look like:
```
testground run composition -f composition.toml --collect=""
```

rather than:

```
testground run composition -f composition.toml --collect
```

---

The following command
```
testground run composition -f composition.toml --collect
```
complains with `flag needs an argument: -collect`

raulk:
@nonsense :-( Ok, let's keep it this way, then. Thanks!

PR #545, opened 2020-02-13
feat: sdk runtime metrics

### LINKED ISSUE #372: Self prometheus scraping and snapshotting (was: sdk: convenience functions to emit go runtime metrics)
This issue originally had to do with capturing snapshots of go runtime metrics and saving them as raw output assets.

In the process, we realised that Prometheus already instruments go runtimes and exposes metrics, automatically. Additionally, various libraries/projects that Testground caters to use Prometheus to export metrics.

Therefore, there's an opportunity to create a lot more value if we implement this correctly.

- [ ] In the runtime sdk, create a function that initialises a default Prometheus http handler: https://prometheus.io/docs/guides/go-application/, exposing it on some local port, and returning that address.
- [ ] In the runtime sdk, create a function that polls a provided HTTP address every N `time.Duration` (argument), and writes the output of each iteration as a raw asset (`runenv.CreateRawAsset()`) following some user-provided filename pattern.

Voilà. Folks wanting to export Go runtime metrics (and all Prometheus metrics), can just call:

```go
addr := runtime.MustExportPrometheus()
go runtime.HTTPPeriodicSnapshots(addr, 5 * time.Second, "metrics.out")
```

--- 

Note: we may run into problems with the way that automatic asset closure has been implemented (we can only track 32 assets per type). This needs to be rethought, as in this case we'll create an asset, write to it, and close it immediately (it won't be a long-lived asset).

hacdias:
This closes #372 and it is divided into two main commits:

1. f9b05aa8703458405d9fe45956ad00c27e0c1720: adds the actual code for collecting runtime metrics using Prometheus.
	1. `runenv.MustExportPrometheus` which returns an address to the entrypoint where the metrics can be collected. This function panics in case of error and starts on a random free port.
	2. `runenv.HTTPPeriodicSnapshots` which periodically checks for new metrics. It accepts an address, a duration and the output file. The output file might include a `$TIME` placeholder which will be replaced by a unix timestamp. If not, the timestamp will be appended to the filename.
2. 0eab06197052a1474b8590c04c1aa22c47baf26c: adds a small example.
	1. `./testground run single placebo/metrics -i 1 -b exec:go -r local:exec`
	2. `./testground collect -r local:exec RUN_ID`

**Questions:**

1. Since I'm adding this to `runenv`, couldn't I just store the address on `RunEnv` and avoid passing it to `HTTPPeriodicSnapshots`?


hacdias:
I'm hitting the `too many unstructured assets; current: 32` error. Is there a way to stop keeping track of them?

sdk/runtime/metrics.go
```diff
@@ -0,0 +1,61 @@
...
+			re.RecordMessage("error while getting prometheus stats: %v", err)
+			continue
+		}
+
+		file, err := nextFile()
+		if err != nil {
+			re.RecordMessage("error while getting metrics output file: %v", err)
+			continue
+		}
+		defer file.Close()
```
nonsense: You probably want to call `file.Close()` within the loop, so maybe add a closure in the loop.

I am also not seeing any call to `resp.Body.Close()` I think we are supposed to close it after copying to `file`.

sdk/runtime/metrics.go
```diff
@@ -0,0 +1,61 @@
...
+// It starts on a random open port and returns the http address.
+func (re *RunEnv) MustExportPrometheus() string {
+	listener, err := net.Listen("tcp", ":0")
+	if err != nil {
+		panic(err)
+	}
+
+	go http.Serve(listener, promhttp.Handler())
+
+	port := strconv.Itoa(listener.Addr().(*net.TCPAddr).Port)
```
nonsense: You could probably just return the listener, or the listener.Addr().String() and not have to do any conversions.

nonsense: Probably best to return the listener, so that we can close it, when we are done with exposing metrics on shutdown?

hacdias: Well, we want the HTTP address, don't we? We'll need to convert it anyhow to use with `http.Get`.

sdk/runtime/metrics.go
hacdias: Thanks for noticing that. Done! However, this still doesn't fix what I said in the [comment](https://github.com/ipfs/testground/pull/545#issuecomment-585658242). Was thinking of a more manual approach.

hacdias:
After talking to @nonsense, we decided to make a few changes:

- `MustExportPrometheus` now returns a listener that the caller can close and that can be used to get the address.
- `HTTPPeriodicSnapshots` takes a context that can be used to stop the goroutine that makes the periodic calls.

sdk/runtime/metrics.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	go http.Serve(listener, promhttp.Handler())
+	return listener
+}
+
+// HTTPPeriodicSnapshots periodically fetches the snapshots from the given address
+// and outputs them to the out directory. Every file will be in the format timestamp.out.
+func (re *RunEnv) HTTPPeriodicSnapshots(ctx context.Context, addr string, dur time.Duration, outDir string) error {
+	if !strings.HasPrefix(addr, "http") {
```
nonsense: I think it would be better to have a strict contract here.

Expect a proper endpoint in this function, and just pass http:// + listener.Addr().String() to it. In the future we might pass an https etc.

hacdias: Done! I agree

plans/placebo/main.go
```diff
@@ -14,8 +16,27 @@ func run(runenv *runtime.RunEnv) error {
...
+		// expose prometheus endpoint
+		listener := runenv.MustExportPrometheus()
+		defer listener.Close()
+
+		// create context for cancelation
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		// snapshot metrics every second and save them into "metrics" directory
+		err := runenv.HTTPPeriodicSnapshots(ctx, listener.Addr().String(), time.Second, "metrics")
```
raulk: I'd imagine that most users will want these snapshots to be taken as long as the process is running (i.e. the lifetime of `context.Background()`), so in most cases folks can just pass in `context.Background()` here, but I like that the control is on the user, i.e. if we want to take snapshots during a particular stage of the test only, we can do so by using a cancellable context 👍

plans/placebo/main.go
```diff
@@ -14,8 +16,27 @@ func run(runenv *runtime.RunEnv) error {
...
+		if err != nil {
+			return err
+		}
+
+		time.Sleep(time.Minute)
```
raulk: Do we want the placebo test to last so long? Won't it slow down integration tests? Just 10 seconds should be enough to validate it's working?

sdk/runtime/go.sum
```diff
@@ -1,12 +1,77 @@
+github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc/go.mod h1:LOuyumcjzFXgccqObfd/Ljyb9UuFJ6TxHnclSeseNhc=
```
raulk: Ooooof, that's a lot of transitive dependencies. I thought prometheus would be lighter. Just to make sure we're not leaking anything unnecessary here, could you run `go mod tidy` once more?

sdk/runtime/metrics.go
raulk: In the future we could make this function more generic by taking an `*http.Request`, which can be reused safely (apparently stores no state: https://github.com/golang/go/issues/19653). Then this func would make no assumptions about the HTTP request method, headers, body, etc. For now, we're fine like this 👍

sdk/runtime/metrics.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				func() {
+					resp, err := http.Get(addr)
```
raulk: Let's use `NewRequestWithContext` instead. Not a biggie, but this http request is not in the scope of the passed-in context, so it won't yield if the context gets cancelled.

sdk/runtime/metrics.go
```diff
@@ -0,0 +1,77 @@
...
+					if err != nil {
+						re.RecordMessage("error while getting prometheus stats: %v", err)
```
raulk: This function knows nothing about Prometheus.

```suggestion
						re.RecordMessage("error while scraping http endpoint: %v", err)
```

sdk/runtime/metrics.go
```diff
@@ -0,0 +1,77 @@
...
+						return
+					}
+					defer resp.Body.Close()
```
raulk: All of these defers are going to accumulate until the goroutine exits, which I don't think is what you want. We need to call Close immediately.

sdk/runtime/metrics.go
```diff
@@ -0,0 +1,77 @@
...
+
+					file, err := nextFile()
+					if err != nil {
+						re.RecordMessage("error while getting metrics output file: %v", err)
+						return
+					}
+					defer file.Close()
```
raulk: Same as above.

sdk/runtime/metrics.go
```diff
@@ -0,0 +1,77 @@
...
+
+					io.Copy(file, resp.Body)
+				}()
+			}
+
+			time.Sleep(dur)
```
raulk: Sleeping won't allow the loop to be cancelled if the context fires. You probably want to use a `time.NewTicker()` and select from the context and from `ticker.C`.

sdk/runtime/go.sum
hacdias: I ran it about 3 times already.

plans/placebo/main.go
hacdias: This is a specific placebo test: `placebo/metrics` just to try out the metrics. It's not being run on the integration tests AFAIK.

sdk/runtime/metrics.go
hacdias: It's inside of a closure so no problem.

sdk/runtime/metrics.go
hacdias: It's inside of a closure so no problem.

sdk/runtime/metrics.go
hacdias: Done! Didn't know that existed so thanks :D

sdk/runtime/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+		return err
+	}
+
+	nextFile := func() (*os.File, error) {
+		timestamp := strconv.FormatInt(time.Now().Unix(), 10)
+		return os.Create(path.Join(re.TestOutputsPath, outDir, timestamp+".out"))
+	}
+
+	go func() {
+		ticker := time.NewTicker(dur)
```
nonsense: I think we need a `defer ticker.Stop()` so that this properly frees resources when the go-routine exits, once the context is canceled.

hacdias:
I'm merging this since all the feedback has been addressed in the meanwhile and everything is passing now.

PR #551, opened 2020-02-13
use shared ssh key for kops hosts ; command to extract kube context

nonsense:
Fixes: https://github.com/ipfs/testground/issues/550

---

With this, if we need to log to another cluster, we just have to update the context for the local kubectl client with:
```
kops export kubecfg --state $KOPS_STATE_STORE --name=$NAME
```

And then we already have the SSH keys to log to the hosts.

### DIFF
--- a/infra/k8s/README.md
+++ b/infra/k8s/README.md
@@ -56,6 +56,7 @@ aws s3api create-bucket \
 - set `kops` state store bucket
 - set number of worker nodes
 - set location for cluster spec to be generated
+- set location of your SSH public key
 - set credentials and locations for `outputs` S3 bucket
 
 You might want to add them to your `rc` file (`.zshrc`, `.bashrc`, etc.)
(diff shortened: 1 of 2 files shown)


raulk:
I defer to @coryschwartz!

infra/k8s/README.md
```diff
@@ -59,7 +73,7 @@ export ZONES=eu-central-1a
 export KOPS_STATE_STORE=s3://kops-backend-bucket
 export WORKER_NODES=4
 export CLUSTER_SPEC=~/cluster.yaml
-export PUBKEY=~/.ssh/id_rsa.pub
+export PUBKEY=~/.ssh/testground_rsa.pub
```
nonsense: I am still keeping this option, because members from the community should be able to use Testground as well, and they don't want to be getting our shared key.

Issue #547, opened 2020-02-13
Redis Writer Panics


aschmahmann:
The sync.Writer panics if it fails to submit a keepAlive message. I've seen three different types of occurrences of this:

https://github.com/ipfs/testground/blob/a1a6d48560375d7f3b9570fdd8365e0017450c00/sdk/sync/write.go#L91-L97

1) Context cancellation

I suspect the issue is that we cancel before taking the lock here:

https://github.com/ipfs/testground/blob/a1a6d48560375d7f3b9570fdd8365e0017450c00/sdk/sync/write.go#L195-L200

2) Something like a redis lookup event

```
{"ts":1581578960316135616,"msg":"","group_id":"cypress","run_id":"8602414488a5","event":{"type":"message","message":"start provide loop"}}       panic: dial tcp: lookup redis-headless on 100.64.0.10:53: write udp 100.96.7.79:48445->100.64.0.10:53: write: invalid argument
goroutine 19 [running]:
github.com/ipfs/testground/sdk/sync.(*Writer).keepAlive(0xc0002920f0, 0xe558c0, 0xc0003440c0)
        /sdk/sync/write.go:95 +0x1d4
github.com/ipfs/testground/sdk/sync.(*Writer).keepAliveWorker(0xc0002920f0, 0xe558c0, 0xc0003440c0)
        /sdk/sync/write.go:79 +0x5c
created by github.com/ipfs/testground/sdk/sync.NewWriter
        /sdk/sync/write.go:69 +0x279
```

3) An I/O timeout

{"ts":1581624908918749128,"msg":"","group_id":"cypress","run_id":"9bd780077d08","event":{"type":"message","message":"keep alive err: attempt: 1, err: dial tcp: i/o timeout"}}

This was collected by a wrapper the retries the writer when there are errors.

Stebalien:
We cancel before taking the lock so we can cancel an in-progress event. We definitely need to check to see if the context has expired before panicing.

However, the message is odd:

> panic: dial tcp: lookup redis-headless on 100.64.0.10:53: write udp 100.96.7.79:48445->100.64.0.10:53: write: invalid argument

I guess we could get an "invalid argument" error if the socket is closed? I'm concerned it could be some other DNS issue.

raulk:
@Stebalien It _is_ another DNS issue. The target port in that panic is `udp/53`.

Stebalien:
@raulk it could also be a canceled context causing us to abort/close something.

[...]

PR #554, opened 2020-02-13
fix(sdk/sync): don't panic on error if we're canceling anyways

Stebalien:
related to #547

### DIFF
--- a/sdk/sync/write.go
+++ b/sdk/sync/write.go
@@ -76,7 +76,12 @@ func (w *Writer) keepAliveWorker(ctx context.Context) {
 	for {
 		select {
 		case <-time.After(KeepAlivePeriod):
-			w.keepAlive(ctx)
+			if err := w.keepAlive(ctx); err != nil {
+				if ctx.Err() != nil {
+					return
+				}
+				panic(err)
+			}
 		case <-ctx.Done():
 			return
 		}
@@ -84,7 +89,7 @@ func (w *Writer) keepAliveWorker(ctx context.Context) {
 }
 
 // keepAlive extends the TTL of all keys in the keepAliveSet.
-func (w *Writer) keepAlive(ctx context.Context) {
+func (w *Writer) keepAlive(ctx context.Context) error {
 	w.lk.RLock()
 	defer w.lk.RUnlock()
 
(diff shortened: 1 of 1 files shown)


Stebalien:
What other panics?

aschmahmann:
#547 lists 3 types of errors:
1) Context cancellation (covered by this PR)
2) Redis lookup (DNS?) event
3) I/O timeouts

Issues 2 + 3 are failures happening in the redis client that probably shouldn't happen which is why we're panicking. However, since we are currently panicking the tests need some protection in the meanwhile (e.g. retries).

Stebalien:
The second issue could easily be context cancellation as well. Do you have exact error messages for all of these?

sdk/sync/write.go
```diff
@@ -76,25 +76,31 @@ func (w *Writer) keepAliveWorker(ctx context.Context) {
 	for {
 		select {
 		case <-time.After(KeepAlivePeriod):
-			w.keepAlive(ctx)
+			if err := w.keepAlive(ctx); err != nil {
+				if ctx.Err() != nil {
```
raulk: Could possibly use:

```go
if errors.Is(err, context.Canceled)
```

aschmahmann:
Only the second one, the third one is rarer and I forgot to record the error. What type of error message for the second one are you looking for? Is it more than what's reported in the other issue (copied below).

{"ts":1581618676264268161,"msg":"","group_id":"cypress","run_id":"5e5397620cd1","event":{"type":"finish","outcome":"failed","error":"dial tcp: lookup redis-headless on 100.64.0.10:53: write udp 100.96.12.27:59536->100.64.0.10:53: write: invalid argument"}}

Is it normal/reasonable that the send port will change in between requests? i.e. 100.96.12.27:53060 for one request and 100.96.12.27:59536 for another.

Stebalien:
So, the errors are:

* Context canceled
* "dial tcp: lookup redis-headless on 100.64.0.10:53: write udp 100.96.12.27:59536->100.64.0.10:53: write: invalid argument"
* Some IO timeout error.

Stebalien:
> Is it normal/reasonable that the send port will change in between requests? i.e. 100.96.12.27:53060 for one request and 100.96.12.27:59536 for another.

For DNS, yes.

sdk/sync/write.go
Stebalien: So... I don't trust errors. I'm seeing an "invalid argument" error which could, maybe, be caused by:

```go
go func() {
  <-ctx.Done()
  myConn.Close()
}()
return myConn.Read()
```

(depending on the connection type, etc.)

PR #555, opened 2020-02-13
feat: resolve redis once

Stebalien:
Instead of hammering the DNS server, resolve redis once when we first connect.

Related to #547.


Stebalien:
Tested.

raulk:
@Stebalien ❤️ the test might have worked because you had a local redis running? To make it failsafe we’d need to call `ensureRedis` (I think; unless I missed something — reviewing on my phone).

Issue #432, opened 2020-01-29
Runner healthcheck + repair + terminate


raulk:
We want to introduce the following commands:

* `testground healthcheck [--repair] [--runner cluster:k8s]`: checks, and optionally heals, the preconditions for the runner to be able to run properly, e.g. redis, sidecar, etc.
* `testground terminate [--runner cluster:k8s]`: terminates all jobs/containers running on that runner.

Rationale:

* runners have preconditions that need to be met for tests to run successfully, e.g. redis processes/containers, sidecar, network rules, etc.
* We are moving fast, and these moving pieces will inevitably break, deadlock, spiral out of control (CPU/memory).
* We want to avoid testground developers having to deal with the internals and nitty-gritty details of the k8s setup. If things fail, they should be able to heal the cluster/setup with a single command, in an automated fashion.

Design notes for healthcheck/repair:

* A single-method interface `Healthcheckable`: `Check(repair bool) error` that runners capable of healthcheck implement.
* On starting the daemon, we should call `Healthcheck.Healthcheck(true)` on every runner.
* On performing a run, we should call `Healthcheck.Healthcheck(false)` on the runner.
* If healthchecks for a runner fail, calling `testground healthcheck --repair --runner <runner>` should attempt to fix the setup, and fail if it wasn't able to.
* Some pathological situations are not detectable via a healthcheck (e.g. sidecar deadlocks, which we know we have). Those cases will be spotted manually (i.e. the run will fail), but the user should be able to call `testground repair` to heal the situation.

Design notes for terminate:

* A single-method interface `Terminatable`: `TerminateAll() error` that runners supporting termination can implement.
* Calling such method will kill all instances (containers, processes, etc.) started by that runner.

raulk:
@nonsense does this capture our conversation accurately?

nonsense:
@raulk yes, sounds right to me. I would just highlight **On starting the daemon, we should call Diagnoseable.Healthcheck() on every runner.** because we don't want users to have to run yet more commands - `testground healthcheck --runner RUNNER` could be there, but users don't have to run it - it will be run automatically when starting the `testground daemon` and also when scheduling a testplan with `testground run`.

hacdias:
Will start working on this today or tomorrow morning. Will let you know if I need any help or insights.

PR #557, opened 2020-02-14
add healthcheck interfaces + support in local:{exec,docker}

hacdias:
This is related to #432 and adds the `healthcheck` command and the `Healthcheckable` interface.

I decided to do this on smaller PRs.

On this PR:
- A single-method interface Healthcheckable: Check(repair bool) error that runners capable of healthcheck implement.
- ~On starting the daemon, we should call Healthcheck.Healthcheck(true) on every runner.~ (see [comment](https://github.com/ipfs/testground/pull/557#discussion_r379360687)) 
- On performing a run, we should call Healthcheck.Healthcheck(false) on the runner.

What's missing to do on separate PRs:
- Implement `Healthcheck(repair bool) error` for each runner.

License: MIT


pkg/api/runner.go
```diff
@@ -94,3 +94,9 @@ type CollectionInput struct {
 	// plan manifest, coalesced with any user-provided overrides.
 	RunnerConfig interface{}
 }
+
+// Healthcheckable is the interface to be implemented by a runner that can be
+// healthcheckable.
+type Healthcheckable interface {
```
nonsense: nitpick, but in Go, interfaces that have a method `Do`, are usually called `Doer`, so this should probably be named `Healthchecker`, so that when you see the interface name, you know there is a `Healthcheck` method :)

hacdias: I think that's a great nitpick though haha. Gonna change.

pkg/daemon/daemon.go
```diff
@@ -34,6 +36,12 @@ func New(listenAddr string) (srv *Daemon, err error) {
 		return nil, err
 	}
 
+	// Check all runners' health.
```
nonsense: Not sure about that, because we might not have all runners setup for the given environment. When you run testground you generally must have >0 runners working correctly, but not necessarily all of the runners we support.

For example on my local machine I might not have the `cluster:k8s` runner setup.
A remove testground daemon might not have the `local:exec` runner working correctly.

hacdias: I was just following what was decided on #432 for this. I can just remove that part so we can merge for now and decide that later.

nonsense: Is `ListRunners` returning all `testground supported` runners, or all `supposedly configured testground runners in this environment` ?

hacdias: All runners from `EngineConfig` so in this case it's `AllRunners` because we get the default engine.

hacdias: Just removed that bit. Can be done on a separate pr.

nonsense:
> Changing review status to "request changes" for my comment to be considered before merging. Also, we shouldn't merge this before actually implementing the interface once or twice -- that's optimistic!

I totally agree with this - I personally find it difficult to reason about correctness of interfaces without having them implemented at least twice. I prefer bigger PRs that provide the MVP of a functionality, than chunking down PRs just for the sake of them being smaller (this is not always possible, but in this case, implementing the interface once is totally doable :)

raulk:
Based on discussion, here's an alternative data model that revolves around "checks and fixes".

```
type HealthcheckStatus string // a string enum

type HealthcheckItem struct {
    Name       string
    Status     HealthcheckStatus
    Message    string
}

type HealthcheckReport struct {
    Checks    []HealthcheckItem
    Fixes     []HealthcheckItem
}
```

hacdias:
For the runner `local:exec`, which kind of health checks should we make besides Redis?

raulk:
@hacdias that the output directory exists. Maybe walk through the runner code and pinpoint the environment preconditions we check before we actually do the run. Those pieces might be a little intertwined with the actual runner logic, that’s why I’m pumped we’re implementing this, because it will separate these concerns more clearly.

raulk:
@hacdias I took the liberty to take this to completion. Thanks for getting it started!

PR #558, opened 2020-02-14
expose pprof port on sidecar

nonsense:
TODO:
- [x] add a flag to sidecar so that it actually starts pprof on 6060

We could template this, so that a dev can enable/disable pprof, but for now we can assume that `sidecar.yaml` is setup correctly (enabled or disabled) and be done with it - I don't want to introduce another level of indirection.

---

Fixes: https://github.com/ipfs/testground/issues/423

### DIFF
--- a/cmd/sidecar.go
+++ b/cmd/sidecar.go
@@ -2,9 +2,12 @@ package cmd
 
 import (
 	"fmt"
+	"net/http"
+	_ "net/http/pprof"
 	"runtime"
 	"strings"
 
+	"github.com/ipfs/testground/pkg/logging"
 	"github.com/ipfs/testground/pkg/sidecar"
 	"github.com/urfave/cli"
 )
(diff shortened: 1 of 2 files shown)


nonsense:
Similar as pods, one could access this port by:

```
kubectl proxy
```

And then navigating to the http service, for example:

```
http://localhost:8001/api/v1/namespaces/default/pods/http:tg-dht-205d92ad7fb8-cypress-99:6060/proxy/debug/pprof/
```

Not sure if we need any more tooling around that (such as abstracting all this in `testground`).

WDYT?

nonsense:
@aschmahmann FYI in case we come across a sidecar that stops working again.

dirkmc:
😍

nonsense:
@dirkmc note that we are exposing 6060 on pods of testplan instances as well (https://github.com/ipfs/testground/blob/master/pkg/runner/cluster_k8s.go#L460) so if you add `pprof` to your testplans, you can access it the same way as described here (via the `kubectl proxy` and then the proxy scheme ; or port-forward directly to a given pod with `kubectl port-forward`)

dirkmc:
Awesome! Thanks Anton, this will be super useful for fuzz testing

PR #564, opened 2020-02-15
Add pushgateway

coryschwartz:
Install the prometheus pushgateway for short-lived tests
The values included are mostly from the stable helm chart,
but edited to take advantage of the prometheus operator.


coryschwartz:
this depends on  the prometheus operator, which is added in https://github.com/ipfs/testground/pull/534

nonsense:
LGTM. Let's merge after #534

PR #565, opened 2020-02-16
Proposal: dealing with upstream API evolution in test plans

raulk:
This design proposal:

* analyses the problem domain.
* studies various tactics and their applicability in the face of different kinds of upstream API evolution.
* proposes a concrete solution for both manual shim activation, and automatic shim activation based on version detection.

### DIFF
--- a/docs/EVOLVING_APIs.md
+++ b/docs/EVOLVING_APIs.md
@@ -1,0 +1,189 @@
+# Dealing with evolving APIs of components under test
+
+## Problem statement
+
+Testground makes it simple to compile test plans against arbitrary versions of
+upstream dependencies.
+
+This feature is powerful in that it serves as the cornerstone of comparative
+testing, enabling users to compare how version A of a specific module performs
+against versions B, C, and D. And even to orchestrate scenarios where all those
+versions are tested together within a single composition run.
+
+Power does not come devoid of challenges. Concretely, the main challenge here is
+sustainable and straightforward maintenance and evolution of the test plan and
+the components under test. None of those are stationary in time. They are bound
+to evolve both behaviourally and contractually (API changes).
+
+Ensuring that a test plan remains compatible and buildable against all relevant
+versions of upstream dependencies over time (or against a declared version range
+thereof) isn't something that will happen automatically, nor trivially.
+
+Testground needs to provide the plumbing for test plans to remain backwards and
+forwards-compatible with upstream dependencies, and this proposal introduces
+approaches to resolve this concern.
+
+## Constraints
+
+1. We want test plans to be _wildcard buildable_.
+     * We categorically want to avoid maintaining forks of test plans, where
+       each fork builds against a specific upstream dependency range.
+     * Such a scenario would rapidly turn into a maintenance chaos, and would
+       scale poorly when targeting multiple changing upstreams, turning into an
+       exponential hairball, as we'd have to maintain forks for the cartesian
+       product of all targeted upstreams.
+2. This problem applies only to test plans written in statically-typed languages
+   (e.g. Go).
+     * Dynamically-typed languages do not suffer from this problem, as they can
+       perform all kinds of tricks to do feature/API detection, and vary the
+       test plan's behaviour in accordance.
+3. How testground deals with this must refrain from using language-specific
+   assumptions and semantics in its core.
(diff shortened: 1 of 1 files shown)


docs/EVOLVING_APIs.md
```diff
@@ -0,0 +1,317 @@
...
+thereof) isn't something that will happen automatically, nor trivially.
+
+Testground needs to provide the plumbing for test plans to remain backwards and
+forwards-compatible with upstream dependencies, and this proposal introduces
+approaches to resolve this concern.
+
+## Constraints
+
+1. We want test plans to be _wildcard buildable_.
+     * We categorically want to avoid maintaining forks of test plans, where
```
nonsense: I am not convinced this solution is so bad. If you want to maintain all testplans until the end of time, then it is bad. If you build testplans `on-demand` to build confidence before release (as Testground is in `exploration/experimentation phase`), this is a simple solution.

On line 18 you say `all relevant versions` - would old versions, where APIs are no longer up-to-date stay `relevant` for months/years after a release? Would you want to run all testplan against them for a long period of time, after a release?

This about `go-ipfs 0.5.0` - once it is released, would we be running tests against older versions (even assuming we have meaningful testplans for all patch 0.4.x versions?)

I think the power of Testground would come from comparative releases - new vs old - not from keeping testplans relevant over time.

---

Discarding this option and opting for shims puts more cognitive load on developers of testplans.

docs/EVOLVING_APIs.md
```diff
@@ -0,0 +1,317 @@
...
+
+As a result, we do not consider this approach generally feasible by itself, in
+that drawbacks (2) and (3) actually make assumptions about available types.
+
+However, this approach _could_ be combined with shims to layer on conditional
+activation/selection based on upstream dependency versions.
+
+### Reflection
+
+Reflection in go is great, but insufficient by itselef in the advent of type
```
nonsense: typos - `Go` and `itself`.

docs/EVOLVING_APIs.md
raulk: > Consider go-ipfs 0.5.0 - once it is released, would we be running tests against older versions (even assuming we have meaningful testplans for all patch 0.4.x versions?)

Yes!

I'd like the ability to re-run a test against an arbitrary version of an upstream dependency at any time. This is important for many reasons:

* we will want to compose test scenarios with hybrid versions, to simulate real networks, where users don't upgrade in unison. deleting the test code for old versions would render this impossible. for example, the IPFS network still has nodes running v0.4.18 (November 2018).
* we won't be the only ones archiving testground results -- other actors will want to run our tests to produce their own observations / data points, to confirm or reject ours. Allowing others to reproduce our results is very important.
* Our infrastructure will vary over time, in ways that it'll affect the baseline numbers. We will want to rerun tests.
* Go versions change and introduce performance/footprint improvements and regressions. We will want to rerun tests.
* We cannot assume that our historical database will be perennial. It can be destroyed, and if we have deleted the code that generated those results, it'll all be a mess. Yes, it'll probably stay buried somewhere in version control, but relying on that is a mistake IMO.

The mechanisms introduced in this document allow us to reason only once about how to test the API of a particular version.

We then proceed to encode this information in actual code, effectively building a database of shims over time that allow us to exercise any historical version of the software under test.

docs/EVOLVING_APIs.md
raulk: Thanks!

raulk:
```suggestion
Reflection in Go is great, but insufficient by itself in the advent of type
```

raulk:
Thanks for the review, @nonsense!

raulk:
Manual shim selection is implemented in #566.

docs/EVOLVING_APIs.md
nonsense: I'm just saying that doing this has a price, and it is a bit hard to say right now what kind of testing will provide the most value vs effort/cost. Furthermore we will have to support multiple languages, so you will have to write a given testplan for a Go impl, Rust impl and JS impl. This is essentially the same thing as writing a given testplan for Go impl ver A, and Go impl ver B, however you suggest we introduce a different construct with `shims`.

PR #567, opened 2020-02-17
Add a state machine

coryschwartz:
The purpose for this is to abstract away synchronization methods and allow modeling for patterns, such as plan factories.

### DIFF
--- a/manifests/example.toml
+++ b/manifests/example.toml
@@ -51,3 +51,7 @@ instances = { min = 1, max = 200, default = 1 }
 [[testcases]]
 name = "sync"
 instances = { min = 2, max = 200, default = 5 }
+
+[[testcases]]
+name = "statemachine"
+instances = { min = 1, max = 200, default = 5 }
--- a/plans/example/main.go
+++ b/plans/example/main.go
@@ -25,6 +25,8 @@ func run(runenv *runtime.RunEnv) error {
 		return ExampleParams(runenv)
 	case "sync":
 		return ExampleSync(runenv)
+	case "statemachine":
+		return ExampleStateMachine(runenv)
 	default:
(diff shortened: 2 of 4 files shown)


Stebalien:
Can we get an example of this in practice?

coryschwartz:
@Stebalien I added an example for you.

I admit it's a bit hard to read, but the idea I'm shooting for is to be able to abstract away some of the details, like barriers between phases of a test.

It seems to me that many of the tests are written in distinct phases, acts, whatever. The example would allow multiple acts to be performed with a barrier between them. The whole test plan is constructed using a factory. 

The factory, in my mind, could be managed by in the SDK and used by plans. We would add factories for common patterns, barrier configurations, etc.


It might be easier just to copy frequently used methods into a 'utils' package. I could imagine just copying methods from the bitswap-tuning and others that plans that have made good utils.

Stebalien:
I guess, at this point, I'm not seeing how this could be used to simplify the tests. It's probably more flexible (and simpler) to just write:

```go
var err error
for i, stage := []func(...){
  Stage1, // defined elsewhere.
  Stage2,
  Stage3,
} {
    if err == nil { // we still want to hit the barriers if we hit a failure.
        err = stage(...)
    }
    BarrierHelper(fmt.Sprintf("barrier-%d", i))
}
teardown(...)
return err
```

We could even put this in a function if it becomes a pattern.

coryschwartz:
True, and not merging.

I was going for re-usable patterns similar to operators in apache airflow, but I have to agree with you I'm not happy with this change. Closing.

Robmat05:
ok, this ticket has been abandoned and will be closed and tagged as ‘Rotten’

PR #568, opened 2020-02-17
Add network utils

coryschwartz:
This is largely taken from the bitswap tuning, and made a little more general

Does it make sense to come up with some common/well-known toml-configuration params?


sdk/utils/network.go
```diff
@@ -0,0 +1,68 @@
+package utils
```
nonsense: nitpick: maybe we could name this package `network`, with methods called `Setup` and `SetupWaitAll`, then we avoid having a generic name for a package.

Stebalien: Agree. If we do that, we should move all network helpers here (e.g., `sync.WaitNetworkInitialized`).

coryschwartz: Sounds good, I will do that.

sdk/sync/helper.go
```diff
@@ -7,6 +7,8 @@ import (
 	"github.com/ipfs/testground/sdk/runtime"
 )
 
+// Depreciated. Use sdk.network.WaitNetworkInitialized
```
Stebalien: This is going to cause the lints to fail. We should do a mass update.

Stebalien:
LGTM once tests pass.

coryschwartz:
I'm closing this for now. I think there will be a more comprehensive plan SDK review in the future.

Robmat05:
Ok, cool. We will close this as ‘Discarded’ and look for it to be linked to a new ticket on SDK review, once it is opened.

PR #573, opened 2020-02-19
Start prometheus and pushgatway on local docker

coryschwartz:
For feature parity with the the k8s runner.

Prometheus and pushgateway start and attach to the control network. 

Right now I haven't considered how to make it resolvable, but we should make it the same as the k8s runner as much as possible.


pkg/runner/local_docker.go
```diff
@@ -337,7 +349,7 @@ func ensureControlNetwork(ctx context.Context, cli *client.Client, log *zap.Suga
 		ctx,
 		log, cli,
 		"testground-control",
-		true,
+		false,
```
coryschwartz: Even with PublishAllPorts set true, ports are not exposed through the NAT when the network is internal. I switched this to enable us to view these containers in our local web browser.

pkg/runner/local_docker.go
```diff
@@ -383,6 +395,46 @@ func newDataNetwork(ctx context.Context, cli *client.Client, log *zap.SugaredLog
...
 
+func ensurePrometheusContainer(ctx context.Context, cli *client.Client, log *zap.SugaredLogger, controlNetworkID string) (id string, err error) {
+	container, _, err := docker.EnsureContainer(ctx, log, cli, &docker.EnsureContainerOpts{
+		ContainerName: "prometheus",
+		ContainerConfig: &container.Config{
+			Image: "prom/prometheus",
+		},
+		HostConfig: &container.HostConfig{
+			NetworkMode:     container.NetworkMode(controlNetworkID),
+			PublishAllPorts: true,
```
coryschwartz: This exposes an ephemeral port accessible by the host.

coryschwartz:
I would say this *works* but the experience is less than ideal.

Unlike the prometheus operator on kubernetes, which can perform auto-discovery for nodes to scrape, the prometheus container running by itself, such as basic docker incantation, does not discover the pushgateway. I am getting around this by bind mounting the config into the container, but really what should happen is just replace the config by adding another layer. I think I'm going to back out the bind mount change and instead just build another local container with the Makefile so it's properly configured. 


Don't approve this yet, is what I'm saying.

coryschwartz:
For the kubernetes prometheus operator, the configuration is not required since it can learn about scrape endpoints from the ServiceMonitor CRD. Of course, we do not have such concepts available in a local docker environment. I'm going to close this one and open a nearly-identical PR so I don't have to deal with the merge conflicts that have built up since this was initially opened, but I'll continue with your suggestions there. Thanks @raulk.

Robmat05:
Thanks @coryschwartz, I’ve tagged this as Discarded. So when you open the new ticket, please remember to link to this one so that we can track the progress from when the issue was originally targeted for dev.

Good job recognizing not to bang your head against the wall!

PR #574, opened 2020-02-20
feat(sync): use a shared redis client

Stebalien:
The redis client is designed to be shared between many goroutines. We were creating separate redis instances everywhere. This meant:

1. Each test instance created two redis clients instead of one (write & watch).
2. The sidecar created two redis clients (and multiple connections) per test instance.

Worse, we weren't _closing_ writers. That meant we were leaking connections on the sidecar. This change sidesteps that problem by just sharing connections everywhere.

### DIFF
--- a/sdk/sync/watch.go
+++ b/sdk/sync/watch.go
@@ -14,10 +14,12 @@ import (
 
 // Watcher exposes methods to watch subtrees within the sync tree of this test.
 type Watcher struct {
-	re     *runtime.RunEnv
-	client *redis.Client
-	root   string
-	subs   sync.WaitGroup
+	re        *runtime.RunEnv
+	client    *redis.Client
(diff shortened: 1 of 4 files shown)


sdk/sync/watch.go
```diff
@@ -149,6 +155,9 @@ func (w *Watcher) Barrier(ctx context.Context, state State, required int64) <-ch
 //
 // Note: Concurrently closing the watcher while calling Subscribe may panic.
 func (w *Watcher) Close() error {
-	defer w.subs.Wait()
-	return w.client.Close()
+	w.closeOnce.Do(func() {
+		close(w.close)
+		w.subs.Wait()
+	})
```
Stebalien: Intentionally don't close the client as it's shared.

PR #577, opened 2020-02-20
benchmarks

coryschwartz:
Still a work in progress, but thee are a few plans that can be used as benchmarks.

### DIFF
--- a/plans/testground-test/signalbarrier.go
+++ b/plans/testground-test/signalbarrier.go
@@ -22,7 +22,15 @@ func signalBarrierBench(runenv *runtime.RunEnv) error {
 		ImprovementDir: -1,
 	}
 
+	// Wait until all the nodes are started
+	runenv.RecordMessage("waiting for others to become ready.")
+	writer.SignalEntry(ctx, sync.State("ready"))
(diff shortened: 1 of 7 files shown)


plans/testground-test/signalbarrier.go
```diff
@@ -0,0 +1,48 @@
...
+		start := time.Now().UnixNano()
+		writer.SignalEntry(ctx, ping)
+		_ = <-watcher.Barrier(ctx, ping, int64(i*runenv.TestInstanceCount))
+		end := time.Now().UnixNano()
+		writer.SignalEntry(ctx, pong)
+		nanoseconds := end - start
+		total = nanoseconds
+		runenv.RecordMetric(&barrierMetric, float64(nanoseconds))
+	}
+	runenv.RecordMessage("Average barrier latency, %d", int(total)/iterations)
```
nonsense: This test does exactly what we need, but also showcases that we need the metrics library - the average is not a good enough metric to tell us if our system is performing well - I think we'd want to measure:
- 95%, 99%, max for both: a) writer.signalEntry and b) watcher.Barrier;

coryschwartz:
Closing, re-implemented as "baseline" benches.

raulk:
For future reference: this was merged in #604.

PR #580, opened 2020-02-20
Listen for more Docker events, not just `stop`

nonsense:
This PR is adding:
1. More docker events, not just `stop` - also `die` and `destroy`.
2. Listen for `ctx.Done()` in the `networkChanges` channel loop
3. Remove of the `errgroup` as we have only 1 goroutine

---

~Looks like this was a problem only on the `cluster:k8s` runner, `master` seems to be working just fine for `local:docker`.~ Actually this seems to be a problem even on `local:docker`.

---

Fixes: https://github.com/ipfs/testground/issues/549

### DIFF
--- a/pkg/dockermanager/dockermanager.go
+++ b/pkg/dockermanager/dockermanager.go
@@ -123,9 +123,7 @@ func (dm *Manager) Manage(
 	defer cancel()
 
 	stop := func(container string) {
-		dm.S().Infow("got stop event", "container", container)
 		if m, ok := managers[container]; ok {
-			dm.S().Infow("calling m.cancel()", "container", container)
 			m.cancel()
 			delete(managers, container)
 
@@ -198,6 +196,8 @@ func (dm *Manager) Manage(
 	eventFilter.Add("type", "container")
 	eventFilter.Add("event", "start")
 	eventFilter.Add("event", "stop")
+	eventFilter.Add("event", "destroy")
+	eventFilter.Add("event", "die")
 
 	// Manage new containers.
 	eventCh, errs := dm.Client.Events(ctx, types.EventsOptions{
@@ -212,6 +212,8 @@ func (dm *Manager) Manage(
 			case "start":
 				start(event.ID)
 			case "stop":
+			case "destroy":
+			case "die":
 				stop(event.ID)
 			default:
 				return fmt.Errorf("unexpected event: type=%s, status=%s", event.Type, event.Status)
--- a/pkg/dockermanager/dockermanager.go
+++ b/pkg/dockermanager/dockermanager.go
@@ -123,7 +123,9 @@ func (dm *Manager) Manage(
 	defer cancel()
 
 	stop := func(container string) {
+		dm.S().Infow("got stop event", "container", container)
 		if m, ok := managers[container]; ok {
+			dm.S().Infow("calling m.cancel()", "container", container)
 			m.cancel()
 			delete(managers, container)
 
--- a/pkg/dockermanager/dockermanager.go
+++ b/pkg/dockermanager/dockermanager.go
@@ -211,9 +211,7 @@ func (dm *Manager) Manage(
 			switch event.Status {
 			case "start":
 				start(event.ID)
-			case "stop":
-			case "destroy":
-			case "die":
+			case "stop", "destroy", "die":
 				stop(event.ID)
 			default:
 				return fmt.Errorf("unexpected event: type=%s, status=%s", event.Type, event.Status)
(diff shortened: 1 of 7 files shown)


pkg/dockermanager/dockermanager.go
```diff
@@ -123,7 +123,9 @@ func (dm *Manager) Manage(
 	defer cancel()
 
 	stop := func(container string) {
+		dm.S().Infow("got stop event", "container", container)
 		if m, ok := managers[container]; ok {
+			dm.S().Infow("calling m.cancel()", "container", container)
 			m.cancel()
```
nonsense: This is currently not working, for some reason. Looks like we don't propagate contexts correctly.

pkg/sidecar/instance.go
```diff
@@ -61,6 +61,9 @@ func NewInstance(ctx context.Context, runenv *runtime.RunEnv, hostname string, n
 
 // Close closes the instance. It should not be used after closing.
 func (inst *Instance) Close() error {
```
nonsense: This doesn't appear to be called when a container is stopped.

nonsense:
Looks like we have bigger problems - we detect `stop` event from Docker, only for the `pause` containers, and not for the actual `testplan` containers... no idea why, when we do detect `start` events for those container (after all we attach networks to them, etc.)

pkg/sidecar/sidecar.go
```diff
@@ -53,53 +51,59 @@ func Run(runnerName string) error {
...
+
+		// Now let the test case tell us how to configure the network.
+		subtree := sync.NetworkSubtree(instance.Hostname)
+		networkChanges := make(chan *sync.NetworkConfig, 16)
+		if err := instance.Watcher.Subscribe(ctx, subtree, networkChanges); err != nil {
+			return fmt.Errorf("failed to subscribe to network changes: %s", err)
+		}
+		for {
+			select {
+			case <-ctx.Done():
```
nonsense: @Stebalien I guess this was not needed previously, as this whole block was inside an `errgroup`.

pkg/dockermanager/dockermanager.go
```diff
@@ -196,6 +196,8 @@ func (dm *Manager) Manage(
 	eventFilter.Add("type", "container")
 	eventFilter.Add("event", "start")
 	eventFilter.Add("event", "stop")
+	eventFilter.Add("event", "destroy")
+	eventFilter.Add("event", "die")
```
nonsense: Not sure if this was a problem for other runners, but on k8s we were getting a `stop` event only for the `pause` containers (which we ignore). All the `testplan` containers, were sending `die`.

pkg/sidecar/sidecar.go
nonsense: Actually this appears to not be the case. Looks like we were never catching that particular cancel?

pkg/sidecar/sidecar.go
```diff
@@ -53,53 +51,59 @@ func Run(runnerName string) error {
...
+				err := ctx.Err()
+				if err != nil && err != context.Canceled {
+					instance.S().Warnw("context return err different to canceled", "err", err.Error())
+				}
+				return nil
+			case cfg := <-networkChanges:
```
raulk: Can `networkChanges` be closed without the ctx firing? If so, this would lead to a nil pointer panic now. Before, the for loop would yield.

pkg/sidecar/sidecar.go
```diff
@@ -53,53 +51,59 @@ func Run(runnerName string) error {
...
+			instance.S().Debugw("closing instance", "instance", instance.Hostname)
 			if err := instance.Close(); err != nil {
 				instance.S().Warnf("failed to close instance: %s", err)
 			}
 		}()
 
-		g, ctx := errgroup.WithContext(ctx)
-
 		// Network configuration loop.
-		g.Go(func() error {
```
raulk: Did we only ever schedule one goroutine in this errgroup?

nonsense: They used to be two, but since it is only one now, I am removing the error group.

pkg/sidecar/sidecar.go
nonsense: I don't think anyone is closing this channel - we use the sync service to subscribe and push to it.

In any case I will add an `ok` check.

pkg/dockermanager/dockermanager.go
```diff
@@ -210,6 +212,8 @@ func (dm *Manager) Manage(
 			case "start":
 				start(event.ID)
 			case "stop":
+			case "destroy":
```
raulk: nit: IIRC you can do `case "stop", "destroy", "die":`

PR #583, opened 2020-02-20
TerminateAll on cluster:k8s

coryschwartz:
This implementation uses  a local exec to `kops` to terminate a k8s cluster.

I think this is really not ideal, but given the way we establish the cluster uses kops. I feel that
we will have to change this later, but this works for me right now.

### DIFF
--- a/pkg/runner/cluster_k8s.go
+++ b/pkg/runner/cluster_k8s.go
@@ -12,6 +12,7 @@ import (
 	"math/rand"
 	"net"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"reflect"
 	"strconv"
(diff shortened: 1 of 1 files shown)


pkg/runner/cluster_k8s.go
```diff
@@ -646,3 +647,28 @@ func (c *ClusterK8sRunner) maxPods() (int, error) {
 
 	return pods, nil
 }
+
+// If the testground daemon has access to enough environment variables, we can terminate using kops
+// if not, then print a messgae about how to do it.
+// I would like to note that when the daemon returns to the client, the cluster may not be deleted
+// yet. This means it will report to the client a success, even if kops is not able to complete the
+// deletion.
+func (c *ClusterK8sRunner) TerminateAll() error {
```
nonsense: I am not sure we want `testground` to have control over infrastructure.

I don't think we want to abstract `kops` or `terraform` or any similar tool inside Testground.

Testground should assume that infrastructure and dependencies are setup correctly. The border between what is Testground and what are dependencies is a bit vague, but I'd say cloud resources are not part of Testground.

nonsense:
`TerminateAll` is about workload, not about infrastructure. Once we go into Testground-as-a-service, users of Testground, won't have so much control, and it should not be possible for the daemon to shutdown the infrastructure.

coryschwartz:
I re-created this in https://github.com/ipfs/testground/pull/590 with a different strategy, to delete only pods labeled for plan.

coryschwartz:
Closed in favor of https://github.com/ipfs/testground/pull/590

PR #587, opened 2020-02-21
Don't share with user go/pkg

coryschwartz:
Fix: https://github.com/ipfs/testground/issues/505

I don't personally see the value in sharing with the system go pkg dir.

I think the intention is to make this useful as a cache, to which I would point out that the goproxy
container already does this even without the bind mount, since it creates an unnamed volume, which
is mounted to the container at `/go`

The unnamed volume is reasonably resiliant for our purpose  -- it remains even when the
container is killed, restarted, etc. I think we should remove the bind mount and keep the volume
that gets created by default.

To see the benefit of this patch, rebuild the ipfs/testground image with
`make docker-ipfs-testgrond`

See that the bind mount is no longer used present with `docker inspect`, etc.

### DIFF
--- a/pkg/build/golang/docker.go
+++ b/pkg/build/golang/docker.go
@@ -15,8 +15,6 @@ import (
 	"sync"
 	"time"
 
-	gobuild "go/build"
-
 	"github.com/ipfs/testground/pkg/api"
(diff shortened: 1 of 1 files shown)


raulk:
We are going with the solution in #591.

PR #588, opened 2020-02-21
fix: run the goproxy as the owner of the go package cache.

### LINKED ISSUE #505: Goproxy creates files in your home directory as root
We setup a bind mount to the default GOAPTH for the goproxy container
https://github.com/ipfs/testground/blob/master/pkg/build/golang/docker.go#L379

This container will drop files into ~/go/pkg as the root user.

Steps to reproduce:

```
1. Kill the goproxy container (if it's running)
2. mv ~/go ~/go.orig
3. ./testground daemon
4. ./testground run any test that will trigger a build
5. find ~/go/pkg -user root (there will be some)
```

I think this bind mount can just be removed, after all the Dockerfile already creates a volume for /go 

https://hub.docker.com/r/goproxy/goproxy/dockerfile

Stebalien:
This ensures we don't create files in the go package cache as the wrong user.

Also run the goproxy even if the local GOPATH doesn't exist. The goproxy container will use a volume so we'll still see some speedup.

fixes #505

### DIFF
--- a/pkg/build/golang/docker_unix.go
+++ b/pkg/build/golang/docker_unix.go
@@ -1,0 +1,19 @@
+//+build linux darwin freebsd
+
+package golang
+
+import (
+	"fmt"
+	"os"
+
+	"golang.org/x/sys/unix"
+)
+
+func getOwner(path string) (owner string, err error) {
+	info, err := os.Stat(path)
+	if err != nil {
+		return "", err
+	}
+	stat := info.Sys().(unix.Stat_t)
+	return fmt.Sprintf("%d:%d", stat.Uid, stat.Gid), nil
+}
--- a/go.mod
+++ b/go.mod
@@ -43,7 +43,7 @@ require (
 	github.com/vishvananda/netns v0.0.0-20191106174202-0a2b9b5464df
 	go.uber.org/zap v1.12.0
 	golang.org/x/sync v0.0.0-20190911185100-cd5d95a43a6e
-	golang.org/x/sys v0.0.0-20200202164722-d101bd2416d5 // indirect
+	golang.org/x/sys v0.0.0-20200202164722-d101bd2416d5
 	golang.org/x/tools v0.0.0-20191216052735-49a3e744a425 // indirect
 	k8s.io/api v0.17.0
 	k8s.io/apimachinery v0.17.0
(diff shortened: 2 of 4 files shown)


Stebalien:
@coryschwartz here's an alternative to #587 that keeps sharing the go cache _and_ should correctly set the UIDs. Does this fix your issue?

pkg/build/golang/docker_unix.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"golang.org/x/sys/unix"
+)
+
+func getOwner(path string) (owner string, err error) {
+	info, err := os.Stat(path)
+	if err != nil {
+		return "", err
+	}
+	stat := info.Sys().(unix.Stat_t)
```
coryschwartz:
```suggestion
	stat := info.Sys().(*syscall.Stat_t)
```

pkg/build/golang/docker_unix.go
```diff
@@ -0,0 +1,19 @@
+//+build linux darwin freebsd
+
+package golang
+
+import (
+	"fmt"
+	"os"
+
+	"golang.org/x/sys/unix"
```
coryschwartz:
```suggestion
	"syscall"
```

coryschwartz:
Irrespective of the code and permission problem, which could be fixed pretty easily, I just don't like the idea of bind-mounting a container into a home directory.

I am unsure how well this works with any MAC, but I think by having the bind mount in the home directory, users with SELINUX enabled would need to edit the context, for example. Even if this were working perfectly, I would be more in favor of just removing the bind mount, though I know I am in the minority with this opinion.

raulk:
Note that the original intention of this feature was precisely to share the host’s go mod cache with the build container. In my experience, this sped up my builds substantially. If some setups are known to fail, we could detect them and bypass the bind mount dynamically. That way we don’t penalise all users of testground to please a subset.

The advantage of this approach is that it works well with CI caches too.

There’s a third option. We could use volume mounts. They are safe from `docker system prune -a` and only deleted if `--volumes` is set. It would still take long to populate the first time, but it could be a fair tradeoff that works everywhere.

coryschwartz:
The goproxy already uses a volume mount for /go as seen [here](https://github.com/goproxyio/goproxy/blob/master/Dockerfile#L16) -- This allows the volume to persist when the container is restarted, but of course the reference to this volume is lost whenever the system is pruned.

I would not be opposed to creating named volume that we could continue to even if the system were pruned.

raulk:
I could live with that. Wanna take a stab, @coryschwartz?

raulk:
Closed in favour of #591.

PR #590, opened 2020-02-24
TerminateAll() delete pods by label

coryschwartz:
This is take two for https://github.com/ipfs/testground/pull/583

In this version, don't delete the whole cluster, just clear out plan pods


pkg/runner/cluster_k8s.go
```diff
@@ -556,6 +556,7 @@ func (c *ClusterK8sRunner) createPod(ctx context.Context, podName string, input
 				"testground.testcase": runenv.TestCase,
 				"testground.run_id":   input.RunID,
 				"testground.groupid":  g.ID,
+				"testground.purpose":  "plan",
```
coryschwartz: I used the same label here as I did in https://github.com/ipfs/testground/pull/581

but I am open to doing this another way, if we don't like this.

nonsense: I think this is fine. We might want to introduce a filter for all things Testground related, such as `app=testground`, but we definitely need to differentiate between testground-related services (redis, sidecar, etc.) and testplan instances, so this filter could be used for that.

pkg/runner/cluster_k8s.go
```diff
@@ -646,3 +647,21 @@ func (c *ClusterK8sRunner) maxPods() (int, error) {
...
+	log := logging.S()
+	client := c.pool.Acquire()
+	defer c.pool.Release(client)
+
+	planPods := metav1.ListOptions{
+		LabelSelector: "testground.purpose=plan",
+	}
+	err := client.CoreV1().Pods(c.config.Namespace).DeleteCollection(&metav1.DeleteOptions{}, planPods)
+	if err != nil {
+		log.Errorw("Could not terminate all pods.", err)
```
nonsense: I think `Errorw` expected an odd number of params: the first param is general msg, and the following params are key-value pairs.

Maybe change to:

```
log.Errorw("could not terminate all pods", "err", err)
```

nitpick: please use lower-case, as this is what we use everywhere for Infow, Warnw and Errorw.

coryschwartz:
```suggestion
		log.Errorw("could not terminate all pods.", "err", err) 
```

PR #596, opened 2020-02-25
Switch S3 store for EFS

nonsense:
Fixes https://github.com/ipfs/testground/issues/584.

This PR is adding infra playbooks and amending the `install.sh` script to create a Kubernetes cluster and AWS EFS (file system) for outputs collection.

The infra code is ready for review:
- [x] S3 references have been removed
- [x] EFS file system is created (after `kops` has setup a cluster) with `terraform`, as we need to attach the EFS to the same security groups and subnets and create a mount target.
- [x] `delete.sh` should be called to tear down the setup, as now we have more cloud resources, not just the `kops` cluster - we also call `terraform destroy`.

Note that all testplans for a given cluster are sharing the same EFS drive. This is not ideal, because a misbehaving testplan might remove the outputs of another testplan.

I am not sure if we can mount only a specific directory under a given EFS drive to a given pod. I think this is not critical for this stage of the project, but definitely critical when we move to TaaS (Testground as a Service).

---

TODO:
- [x] Adjust `CollectOutputs` command to get all files from EFS to the local `testground client`.
- [x] `tar -cz` all the files into one file at the end of a successful run. This means that we add latency at the end of test runs, but the download will be orders of magnitude faster.
- [x] `tar -cz` all the files on call to `testground collect` if a run was NOT successful.
- [x] Clean up commented out Go code.
- [x] Investigate why starting of containers is slow - it wasn't slow a few days ago when I tested with 3k containers, but now we are running an older version of k8s. - some results can be found at: https://gist.github.com/nonsense/84ca481efa46b740f564364e4d2cf7e1
- [x] Add dependencies to README - `terraform` and ~~`jq`~~ currently.

---

TODO NICE TO HAVE:
- [x] figure out the DNS issue when starting the `efs-provisioner` - currently starting this pod is rather slow due to some DNS issues - this results in the initial installation of the cluster becoming with another ~40-60sec. slower.
Resolution: not sure if this is a problem on our end, today this was starting up from first attempt, could have been a DNS issue on AWS side.


infra/k8s/efs-terraform/backend.tf
```diff
@@ -0,0 +1,3 @@
+terraform {
+  backend "s3" {}
```
nonsense: We keep the Terraform state in the same bucket as the KOPS state.

infra/k8s/efs/manifest.yaml.spec
```diff
@@ -0,0 +1,72 @@
...
+  annotations:
+    volume.beta.kubernetes.io/storage-class: "aws-efs"
+spec:
+  accessModes:
+    - ReadOnlyMany
+    - ReadWriteOnce
+  storageClassName: aws-efs
+  resources:
+    requests:
+      storage: 100Mi
```
nonsense: We might want to configure the size of volume later on, for now I think it's fine.

infra/k8s/install.sh
```diff
@@ -66,19 +54,70 @@ while [ "$READY_NODES" -ne $(($WORKER_NODES + 1)) ]; do READY_NODES=$(kubectl ge
...
+
+kubectl apply -f ./efs/rbac.yaml \
+              -f $EFS_MANIFEST_SPEC
 
 echo "Install Weave, CNI-Genie, s3bucket DaemonSet, Sidecar Daemonset..."
 echo
 kubectl apply -f ./kops-weave/genie-plugin.yaml \
               -f ./kops-weave/weave.yml \
-              -f ./kops-weave/s3bucket.yml \
+              -f ./kops-weave/dummy.yml \
```
nonsense: We no longer need the `s3bucket` daemonset, but we still need at least one container on each host, so that the `cni0` interface is created on Flannel.

infra/k8s/install.sh
```diff
@@ -66,19 +54,70 @@ while [ "$READY_NODES" -ne $(($WORKER_NODES + 1)) ]; do READY_NODES=$(kubectl ge
...
+  echo "Couldn't detect AWS Subnet created by `kops`"
+  exit 1
+fi
+
+echo "Detected Subnet ID: $subnetId"
+
+pushd efs-terraform
+
+# extract s3 bucket from kops state store
+S3_BUCKET="${KOPS_STATE_STORE:5:100}"
```
nonsense: KOPS_STATE_STORE has the format of `s3://bucket-name`, whereas here we need `bucket-name`.

pkg/runner/cluster_k8s.go
```diff
@@ -540,13 +548,13 @@ func (c *ClusterK8sRunner) createPod(ctx context.Context, podName string, input
...
-	sharedVolumeName := "s3-shared"
+	//hostpathtype := v1.HostPathType("DirectoryOrCreate")
+	sharedVolumeName := "efs-shared"
 
-	mnt := v1.HostPathVolumeSource{
-		Path: fmt.Sprintf("/mnt/%s/%s/%d", input.RunID, g.ID, i),
-		Type: &hostpathtype,
-	}
+	//mnt := v1.HostPathVolumeSource{
+	//Path: fmt.Sprintf("/mnt/%s/%s/%d", input.RunID, g.ID, i),
```
nonsense: This is a big difference - previously we had isolation since we mounted `/mnt/%s/%s/%d` to individual pods, whereas now we mount the full EFS file system to each pod and ask individual testplan instances to write to `/outputs/%s-%s-%s-%d` - not ideal :/

infra/k8s/delete.sh
```diff
@@ -0,0 +1,47 @@
...
+
+set -e
+
+err_report() {
+    echo "Error on line $1"
+}
+
+trap 'err_report $LINENO' ERR
+
+vpcId=`aws ec2 describe-vpcs --filters Name=tag:Name,Values=$NAME | jq ".Vpcs | .[] | .VpcId" | tr -d "\""`
```
nonsense: @coryschwartz any ideas/suggestions on how to fetch `vpcId` `securityGroupId` and `subnetId` without introducing a dependency on `jq` ?

coryschwartz: Not a whole lot better, really:

```
vpcId=$(aws ec2 describe-vpcs --output text | awk '/VPCS/ {print $8 }')
```

nonsense: I wasn't aware of this `--output` flag, thanks a lot, I think I'd rather rely on `awk` than on `jq` :)

nonsense:
> I wonder if we can jsut use something like https://github.com/cloudposse/terraform-aws-kops-efs for the terraform parts

I considered this, but this is outdated (pre terraform 0.12, upgrading doesn't get it to work out of the box either), and also requires that we again maintain both `terraform` and `kops`, resulting in a similar flow. The only benefit would be that we won't have to `detect` the vpc, subnet and sg the way we do here, but the complexity of pushing all cloud resources to `terraform` is not great.

pkg/runner/cluster_k8s.go
```diff
@@ -288,6 +283,13 @@ func (c *ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow io.W
 		return nil, err
 	}
 
+	runName := fmt.Sprintf("compress-tg-%s-%s", input.TestPlan.Name, input.RunID)
```
nonsense: Note that this part runs only when the run has `completed` (i.e. it is successful, or it has timed-out).

I imagine that devs will be killing Testground for a while, until their testplan is correct, so we should probably have a way to compress and propagate back the outputs on a canceled run as well.

nonsense:
I believe this is now complete and ready for review. I think it is more robust than the `s3fs` solution.

pkg/runner/cluster_k8s.go
```diff
@@ -303,54 +305,114 @@ func (*ClusterK8sRunner) CompatibleBuilders() []string {
...
+	err := c.ensureCollectOutputsPod(ctx)
+	if err != nil {
+		return err
+	}
 
-	sess, err := session.NewSession(&aws.Config{
-		Region: aws.String(cfg.OutputsBucketRegion)},
-	)
+	file := fmt.Sprintf("/outputs/%s.tgz", input.RunID)
+	args := fmt.Sprintf("kubectl exec %s cat %s -- ", collectOutputsPodName, file)
```
nonsense: I considered doing this through the Kubernetes API, but couldn't get it to work, and the code was much more complex, compared to just calling `kubectl` with `exec`.

pkg/runner/cluster_k8s.go
```diff
@@ -251,7 +237,15 @@ func (c *ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow io.W
 			eg.Go(func() error {
 				defer func() { <-sem }()
 
-				return c.createPod(ctx, podName, input, runenv, env, g, i)
+				currentEnv := make([]v1.EnvVar, len(env))
+				copy(currentEnv, env)
+
+				currentEnv = append(currentEnv, v1.EnvVar{
+					Name:  "TEST_OUTPUTS_PATH",
+					Value: fmt.Sprintf("/outputs/%s/%s/%d", input.RunID, g.ID, i),
```
nonsense: @raulk note that we no longer have the `testplan name` (for example `tg-dht`) in the `/outputs` path.

The reason for the change is that if a run is canceled or fails, and a user calls `testground collect --runner cluster:k8s 920588cc18f3`, we have no idea in which dir to look for `920588cc18f3`. In the current scheme, we check:
1. Is a `tgz` file present? If yes - then the run must have succeeded (or someone already called `testground collect`)
2. If a `tgz` is not present, it means that the run was not successful, so we archive the outputs at that point - whatever has been stored...

I hope that makes sense.

nonsense:
cc @aschmahmann you could use this instead of S3 - it is much faster and seems quite robust on my end.

nonsense:
Adding @Stebalien here, just so that I highlight how this differs from the other good idea we had on how to get rid of `s3fs`.

1. Without EFS, we would have to enhance the `sidecar` to manage logs for every testplan container, and `move` the logs to S3 for example on `stop` of the container. We wouldn't be able to just archive all outputs at all hosts.

2. With EFS, as implemented in this PR, it is rather easy to just spin up a container (or `exec` into one - this is the `collect-outputs` pod), and have all the outputs at the same place and just archive them - I find this simpler than what we'd have to do in 1. but ultimately we are moving some complexity to the infrastructure layer - creating EFS, attaching it, etc.

---

Nevertheless this works fast IMO and is reliable so mostly solves the problems we had with `s3fs`.

Robmat05:
Reassigning to @coryschwartz to resolve conflicts, review and merge if satisfied. Cory, please reach out to @Stebalien if you have any questions

coryschwartz:
I made an edit to CollectOutputs so it no longer uses local exec, and removed the intermediate compression step. Instead the compression is done on the fly during collection. but the overall way this works is the same. -- there is always a collect-outputs pod hanging around and the collection is done 

@raulk if you agree, I would like to transition this to an EphemeralContainer, and then remove the ensureCollectContainer and waitForPod methods.

raulk:
@coryschwartz -- you had various follow-ups you wanted to capture somewhere. Could you open an issue and reference this PR from it? Thanks!

Robmat05:
This ticket is now closed, with the remainder of the to do tasks identified in #668

nonsense:
> I made an edit to CollectOutputs so it no longer uses local exec

Great stuff, I tried that initially and failed, but I see you got it working!

> and removed the intermediate compression step. Instead the compression is done on the fly during collection.

Not sure what you mean with that - we seem to be doing the compression in the same way, with `tar -czf` ...

> but the overall way this works is the same. -- there is always a collect-outputs pod hanging around and the collection is done

> @raulk if you agree, I would like to transition this to an EphemeralContainer, and then remove the ensureCollectContainer and waitForPod methods.

This is just so that we don't have to attach to EFS multiple times. The idea is that if you have multiple runs and need to exec into a container that has access to EFS, you might as well keep the container around, rather than start/stop it which is slow.

PR #602, opened 2020-02-26
Do Not Merge. WIP: Lotus Filecoin testplan

jimpick:
I've been experimenting with using Testground as a basis for writing testplans for the [Lotus](https://github.com/filecoin-project/lotus) Filecoin implementation.

I'm only creating this PR as a pointer to my temporary working branch.

https://github.com/jimpick/testground/tree/scratch/lotus-base

The branch is super messy as I'm just experimenting with how to do things.

There are a couple of things I think could be turned into testground PRs:

* I hacked it up so I could have a custom Dockerfile for a single testplan.
* I also need to have a configurable filesystem mount (using a bind mount for docker:local, a "PersistentVolumeClaim" mount on k8s).

One of the main things I'm doing there is having my testplan binary exec non-testplan target binaries (lotus and lotus-storage-miner) and just using a client API to control a test, which is a different way of using testground than the intended pattern.

When there's time, I will create some smaller PRs that add specific features so I'm not using a forked version of testground.


nonsense:
Thanks, @jimpick, this is useful.

Are you aware of any limits on attaching a `PersistentVolumeClaim` to many pods, for example 10k? I tried EFS with 3k containers, and it worked fine, but I wonder if there is another limit I might not be aware of, which might prevent us from attaching the same EFS filesystem to more pods in the future when we scale Testground?

jimpick:
@nonsense If there are any limits, it would just be on the AWS EFS side of things. EFS is just NFS ... all Kubernetes is doing is setting up an NFS mount inside the container.

jimpick:
@nonsense There are many other interesting things we can do with PersistentVolumeClaims, eg. mount local disks, or EBS disks. It would be nice to have a generic solution ... perhaps we could define the name of a PersistantVolumeClaim in the .env.toml file and allow mounting of that. Might also be nice to handle binding local directory mountpoints with a similar mechanism (eg. Filecoin implementations all default to looking for proof files in /var/tmp/filecoin-proof-parameters, so all the devs have the files there).

nonsense:
@jimpick thanks for the suggestions!

I think mounting EBS drives doesn't work for our use case, because you can only mount a given EBS drive to one host, and ideally we want all the outputs assets to be at the same place without too much complexity on Testground side.

So far we've been using a mount to a local disk, which was running `s3fs`, but this turns out is not very reliable.

Mounting the same EFS to all testplan instances seems to be working fine for now. Alternatively we will add a daemon to every host to listen for Docker events (same as sidecar) and move outputs at the end of testplan runs.

jimpick:
Here's a video of it in action:

https://gateway.ipfs.io/ipfs/QmYhhbscDSGcc2T1t6mdb2eV3DRyTCRtyF9qMe2rwQTr9a/testground-lotus-1.mp4

pkg/sidecar/k8s_instance.go
```diff
@@ -322,6 +323,7 @@ func (n *K8sNetwork) ConfigureNetwork(ctx context.Context, cfg *sync.NetworkConf
 		if err != nil {
 			return fmt.Errorf("failed to generate new network config list: %w", err)
 		}
+		logging.S().Debugw("Jim", "netconf", netconf)
```
nonsense, line 326: @jimpick heads up, that at larger clusters we seem to be having issues with this part and it is not yet clear why: https://github.com/testground/testground/pull/924

Looks like currently it is possible to block Weave somehow.

raulk:
@jimpick I'm closing this experimentation PR because it's been inactive for a while and Lotus testplans already landed on the Lotus repo ;-)

PR #604, opened 2020-02-26
baseline benchmarks

coryschwartz:
Benchmarks for https://github.com/ipfs/testground/issues/597

I'm missing a few things that would be required to close out that issue. I'm missing a benchmark for doing network changes, and I am missing the visualization portion (because I'm just using ReportMetric right now)

I'll get those down tomorrow.


plans/benchmarks/benchmarks.go
```diff
@@ -0,0 +1,164 @@
...
+	doneState := sync.State("net configured " + name)
+
+	// A new network configuration
+	netConfig := sync.NetworkConfig{
+		Network: "default",
+		IPv4:    &runenv.TestSubnet.IPNet,
+		State:   doneState,
+	}
+	// Change the IP address.
+	// Not checking if the IP address I'm changing to already exists, by the way
```
nonsense: I don't think we can do that, this is expected to fail, but it would be good to know how exactly it fails. In any case I wouldn't spend too much time on it now, as manual IP changing from testplans is not on the critical path for the short term.

coryschwartz: I *can* do this on local docker... but you're right on k8s... Taking it out!

coryschwartz:
I think these become much more useful once the metrics are added to the SDK, like in https://github.com/ipfs/testground/pull/606 To edit these to push metrics using that method would be very simple.

sdk/runtime/runenv.go
```diff
@@ -7,28 +7,30 @@ import (
...
-	EnvTestRepo               = "TEST_REPO"
-	EnvTestSubnet             = "TEST_SUBNET"
 	EnvTestCaseSeq            = "TEST_CASE_SEQ"
-	EnvTestSidecar            = "TEST_SIDECAR"
-	EnvTestInstanceCount      = "TEST_INSTANCE_COUNT"
-	EnvTestInstanceRole       = "TEST_INSTANCE_ROLE"
-	EnvTestInstanceParams     = "TEST_INSTANCE_PARAMS"
 	EnvTestGroupID            = "TEST_GROUP_ID"
 	EnvTestGroupInstanceCount = "TEST_GROUP_INSTANCE_COUNT"
+	EnvTestInstanceCount      = "TEST_INSTANCE_COUNT"
```
coryschwartz: Sorry for editing more lines, I just sorted these lexical order. 

There is one line I added, for TestStartTime.

Similarly for other areas in this file

coryschwartz:
Once the metrics PR is merged https://github.com/ipfs/testground/pull/606/ I'll re-visit this, and add metrics onto it.

PR #605, opened 2020-02-26
testplan metrics with go-metrics, InfluxDB, Grafana

nonsense:
A sample PR, which is using the `go-metrics` library for metrics collection and reporting to InfluxDB.


plans/dht/test/barrier.go
```diff
@@ -0,0 +1,82 @@
...
+		if err != nil {
+			return err
+		}
+		runenv.RecordMetric(&runtime.MetricDefinition{
+			Name:           fmt.Sprintf("stage-time"),
+			Unit:           "ns",
+			ImprovementDir: -1,
+		}, float64(time.Since(t).Nanoseconds()))
+
+		metrics.GetOrRegisterResettingTimer("barrier.test", nil).UpdateSince(t)
```
nonsense: Example for how to measure a timer with predefined percentiles (avg, 95, 99, etc.)

plans/dht/test/find_providers.go
```diff
@@ -15,15 +18,25 @@ import (
 )
 
 func FindProviders(runenv *runtime.RunEnv) error {
+	metrics.GetOrRegisterCounter("find-providers.run", nil).Inc(1)
```
nonsense: Example for measuring a counter.

plans/dht/test/find_providers.go
```diff
@@ -15,15 +18,25 @@ import (
...
+	//errm := m.EmitMetrics()
+	//if errm != nil {
+	//fmt.Println("couldn't emit metrics: %v", errm)
+	//}
+
+	defer metrics.GetOrRegisterResettingTimer("find-providers.total", nil).UpdateSince(time.Now())
```
nonsense: Another example for measuring a timer with percentiles.

plans/dht/metrics/metrics.go
```diff
@@ -0,0 +1,29 @@
...
+var (
+	endpoint  = "http://influxdb:8086"
+	database  = "metrics"
+	username  = ""
+	password  = ""
+	namespace = ""
+	tags      = map[string]string{}
+)
+
+func Setup() {
```
nonsense: Initialisation for the InfluxDB reporter, with sane defaults that could work out-of-the-box in our cluster env.

coryschwartz:
By the way, there is a grafana dashboard in the monitoring namespace which is added by the prometheus operator kops addon [see here](https://github.com/kubernetes/kops/tree/master/addons/prometheus-operator).

It doesn't include influxdb. Would you prefer this over https://github.com/ipfs/testground/pull/606 or would you want to add influxdb in addition?

nonsense:
> By the way, there is a grafana dashboard in the monitoring namespace which is added by the prometheus operator kops addon [see here](https://github.com/kubernetes/kops/tree/master/addons/prometheus-operator).

Yes, I saw that, the Grafana instance coming from the `kops` addon is nice, we have to however make sure that the dashboards coming with it are meaningful and be able to provision automatically additional dashboards on it.

> It doesn't include influxdb. Would you prefer this over #606 or would you want to add influxdb in addition?

I don't have preference for influxdb over prometheus, but we need to make sure that data models are good for our needs - with this I mean that timers and histograms are tracked are reset per interval, so that we don't encounter the problems described at: http://taint.org/2014/01/16/145944a.html

raulk:
Final solution with InfluxDB was implemented here: https://github.com/testground/sdk-go/pull/3. Thanks a lot for this experiment/spike!

PR #606, opened 2020-02-26
Prometheus pushgateway sdk integration

coryschwartz:
Added the pushgateway to the prometheus SDK, although RecordMetric does not use this.

Plan writers can Add() or Push() metrics into the pushgateway and they become available for
the prometheus scraper.

### DIFF
--- a/sdk/runtime/metrics.go
+++ b/sdk/runtime/metrics.go
@@ -32,10 +32,11 @@ func NewGauge(runenv *RunEnv, name string, help string) prometheus.Gauge {
 	return g
 }
 
-func NewHistogram(runenv *RunEnv, name string, help string) prometheus.Histogram {
+func NewHistogram(runenv *RunEnv, name string, help string, buckets ...float64) prometheus.Histogram {
 	h := prometheus.NewHistogram(prometheus.HistogramOpts{
-		Name: name,
-		Help: help,
+		Name:    name,
+		Help:    help,
+		Buckets: buckets,
 	})
 	runenv.MetricsPusher.Collector(h)
 	return h
--- a/sdk/runtime/metrics.go
+++ b/sdk/runtime/metrics.go
@@ -50,10 +50,6 @@ func NewSummary(runenv *RunEnv, name string, help string) prometheus.Summary {
 	return s
 }
 
-func NewTimer(o prometheus.Observer) *prometheus.Timer {
-	return prometheus.NewTimer(o)
-}
-
 // MustExportPrometheus starts an HTTP server with the Prometheus handler.
 // It starts on a random open port and returns the listener. It is the caller
 // responsability to close the listener.
--- a/sdk/runtime/metrics.go
+++ b/sdk/runtime/metrics.go
@@ -8,38 +8,50 @@ import (
 	"os"
 	"path"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
-	"github.com/prometheus/client_golang/prometheus/push"
 )
 
-// NewPrometheusPusher instantiates a prometheus pusher object which is configured to use
-// the testgroudn prometheus pushgateway
-func (runenv *RunEnv) NewPrometheusPusher(metricName string, collectors ...prometheus.Collector) *push.Pusher {
-
-	fullMetricName := strings.Join([]string{
-		runenv.TestPlan,
-		runenv.TestCase,
-		metricName},
-		"/")
-
-	registry := prometheus.NewRegistry()
-	registry.MustRegister(collectors...)
-
-	return push.New("http://pushgateway:9091", fullMetricName)
-}
-
-// NewPrometheusGauge is a helper function for creating prometheus metrics.
-// This is here to prevent plans from needing to import prometheus directly,
-// but the object returend is the same as that returend by prometheus.NewGauge
-func (runenv *RunEnv) NewPrometheusGauge(name string, help string) prometheus.Gauge {
-	return prometheus.NewGauge(prometheus.GaugeOpts{
+func NewCounter(runenv *RunEnv, name string, help string) prometheus.Counter {
+	c := prometheus.NewCounter(prometheus.CounterOpts{
 		Name: name,
 		Help: help,
 	})
+	runenv.MetricsPusher.Collector(c)
+	return c
+}
+
+func NewGauge(runenv *RunEnv, name string, help string) prometheus.Gauge {
+	g := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: name,
+		Help: help,
+	})
+	runenv.MetricsPusher.Collector(g)
+	return g
+}
+
+func NewHistogram(runenv *RunEnv, name string, help string) prometheus.Histogram {
+	h := prometheus.NewHistogram(prometheus.HistogramOpts{
+		Name: name,
+		Help: help,
+	})
+	runenv.MetricsPusher.Collector(h)
+	return h
+}
+
+func NewSummary(runenv *RunEnv, name string, help string) prometheus.Summary {
+	s := prometheus.NewSummary(prometheus.SummaryOpts{
+		Name: name,
+		Help: help,
+	})
+	runenv.MetricsPusher.Collector(s)
+	return s
+}
+
+func NewTimer(o prometheus.Observer) *prometheus.Timer {
+	return prometheus.NewTimer(o)
 }
 
 // MustExportPrometheus starts an HTTP server with the Prometheus handler.
--- a/sdk/runtime/metrics.go
+++ b/sdk/runtime/metrics.go
@@ -8,10 +8,39 @@ import (
 	"os"
 	"path"
 	"strconv"
+	"strings"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/prometheus/client_golang/prometheus/push"
 )
+
+// NewPrometheusPusher instantiates a prometheus pusher object which is configured to use
+// the testgroudn prometheus pushgateway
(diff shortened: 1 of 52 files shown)


coryschwartz:
Proof that this works:

in a container with access to the control network:

```
curl pushgateway:9091/metrics
```

and I see the following which appear from the example:

```
# HELP db_backup_last_success_timestamp_seconds The timestamp of the last successful completion of a DB backup.
# TYPE db_backup_last_success_timestamp_seconds gauge
db_backup_last_success_timestamp_seconds{instance="",job="example/prometheus/db_backup"} 1.5827552691531312e+09
```

This is exactly what I would expect to see.

coryschwartz:
@raulk I've added the pusher to the RunEnv, and edited the Invoke method so that it automatically Push()'s metrics.

I am using the methods in metrics.go to automatically add the collector to the new runenv pusher,

There's a bug in the current commit, and metrics aren't actually pushed, but I want to update this to show the direction it's heading and solicit for more comments

sdk/runtime/runner.go
```diff
@@ -10,17 +10,25 @@ import (
...
 
+	startGauge := NewGauge(runenv, "start_time", "time of plan start")
+	startGauge.SetToCurrentTime()
+	runenv.RecordStart()
+
+	endGauge := NewGauge(runenv, "end_time", "time of plan end")
 	errfile, err := runenv.CreateRawAsset("run.err")
 	if err != nil {
 		runenv.RecordCrash(err)
+		endGauge.SetToCurrentTime()
```
nonsense: You can use `defer` to have a single call to `endGauge.SetToCurrentTime()` instead of 7 - otherwise it is very easy for someone to forget to add it if a new case appears, once this code gets refactored in the future.

nonsense:
@coryschwartz does this fit with the requirement that we want to see periodic stats per a given interval (5sec. or 10sec.) ? I think what @raulk says, is only part of the problem.

> In runner.Invoke, maybe call the pusher once the test ends (whether OK, in error, or in panic).

If we have testplans running for 10mins, we don't want to emit metrics once at the end of the testplan, but periodically.

sdk/runtime/metrics.go
```diff
@@ -10,9 +10,50 @@ import (
...
+	g := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: name,
+		Help: help,
+	})
+	runenv.MetricsPusher.Collector(g)
+	return g
+}
+
+func NewHistogram(runenv *RunEnv, name string, help string) prometheus.Histogram {
+	h := prometheus.NewHistogram(prometheus.HistogramOpts{
```
nonsense: I think we are supposed to define the buckets here, as per https://godoc.org/github.com/prometheus/client_golang/prometheus#Histogram ?

coryschwartz:
I added an additional example, and  he ability to override buckets on the histogram.

I also added automatic pushing, which the plan writer user doesn't need to think about for long-running plans

sdk/runtime/metrics.go
coryschwartz: You're right, fixed!

sdk/runtime/runenv.go
```diff
@@ -99,14 +100,24 @@ type RunEnv struct {
...
+	MetricsPusher *push.Pusher
+	unstructured  chan *os.File
+	structured    chan *zap.Logger
 }
 
 // NewRunEnv constructs a runtime environment from the given runtime parameters.
 func NewRunEnv(params RunParams) *RunEnv {
 	re := &RunEnv{
-		RunParams:    params,
+		RunParams: params,
```
coryschwartz: These groupings cause the metrics to have searchable fields, so I have just added a bunch of them here.

The metric with this configuration looks like this:
```
db_backup_records_processed{TestCase="prometheus",TestCaseSeq="",TestCommit="",TestGroupID="single",TestPlan="example",TestRun="626439704b90",TestTag="",instance="",job="testground/plan"} 42
```

sdk/runtime/runenv.go
```diff
@@ -99,14 +100,24 @@ type RunEnv struct {
...
+		MetricsPusher: push.New("http://pushgateway:9091", "testground/plan").
```
coryschwartz: The address of the push gateway is hardcoded here.

On the k8s runner, this the pushgateway has a service.

I could imagine making this a configurable option if/when we want to publish metrics to a remote source.

nonsense:
The tests are failing because there is no push service running. We should probably have an option to disable pushing for tests.

coryschwartz:
Some images of the graphs in action, these are silly graphs from the example. Shown is load of the cluster, average random number generated over time, and rate of bucket insertion into the histogram.


2020-02-28-183009_1737x1071_scrot
2020-02-28-183025_1731x1073_scrot
2020-02-28-183037_1742x1054_scrot

coryschwartz:
I'm not sure that the tests are actually because of the failure to push to prometheus -- It should be *recording* a failure, but not actually failing the test.

coryschwartz:
To prevent the test failures from being recorded, I changed the prometheus push failure message to use RecordMessage rather than RecordFailure... This makes sense to me since a single message push shouldn't result in the test failing overall.

coryschwartz:
@raulk @nonsense Are the instructions I added to the readme good enough to show how this works?

sdk/runtime/runenv.go
nonsense: We could pass it the same way as we do for Redis. Not sure if we are going to add support for Prometheus for the rest of the runners - `local:docker`, etc.

For now I am happy to leave it hardcoded to the k8s service address. Once we have useful dashboards, we could think about adding support for the other runners if necessary.

nonsense:
@coryschwartz instructions for opening Prometheus web UI are fine.

I think we need more examples on the visualisation part in Grafana and example dashboards. This is not trivial, even if you've used other metrics systems in the past:
- visualise a counter
- visualise a 95%ile, 99%ile and mean for a given set of timers (for example the 3 stages I added as a draft PR at https://github.com/ipfs/testground/pull/616)
- visualise the same set of metrics (counter and timers) coming from a number of different testrun instances and testrun ids - we can start with 10 or 50 instances, but we have to think how we are going to aggregate when we have 1k and 10k.

Also it appears that the scrape intervals on Prometheus and Push gateway are a bit too high - 30sec. - we should investigate if we can lower them.

coryschwartz:
Fixed concerns brought up by@nonsense

Quintile queries metrics work correctly -- at this time, I've just added a comment to the exampe to give ideas to readers about what to query for, and how do perform certain kinds of queries. 

I think we can add dashboards to the grafana built into the prometheus operator, and that would help people to see the queries more quickly and easily, and without having to figure out just how to write the query. The dashboards will be specific to every test, so I wouldn't add them here.

sdk/runtime/runenv.go
```diff
@@ -102,14 +103,26 @@ type RunEnv struct {
...
 	re := &RunEnv{
-		RunParams:    params,
+		RunParams: params,
+		MetricsPusher: push.New("http://prometheus-pushgateway:9091", "testground/plan").
+			Gatherer(prometheus.NewRegistry()).
+			Grouping("TestPlan", params.TestPlan).
+			Grouping("TestCase", params.TestCase).
+			Grouping("TestRun", params.TestRun).
+			Grouping("TestGroupID", params.TestGroupID).
+			Grouping("TestCaseSeq", string(params.TestCaseSeq)).
```
nonsense: I don't think we need all these groupings - they probably add a lot of overhead to the measurements.

We can remove them later though.

`TestCaseSeq` is def not needed.

nonsense:
Overall LGTM.

sdk/runtime/runenv.go
```diff
@@ -102,14 +103,26 @@ type RunEnv struct {
...
+	structured    chan *zap.Logger
 }
 
 // NewRunEnv constructs a runtime environment from the given runtime parameters.
 func NewRunEnv(params RunParams) *RunEnv {
+	containerName, _ := os.Hostname()
 	re := &RunEnv{
-		RunParams:    params,
+		RunParams: params,
+		MetricsPusher: push.New("http://prometheus-pushgateway:9091", "testground/plan").
```
nonsense: @coryschwartz on a second thought - how is this actually working? I would expect for testplan instances to not have route to `prometheus-pushgateway` by default, only be able to resolve it on DNS. I don't see changes to the `sidecar`, so `prometheus-pushgateway` would be resolved correctly, but I am not sure why individual testplans would have a route to the IP of the `pushgateway`?

Then again I did run this with https://github.com/ipfs/testground/pull/616 and was getting metrics in Prometheus...

For the `influxdb` example, I had to add route to influxdb: see https://github.com/ipfs/testground/pull/605/files#diff-f6eeba9855b7c2b6fb2f75fcfcd26a48

sdk/runtime/runner.go
```diff
@@ -5,13 +5,43 @@ import (
...
 	runenv := CurrentRunEnv()
+	start := time.Now()
+	durationGuage := NewGauge(runenv, "plan_duration", "Run time (Seconds)")
+
+	// The prometheus pushgateway has a customized scrape interval, which is used to hint to the
+	// prometheus operator at which interval the it should be scraped. This is currently set to 5s.
+	// To provide an updated metric in every scrape, jobs will push to the pushgateway at the same
+	// interval. When this "push_interval" is changed, you may want to change the scrape interval
+	// on the pushgateway
+	push_interval := 5
```
coryschwartz: I *am* worried about the scalability when we have a high number of nodes.

I think the nodes will be started fairly unevenly, so the pushes will come in probably evenly-distributed, but the load of the pushgateway would grow with the size of the cluster. A test with 1,000 instances will cause the pushgateway to take on 250 PUT requests per second. Although I think we can deal with this when it comes up.

sdk/runtime/runenv.go
coryschwartz: it's unroutable to the sidecar, but for the plan containers, the service resource created is within a routable range.

PR #616, opened 2020-03-02
example measurements - counters and timers

nonsense:
This PR is adding a set of very simple measurements:
1. Counter that gets incremented once every 2 seconds.
2. Three stages with 100 calls to a function that has latency around 100ms, 200ms and 500ms

We should have an example dashboard on how to visualise those, so that developers using Testground know how to use the `runenv` metrics SDK and how to visualise the results.

- [ ] visualise a counter
- [ ] visualise a 95%ile, 99%ile and mean for the 3 stages
- [ ] visualise all the metrics above grouped by a number of different testrun instances and testrun ids - we can start with 10 or 50 instances, but we have to think how we are going to aggregate when we have 1k and 10k instances and multiple testruns.

cc @coryschwartz

### DIFF
--- a/manifests/example.toml
+++ b/manifests/example.toml
@@ -62,3 +62,7 @@ instances = { min = 1, max = 200, default = 5 }
 [[testcases]]
 name = "prometheus2"
 instances = { min = 1, max = 200, default = 5 }
+
+[[testcases]]
+name = "prometheus3"
+instances = { min = 1, max = 200, default = 5 }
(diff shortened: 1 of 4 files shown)


Robmat05:
looks like the success criteria for this ticket is ready to test. Assigning it to @coryschwartz to review and merge. I may be misunderstanding this tickets readiness for review though, so don’t hesitate to let to move it back into In Progress if I’ve misunderstood

coryschwartz:
I added some new timers to the prometheus2 example, I'll create a dashboard for it as well. 

This is a draft status still, but for monitoring, more is more. Yes to merging more and better examples.

Robmat05:
Good Job Cory, So should this be merged, closed and another ticket opened to improve upon this for v0.4? Or should we keep this open and try to improve it for v0.3 in the next sprint?

Sorry for being dense, just looking for a bit more clarity - RM

PR #617, opened 2020-03-02
remove time.Sleep; fix coordination between k8s-sidecar

nonsense:
Fixes: https://github.com/ipfs/testground/issues/414

This is supposed to fix the `time.Sleep` hack I introduced a while ago, so that we make sure that Kubernetes is done getting a `PodIP`, before we mess with container networking with `sidecar`.

This PR makes the assumption that it is `safe` to change the networking after a container is in `Running` phase. I am not 100% sure that Kubernetes is done getting the `PodIP` at that stage, but it probably it.

### DIFF
--- a/pkg/dockermanager/dockermanager.go
+++ b/pkg/dockermanager/dockermanager.go
@@ -163,7 +163,7 @@ func (dm *Manager) Manage(
 			err := worker(cctx, handle)
 			if err != nil {
 				if errors.Is(err, context.Canceled) {
-					handle.S().Warnf("sidecar worker failed: %s", err)
+					handle.S().Debugf("sidecar worker failed: %s", err)
 				} else {
 					handle.S().Errorf("sidecar worker failed: %s", err)
 				}
--- a/pkg/sidecar/k8s_instance.go
+++ b/pkg/sidecar/k8s_instance.go
@@ -9,6 +9,7 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/containernetworking/cni/libcni"
 	"github.com/vishvananda/netlink"
@@ -486,6 +487,8 @@ func waitForPodRunningPhase(ctx context.Context, podName string) error {
 			}
 
 			phase = string(pod.Status.Phase)
+
+			time.Sleep(1 * time.Second)
 		}
 	}
 }
--- a/pkg/sidecar/k8s_instance.go
+++ b/pkg/sidecar/k8s_instance.go
@@ -9,11 +9,13 @@ import (
 	"net"
 	"os"
 	"path/filepath"
-	"time"
 
 	"github.com/containernetworking/cni/libcni"
 	"github.com/vishvananda/netlink"
 	"github.com/vishvananda/netns"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
 
 	"github.com/ipfs/testground/pkg/dockermanager"
 	"github.com/ipfs/testground/pkg/logging"
@@ -83,14 +85,6 @@ func (d *K8sInstanceManager) Close() error {
 }
 
 func (d *K8sInstanceManager) manageContainer(ctx context.Context, container *dockermanager.Container) (inst *Instance, err error) {
-	// TODO: sidecar is racing to modify container network with CNI and pod getting ready
-	// we should probably adjust this function to be called when a pod is in `1/1 Ready` state, and not just listen on the docker socket
-	select {
-	case <-time.After(20 * time.Second):
-	case <-ctx.Done():
-		return nil, ctx.Err()
-	}
-
 	// Get the state/config of the cluster
 	info, err := container.Inspect(ctx)
 	if err != nil {
@@ -98,7 +92,7 @@ func (d *K8sInstanceManager) manageContainer(ctx context.Context, container *doc
 	}
 
 	if !info.State.Running {
-		return nil, fmt.Errorf("not running")
+		return nil, fmt.Errorf("container is not running: %s", container.ID)
 	}
 
 	// Construct the runtime environment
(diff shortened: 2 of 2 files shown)


pkg/dockermanager/dockermanager.go
```diff
@@ -163,7 +163,7 @@ func (dm *Manager) Manage(
 			err := worker(cctx, handle)
 			if err != nil {
 				if errors.Is(err, context.Canceled) {
-					handle.S().Warnf("sidecar worker failed: %s", err)
+					handle.S().Debugf("sidecar worker failed: %s", err)
```
nonsense: This is an expected condition for every single `initContainer` within a testplan instance pod. Currently we have an `initContainer` which creates the `outputs` directory prior to the testplan binary starting, so that the loggers can store logs there.

Therefore I am downgrading to Debug, from Warn.

Stebalien: We probably shouldn't even log this.

pkg/sidecar/k8s_instance.go
```diff
@@ -454,3 +458,34 @@ func getRedisRoute(handle *netlink.Handle, redisIP net.IP) (*netlink.Route, erro
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("wait for pod context (pod name: %s) erred with: %w", podName, ctx.Err())
+		default:
+			if phase == "Running" {
+				return nil
+			}
+			pod, err := k8sClientset.CoreV1().Pods("default").Get(podName, metav1.GetOptions{})
```
Stebalien: Sleep?
Also, does it scale to run this once per instance?

pkg/dockermanager/dockermanager.go
nonsense: Well, we can disable those log lines with the command flag for level of logs. This log line was still useful in confirming what's happening with initContainers that inherit the env vars from the pod.

pkg/sidecar/k8s_instance.go
nonsense: It def scales to run it once per instance - not sure if we need a pool of clients though (as we have in `testground` which makes thousands such calls), but since we have no more than 100 pods per sidecar, I think we are fine, unless we notice this to get slower. I tested with 30 pods on an instance, and the call was instant.

---

I will add a time.Sleep!

pkg/dockermanager/dockermanager.go
Stebalien: (we should at least note that this isn't a failure)

pkg/sidecar/k8s_instance.go
Stebalien: Ah, so it doesn't call a _central_ server?

Stebalien: Not taking a context here is a bit annoying but not the end of the world.

pkg/sidecar/k8s_instance.go
```diff
@@ -454,3 +459,36 @@ func getRedisRoute(handle *netlink.Handle, redisIP net.IP) (*netlink.Route, erro
...
+			if err != nil {
+				return fmt.Errorf("error in wait for pod running phase: %v", err)
+			}
+
+			phase = string(pod.Status.Phase)
+
+			time.Sleep(1 * time.Second)
```
Stebalien: This should probably select on a context.

Stebalien: Really, we could replace the `default:` with a ticker/timer.

pkg/sidecar/k8s_instance.go
Stebalien: It looks like the next, unreleased version of this API supports contexts.

pkg/dockermanager/dockermanager.go
nonsense: Not sure what you mean. Do you mean renaming `sidecar worker: failed` to `sidecar worker: returned` ?

This is not a failure, and we don't log it as one. The problem is that the `initContainer` inherits the env vars, so sidecar `thinks` it should manage networks for it - something we don't really want or need.

The worker did return an error and it should, because its operation was canceled, I think. We could alternatively detect the cancelation of context in the worker and try not to return an error, but `nil`, but I am not sure how much worth is in doing that.

---

On a second though we might have a race here, if the `initContainer` completes in the time we get an event from Docker daemon and call `start()`, but the time we get to:

```
	info, err := container.Inspect(ctx)
	if err != nil {
		return nil, fmt.Errorf("inspect failed: %w", err)
	}

	if !info.State.Running {
		return nil, fmt.Errorf("not running")
	}
```

Then we could even get an error log line for the initContainer that it is `not running` anymore.

pkg/sidecar/k8s_instance.go
nonsense: It is calling the `kube-apiserver`, which we could scale horizontally if we need to at some point (and these calls would then be load balanced).

pkg/sidecar/k8s_instance.go
nonsense: I think this is simpler, than having a ticker that has to instantly tick (as we don't want to wait on the first entry of this case), and then have to `Timer.Reset` it to the period (1 second in this case)... just time.Sleep seems much simpler, no?

nonsense: The benefit of the ticker is that we won't be delaying the cancelation if it happens though...

pkg/dockermanager/dockermanager.go
Stebalien: Don't we return "context canceled" every time we shut down because the context was canceled (e.g., we got a stop event)?

nonsense: We do shut down by canceling the context, but we handle it differently at various places.

1. If get context canceled during the initial phase, we return it, and currently log it (the change in this PR, because this used to be something that "shouldn't happen")

2. If we cancel the context during the `network changes` for-loop (at `pkg/sidecar/sidecar.go`), we detect it and don't return it (we return `nil`).

Stebalien: Ah...

PR #623, opened 2020-03-03
initial cluster:k8s healthcheck support

nonsense:
TODO:
- [ ] refactor `initRunner`
- [ ] add `fix` functionality

--

Duplicates #625 .


raulk:
I just opened #625 and immediately after realised that this PR exists! ;-)

Robmat05:
@raulk @coryschwartz , please review this issue and approve/merge if satisfied

coryschwartz:
This shouldn't get merged until https://github.com/ipfs/testground/pull/596 to provide efs support, or else health checks will always fail.

coryschwartz:
Using a cluster I built using the efs branch, and the health check comes back with this:

```
↬ ./testground healthcheck -runner cluster:k8s
Mar  5 22:48:47.230918	INFO	testground client initialized	{"addr": "localhost:8080"}
checking runner cluster:k8s
finished checking runner cluster:k8s
Checks:
- k8s: ok; k8s cluster is running
- efs: ok; efs provisioner is running
- redis: ok; redis service is running
- sidecar: ok; sidecar service is running
No fixes applied.
```

Robmat05:
Roger that, Moving this ticket out of Review status and back to In Progress pending review and merge of #596. Once 596 merged we will test again and if pass healthchecks will merge and close this ticket

pkg/runner/cluster_k8s.go
```diff
@@ -293,6 +295,131 @@ func (*ClusterK8sRunner) ID() string {
...
+		}
+	}
+
+	sidecarCheck = api.HealthcheckItem{Name: "sidecar", Status: api.HealthcheckStatusOK, Message: "sidecar service is running"}
+	return
+}
+
+func (c *ClusterK8sRunner) Healthcheck(fix bool, engine api.Engine, writer io.Writer) (*api.HealthcheckReport, error) {
+
+	// TODO: initialize the runner correctly.
```
coryschwartz: @raulk @Robmat05 

This is obviously not the right thing to do.

DoRun in engine.go runs the healthcheck for the build and run. This is poorly implemented here. The "fix" is unused, and because the runner is uninitialized, even the checks don't work the first time it's tried.

If this is to go in 0.3, I will have to focus on this to tackle these two points. I would also be okay with opening two tickets as separate items
1. Implement healthcheck fix on k8s
2. come up  with a better k8s runner init.

pkg/runner/cluster_k8s.go
```diff
@@ -293,6 +295,131 @@ func (*ClusterK8sRunner) ID() string {
...
+	efsCheck = api.HealthcheckItem{Name: "efs", Status: api.HealthcheckStatusFailed, Message: "efs provisioner is not running"}
+
+	client := c.pool.Acquire()
+	defer c.pool.Release(client)
+
+	res, err := client.CoreV1().Pods(c.config.Namespace).List(metav1.ListOptions{
+		LabelSelector: "app=efs-provisioner",
+	})
+	if err != nil {
+		return
```
raulk: Should override `efsCheck.Message` with the error.

pkg/runner/cluster_k8s.go
```diff
@@ -293,6 +295,131 @@ func (*ClusterK8sRunner) ID() string {
...
+	}
+	if len(res.Items) != 1 {
+		return
```
raulk: Should override `efsCheck.Message` with a string like: "expected 1 EFS provisioner pod; found: $len(res.Items)"

pkg/runner/cluster_k8s.go
```diff
@@ -293,6 +295,131 @@ func (*ClusterK8sRunner) ID() string {
...
+	}
+
+	pod := res.Items[0]
+	if pod.Status.Phase != "Running" {
+		return
+	}
```
raulk: Let's factor this logic out to a method. It's duplicated above too.

pkg/runner/cluster_k8s.go
```diff
@@ -293,6 +295,131 @@ func (*ClusterK8sRunner) ID() string {
...
+	if len(pods.Items) != nodes {
+		return
+	}
+
+	for _, pod := range pods.Items {
+		if pod.Status.Phase != "Running" {
+			return
+		}
+	}
```
raulk: Extract this assertion to a method, that's reused from the points above too.

pkg/runner/cluster_k8s.go
raulk: Yeah, the init logic needs to be untangled. It ends up initialising a singleton k8s runner with the configuration it first sees. But that config is really a _run-specific_ configuration, so it's incorrect to globally initialise the runner with it.

IMO, what should happen is:

* Healthcheck checks and/or initialises a pool (fix action) by picking up the Kubernetes config from the env (`Engine.EnvConfig()`), or defaulting to the default configuration.
* The `podResourceCPU` and `podResourceMemory` state variables that are populated on init from the first run configuration we see should not be in the state of the runner. They should be calculated per-run.
* I don't understand why `c.maxAllowedPods` is even in the state of the runner -- it's calculated every time a run happens, and it's racy under a concurrent scenario.

pkg/runner/cluster_k8s.go
```diff
@@ -293,6 +295,131 @@ func (*ClusterK8sRunner) ID() string {
...
+			c.healthcheckSidecar(),
+		}
+	}
+
+	if !fix {
+		return &report, nil
+	}
+
+	// TODO: implement fix
+	return &report, nil
```
raulk: Could you add a check such that if any of the checks are failed, we populate the fixes slice indicating that the pertinent fix was _omitted_ with message "unimplemented"?

Robmat05:
Per @coryschwartz :

“For the local docker runner, there are a few ensure-xxx, for the required containers. but on k8s, the infra is currently started outside of testground, Besides, kubernetes has some self-healing properties. So I would probably not focus on the "fix" path at first, But I really don't want to have that nil pointer problem for very long.  Between the two, I see that as the bigger problem”

pkg/runner/cluster_k8s.go
coryschwartz: Added  the not implemented message

Now if everything is actually working, it returns like this, with or without the `-fix` flag:
```
checking runner cluster:k8s
finished checking runner cluster:k8s
Checks:
- k8s: ok; k8s cluster is running
- efs: ok; efs provisioner is running
- redis: ok; redis service is running
- sidecar: ok; sidecar service is running
No fixes applied.
```

If I kubectl delete the sidecar resources, I get an expected failure:

```
checking runner cluster:k8s
finished checking runner cluster:k8s
Checks:
- k8s: ok; k8s cluster is running
- efs: ok; efs provisioner is running
- redis: ok; redis service is running
- sidecar: failed; sidecar service is not reachable
No fixes applied.
```

And with the `-fix` flag applied, it shows the message.
```
checking runner cluster:k8s
finished checking runner cluster:k8s
Checks:
- k8s: ok; k8s cluster is running
- efs: ok; efs provisioner is running
- redis: ok; redis service is running
- sidecar: failed; sidecar service is not reachable
Fixes:
- sidecar: failed; Fix not implemented yet for this check.

```

coryschwartz:
I've opened an issue to fix the runnerInit, and for now the fixes are not yet implemented.  Merging now that @raulk's suggested changes have been made

pkg/runner/cluster_k8s.go
```diff
@@ -293,6 +295,146 @@ func (*ClusterK8sRunner) ID() string {
...
+			c.healthcheckSidecar(),
+		}
+
+		if fix {
+			fakeFixes := []api.HealthcheckItem{}
+			for _, chk := range report.Checks {
+				if chk.Status != api.HealthcheckStatusOK {
+					fakeFixes = append(fakeFixes, api.HealthcheckItem{
+						Name:    chk.Name,
+						Status:  chk.Status,
```
raulk: `Status` should be "Omitted" here. We haven't performed the fix, it can't have failed.

PR #627, opened 2020-03-03
adopt an issue labelling scheme.

raulk:
Inspired by https://github.com/libp2p/libp2p/blob/master/ISSUE-LABELLING.md.

We can use as much of this as we want. For now, I'm interested in the area, hints, statuses, and priority labels.

### DIFF
--- a/.github/labels.yml
+++ b/.github/labels.yml
@@ -1,0 +1,235 @@
+## Color palette
+
+# yellows:  dba355 d8a038 d8bd36 edd17d
+# greens:   92ef92 6bbf3b 1cef5c 75b72d 9fea8f c6e84e c1f45a b8d613 fcf0b5 
+# reds:     dd362a
+# blues:    5b91c6 2a7d93 0bb1ed
+# pinks:    bf0f73
+# oranges:  ba500e ce8048
+# teals:    40c491
+
+###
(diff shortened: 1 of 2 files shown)


Robmat05:
LGTM

.github/labels.yml
```diff
@@ -0,0 +1,235 @@
...
+- name: area/runners
+  color: 0bb1ed
+  description: "Area: Runners"
+- name: area/builders
+  color: 0bb1ed
+  description: "Area: Builders"
+- name: area/infrastructure
+  color: 0bb1ed
+  description: "Area: Infrastructure"
+- name: area/infra
```
nonsense: This seems off, should be `area/specs`

.github/labels.yml
```diff
@@ -0,0 +1,235 @@
...
+  color: 0bb1ed
+  description: "Area: Specs"
+- name: area/specs
```
nonsense: `area/docs`

.github/labels.yml
```diff
@@ -0,0 +1,235 @@
...
+  color: 0bb1ed
+  description: "Area: Documentation"
+- name: area/docs
```
nonsense: `area/testing`

PR #630, opened 2020-03-04
plans/dht/test: Remove duplicate log.

petar:
Fix logging the same info twice.


nonsense:
Tests are failing due to the same problem as the one mentioned at https://github.com/ipfs/testground/issues/624

```
--- FAIL: TestBuildDockerGo (45.79s)
    build_test.go:34: fatal error from daemon: engine build error: network testground-build is ambiguous (2 matches found on name)
```

---

We started seeing this a few days ago, so my hunch is we somehow broke the builder with one of the recent commits.

PR #632, opened 2020-03-04
install redis-exporter

nonsense:
TODO:
- [x] install `redis-exporter` helm chart, that deploys a service which exposes /metrics endpoint
- [x] scrape that /metrics endpoint and get measurements into prometheus
- [x] visualise measurements in grafana
- [x] automate installation of exporter, scraping config
- [ ] automate import of grafana dashboard

### DIFF
--- a/infra/k8s/redis-exporter-values.yaml
+++ b/infra/k8s/redis-exporter-values.yaml
@@ -37,17 +37,17 @@ annotations:
 
 serviceMonitor:
   # When set true then use a ServiceMonitor to configure scraping
-  enabled: false
+  enabled: true
   # Set the namespace the ServiceMonitor should be deployed
   # namespace: monitoring
   # Set how frequently Prometheus should scrape
-  # interval: 30s
+  interval: 10s
   # Set path to redis-exporter telemtery-path
   # telemetryPath: /metrics
   # Set labels for the ServiceMonitor, use this to define your scrape label for Prometheus Operator
   # labels:
   # Set timeout for scrape
-  # timeout: 10s
+  timeout: 10s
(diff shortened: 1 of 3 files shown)


infra/k8s/redis-exporter-values.yaml
```diff
@@ -0,0 +1,67 @@
...
+#       key: redis-password
+#       name: redis-config-0.0.2
+service:
+  type: ClusterIP
+  port: 9121
+  annotations: {}
+  labels: {}
+resources: {}
+redisAddress: redis://redis-master:6379
+annotations:
```
nonsense: Based on what I read, adding those `annotations` should tell `prometheus` to scrape that endpoint, but this doesn't seem to be working, so I've done something wrong.

nonsense: Note to self - read the full values.yaml, no matter how long they are (facepalm)

Robmat05:
@coryschwartz Good points. So let’s either amend the ticket by removing the dashboard requirement item, and creating a new ticket in the backlog for a future milestone. Or simply delete it if we don’t think we’re going to address this issue anytime soon

coryschwartz:
I want to cancel *most* of this, since a separate node exporter is not required. I have enabled the node exporter in the existing redis values with just this line, and the prometheus operator picks this up  fine. https://github.com/ipfs/testground/pull/640/files#diff-719513b66178c36b1cd1c887da1b08a1R469

However, I do want to keep the actual dashboard.

Since @nonsense isn't here right now I'm going to go ahead and close this, but I'll create a PR with the same dashboard

PR #635, opened 2020-03-04
Bitswap tuning version compatibility plan

dirkmc:
Fixes https://github.com/ipfs/testground/issues/435

Test plan to verify that new bitswap can pull from old bitswap nodes.

This plan creates a leech pulling from two seed nodes, where each seed has two thirds of the blocks. The leech / seed combinations are:
- old <- two old
- master <- two master
- master <- two old
- master <- one old & one master

Screen Shot 2020-03-04 at 4 24 17 PM

As expected, it appears that master fetching from old bitswap nodes performs slightly more slowly than an old bitswap node fetching from old bitswap nodes (because old bitswap nodes don't support the DONT_HAVE message so it must be simulated with timeouts). 

However master does appear to be more consistent - note the cluster of red points at 32MB, showing that when an old bitswap leech pulls from two old bitswap seeds it is slow for that file size (inconsistent results at 32MB have been noted in other test plans for old bitswap)


Robmat05:
Not Testground ticket. Issue transferred to IPFS

PR #647, opened 2020-03-09
Dashboard manager

coryschwartz:
This is really less than ideal, I think, but it is better than saving json files using the UI.

This provides a basic tool to upload and download dashboards and save them in source control.


coryschwartz:
This is not exactly the experience I would hope for, but when you review this, I am considering the following pattern in which a user might create and edit dashboards

1. A test ground user, who is no expert in grafana dashboards, can use the web UI to experiment in their own cluster.
2. The testground user runs the script to download the dashboard once they are happy with it.
3. Commit the dashboard to source control, for review and acceptance.

in the hopefully-not-so-distant future, a CI job can import dashboards from source control


There is *no*  template engine or DSL to learn to generate the json files. Nothing stops us from adding that in the future, but I imagine most people will just create dashboards using the grafana web UI, and this gives us

dashboards/main.go
```diff
@@ -0,0 +1,192 @@
...
+		log.Fatal("Unknown grafana API key. Please set GRAFANA_API_KEY or use the -apikey flag")
+	}
+
+	client := sdk.NewClient(*grafana, *apikey, sdk.DefaultHTTPClient)
+
+	// import mode
+	if *imode {
+		ImportDatasources(client)
+		ImportDashboards(client)
+		os.Exit(0)
```
gmasgras: putting the backup logic in an `else` block might be a little easier to reason about

coryschwartz: Thanks. Sounds fine to me -- fixing!

nonsense:
@coryschwartz I haven't reviewed this in detail, but we could also consider https://www.terraform.io/docs/providers/grafana/index.html for importing dashboards, as we are already using terraform to provision EFS on the clusters.

PR #648, opened 2020-03-09
fix bugs when parsing default test case params

yusefnapora:
This fixes two bugs around parsing default params. Unless overridden by a composition or at the command line, string params are getting parsed surrounded by `"`  chars, e.g. `"foo"` instead of `foo`, and params with no default value are getting parsed as the string `null`.

This just checks if the `Default` value is nil or already a string before parsing.

I didn't add tests, but I did update the `example/params` test definition with params  that demonstrate the bug.  Before the patch, you get output like

```
   0.2341s    MESSAGE << instance  10 >> key: param4, value: "foo"
   0.2343s    MESSAGE << instance  10 >> key: param5, value: null
```

and afterwards you get

```
   0.4090s    MESSAGE << instance   4 >> key: param4, value: foo
```

and no value printed for `param5` that doesn't have a default set.

### DIFF
--- a/manifests/example.toml
+++ b/manifests/example.toml
@@ -50,6 +50,9 @@ instances = { min = 1, max = 200, default = 1 }
   param1 = { type = "int", desc = "some param 1", unit = "widgets", default=1 }
   param2 = { type = "int", desc = "some param 2", unit = "widgets", default=2 }
   param3 = { type = "int", desc = "some param 3", unit = "widgets", default=3 }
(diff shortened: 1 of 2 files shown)


yusefnapora:
@raulk this is the bug I was talking about yesterday; I figured it was basically the same amount of effort to fix as to write up a decent issue for it. The tests are failing on linux but not windows, and it doesn't seem related to the change, so maybe just a flaky test somewhere?

Issue #597, opened 2020-02-25
baseline test plan


raulk:
I'd like a test plan that I can run with any arbitrary number of nodes that exercises ONLY the testground components. We should exercise:

1. Network initialised.
2. Configuring the network. Waiting for a callback.
3. Signal entry on a barrier.
4. Wait on barrier on 20% nodes, barrier on 40% nodes, barrier on 60%, barrier on 80%, barrier on 100% of nodes.
5. Publish an item on a subtree.
6. Subscribe to the subtree.

Print time elapsed:

1. Between the time the run was scheduled and the task was started (for this, we need to inject a "run start" time in an env variable?).
2. Until network is initialised.
3. Until network is reconfigured.
4. Time to hit the barrier at the above checkpoints.
5. Time to publish the item on the subtree.
6. Time until we received 20%, 40%, 60%, 80%, 100% of items in the subtree via the subscription.

We should be able to visualise all these metrics in Grafana.

coryschwartz:
The baseline benchmarks I added in 604 did *most* of the things requested here.
I have added the prometheus metrics to the SDK, and then integrated the new metrics SDK to the baseline tests in https://github.com/ipfs/testground/pull/640/. The grafana dashboard is added there as well... although the dashboard does need some work. It's a bit too basic right now.

The last thing I need to do for this is to test publishing items onto a subtree and time the subscription.

coryschwartz:
Out of 12 items, I would count this as 10/12. Missing point 5 and 6 from the second group

Robmat05:
This can now be closed as #651 has been merged.

PR #651, opened 2020-03-10
baseline benchmarks: enhance with subtree benchmarks

coryschwartz:
Subtree publish bench,  the last item on https://github.com/ipfs/testground/issues/597


Robmat05:
This ticket has no tasks or dependency issues, so we will investigate it further during sprint plan to see what may be involved to close

raulk:
@Robmat05 this is a pull request (a contribution) that would effectively close #597, as @coryschwartz duly notes in the description above.

plans/benchmarks/benchmarks.go
```diff
@@ -164,3 +172,101 @@ func BarrierBench(runenv *runtime.RunEnv) error {
...
+			return string(*val.(*[]byte))
+		}}
+
+	seq, err := writer.Write(ctx, &st, runenv.TestRun)
+	if err != nil {
+		return err
+	}
+
+	// Create several tests up to 1MB
+	// Anything over 1500 is likely to have ethernet fragmentation
```
raulk: Yeah, and 1mb is certainly an abusive usage of the sync service. I would test up to 64kb.

plans/benchmarks/benchmarks.go
```diff
@@ -164,3 +172,101 @@ func BarrierBench(runenv *runtime.RunEnv) error {
...
+		for _, tst := range tests {
+			readyState := sync.State(fmt.Sprintf("ready_%d_%s", i, tst.Name))
+			testState := sync.State(fmt.Sprintf("test_%d_%s", i, tst.Name))
+			rand.Read(tst.Data)
+			_, err := writer.SignalEntry(ctx, readyState)
+			if err != nil {
+				return err
+			}
+			<-watcher.Barrier(ctx, readyState, int64(runenv.TestInstanceCount))
+			writer.SignalEntry(ctx, testState)
```
raulk: I'm not sure we need any barrier-based synchronisation here. Both publisher and subscriber know how many messages (iterations) we're going to send/receive. We should run them sequentially:

1. First the publisher. It writes each message and records the time it took to do so. I would not usage a Prometheus Gauge, but a Histogram or a Summary. A Gauge is going to give us insufficient data IMO.
2. After the publisher is done, we run the subscribers, who subscribe, start receiving, and record on a histogram the time it took between the last receive and the current one.

While it would be great to run concurrent publishers and subscribers, the resulting scenario would be highly racy and volatile, dependent on timing.

Robmat05:
@raulk  Given that this links to #597, should this be part of Milestone v0.3 or part of next sprint v0.4?

raulk:
This should be merged today.

Robmat05:
ok, tagging it v0.3 and placing it in the ‘Review’ Pipeline

sdk/runtime/runenv.go
```diff
@@ -119,9 +119,6 @@ func NewRunEnv(params RunParams) *RunEnv {
 			Grouping("TestCase", params.TestCase).
 			Grouping("TestRun", params.TestRun).
 			Grouping("TestGroupID", params.TestGroupID).
-			Grouping("TestCaseSeq", string(params.TestCaseSeq)).
-			Grouping("TestCommit", params.TestCommit).
-			Grouping("TestTag", params.TestTag).
```
raulk: I removed these groupings because they were irrelevant.

sdk/runtime/metrics.go
```diff
@@ -42,11 +42,10 @@ func NewHistogram(runenv *RunEnv, name string, help string, buckets ...float64)
 	return h
 }
 
-func NewSummary(runenv *RunEnv, name string, help string) prometheus.Summary {
```
raulk: Summaries require more customisation, so I exposed the `SummaryOpts` struct.

Robmat05:
Reviewed and merged. This can now be closed

Issue #593, opened 2020-02-25
docs for k8s dashboard appear to be wrong


nonsense:
README.md in `infra/k8s` says:

```
7. Get access to the kubernetes dashboard 

kubectl proxy

and then, direct your browser to `http://localhost:8001/ui`
```

This is however not pointing to the k8s dashboard:

<img width="671" alt="Screenshot 2020-02-25 at 13 15 06" src="https://user-images.githubusercontent.com/50459/75246833-e11fe780-57d0-11ea-8d05-2541f19a2a58.png">

nonsense:
http://localhost:8001/api/v1/namespaces/kubernetes-dashboard/services/https:kubernetes-dashboard:/proxy/ returns `services \"kubernetes-dashboard\" not found`, so I assume `kops` is installing the k8s dashboard in a different way / location.

nonsense:
Related: https://github.com/ipfs/testground/issues/228

coryschwartz:
This is true. The dashboard requires a bit more configuration, and needs additional permissions in order to act on the cluster.

https://github.com/kubernetes/kops/blob/master/docs/operations/addons.md

I think I would rather just remove the dashboard

PR #652, opened 2020-03-10
Remove broken dashboard

coryschwartz:
Fix: https://github.com/ipfs/testground/issues/593


Robmat05:
See #593 for detailed discussion

PR #657, opened 2020-03-10
Fix inverted nil check in RunParams.MarshalLogObject

yusefnapora:
fixes an inverted nil check for the `TestSubnet` param


raulk:
CI tests stalled. Gonna merge as this is non-controversial.

Robmat05:
May revisit this later depending on outcome of what is completed in #623

PR #664, opened 2020-03-10
weave scrape endpoint.

coryschwartz:
I added the prometheus configuration earlier, and this provides the service endpoint

the weave endpoints look like this in prometheus

2020-03-10-154842_2355x867_scrot


coryschwartz:
This should have been done for part of https://github.com/ipfs/testground/issues/634 . Better late than never.

PR #665, opened 2020-03-10
Break up CI steps a little bit.

coryschwartz:


cmd/itest/common_test.go
```diff
@@ -11,7 +11,9 @@ import (
 )
 
 func runSingle(t *testing.T, args ...string) error {
+	t.Log("running runSingle")
```
raulk: Can we remove the log statements? I guess you put them in to verify that `t.Parallel()` was doing the right thing?

Makefile
```diff
@@ -20,7 +20,10 @@ lint:
 
 test-build:
 	$(call eachmod,go build -o /dev/null ./...)
-	docker build -t ipfs/testground .
 
-test:
-	$(call eachmod,go test -v $(GOTFLAGS) ./...)
+test-quick:
```
raulk: We're better off using [`testing.Short()`](https://golang.org/pkg/testing/#Short) than introducing this ad-hoc config.

coryschwartz:
I am a bit surprised this does not speed up the build any more than it does.

1. I broke up build into steps that can be performed in parallel.
2. Most of the steps take less than three minutes, but of course the long test takes much longer.

I realized there was an unintended consequence to the move to the docker volume on testground-goproxy. I had no way to cache the downloaded packages! My speculation was that a considerable amount time was spent downloading packages, so to test that theory, I created a goproxy for the test and configured the long-running tests to use the goproxy I installed. This way, I could place the bind mount where it could be cached.

Even still, during the test_long job, it takes nearly eight minutes to build the sidecar docker image and the tests take up most of the remaining time. Overall taking up about 15 minutes. On the last run, we see that there were no changes in the cache, so I think this must mean there was no files for the goproxy to download.

For comparison, my laptop takes about 10-11 minutes if the cache is empty, but running the test takes a little over 2 minutes when everything is cached by the local goproxy. Unfortunately, I am unable to achieve similar results here.

I think this is a little faster than before, but not as much gain as I had anticipated.

coryschwartz:
I still want the CI to be faster, but this, does not accomplish that goal. Closing.

PR #672, opened 2020-03-11
upgrade golangci-lint to support go 1.14

aschmahmann:


raulk:
Maybe this will fix the sporadic linting timeout issues on CI too.

PR #677, opened 2020-03-12
automatically expose pprof+prometheus endpoint.

### LINKED ISSUE #698: uncomment the panic test in manifests/placebo.toml
and also make the integration tests pass.

Right now, the integration tests return a panic, when described, and this causes the integration tests to fail

Also, just running `./testground describe` will result in a similar panic response.

raulk:
With this change, test plans using `runtime.Invoke()` will automatically benefit from pprof and prometheus HTTP exports. Fixes https://github.com/ipfs/testground/issues/673.

Works in all runneres (`local:exec`, `local:docker`, `cluster:k8s` (via k8s port forwarding).

Fixes #698.


raulk:
TODO:
- [ ] publish ports on local:docker.
- [ ] make sure sidecar doesn't annul the routes.

pkg/build/golang/Dockerfile.template
```diff
@@ -64,4 +64,5 @@ COPY --from=0 /tmp/delete_me /tmp/go-ipfs/ipfs* /usr/local/bin/
 RUN rm -f /usr/local/bin/delete_me
 ENV PATH="/usr/local/bin:${PATH}"
 
+EXPOSE 6060
```
raulk: This exposes the pprof port from the docker container.

manifests/placebo.toml
```diff
@@ -52,7 +52,12 @@ instances = { min = 1, max = 250, default = 1 }
...
   some_param = { type = "int", desc = "some param", unit = "peers" }
 
-# # seq 3
-# [[testcases]]
-# name = "panic"
-# instances = { min = 1, max = 250, default = 1 }
+# seq 3
+[[testcases]]
+name = "panic"
+instances = { min = 1, max = 250, default = 1 }
```
raulk: https://github.com/ipfs/testground/pull/689 commented this out by error.

pkg/docker/container.go
```diff
@@ -127,13 +128,15 @@ func EnsureContainer(ctx context.Context, log *zap.SugaredLogger, cli *client.Cl
 	log.Infow("starting new container", "id", res.ID)
 
 	err = cli.ContainerStart(ctx, res.ID, types.ContainerStartOptions{})
-	if err == nil {
-		log.Infow("started container", "id", res.ID)
+	if err != nil {
+		return nil, false, err
```
raulk: Took the chance to rectify this logic.

pkg/runner/local_docker.go
```diff
@@ -588,6 +588,11 @@ func ensureControlNetwork(ctx context.Context, cli *client.Client, log *zap.Suga
 		ctx,
 		log, cli,
 		"testground-control",
+		// making internal=false enables us to expose ports to the host (e.g.
+		// pprof and prometheus). by itself, it would allow the container to
+		// access the Internet, and therefore would break isolation, but since
+		// we have sidecar overriding the default Docker ip routes, and
+		// suppressing such traffic, we're safe.
```
raulk: Added an explanation of why the flip from `internal=true` to `internal=false` is needed, and why it doesn't impact isolation. cc @coryschwartz

pkg/sidecar/docker_reactor.go
```diff
@@ -174,6 +174,25 @@ func (d *DockerReactor) handleContainer(ctx context.Context, container *docker.C
...
+		}
+		nlroutes, err := netlinkHandle.RouteGet(route.Gw)
+		if err != nil {
+			return nil, fmt.Errorf("failed to resolve route: %w", err)
+		}
+		gwIPs[route.Gw.String()] = struct{}{}
+		gwRoutes = append(gwRoutes, nlroutes...)
+	}
+
+	controlRoutes = append(controlRoutes, gwRoutes...)
```
raulk: This is the key. In kubernetes we don't need this, because kubernetes has native port-forwarding out of the box. In our case, we connect our test containers to the control and data networks. The former is tightly restricted, and the latter is internal and connects only the containers participating in the test. For containers to successfully expose ports to the host, besides publishing those ports, we need to ALLOW traffic from the gateway (host).

coryschwartz: Just for comparison... Without this change, from the current master, 
the route table looks like this:
```
docker exec tg-example-prometheus2-19979292b320-single-1 ip route
16.1.0.0/16 dev eth1 proto kernel scope link src 16.1.0.3 
192.18.0.3 dev eth0 src 192.18.0.7 
192.18.0.4 dev eth0 src 192.18.0.7 
```
and there is no *default* route, so no place to send packets outside of known networks.
```
docker exec tg-example-prometheus2-19979292b320-single-1 ping -c 2 google.com
ping: google.com: Temporary failure in name resolution
```

manifests/placebo.toml
raulk: :-)

pkg/sidecar/docker_reactor.go
raulk: Thanks for spotting this. Definitely unintended. I've fixed this logic in eaa06cd. Can you take it out for a spin again? Works for me.

* The host can access ports exposed by the container.
* The container cannot access public IPs.

raulk:
Re-requested review from @coryschwartz. I've rectified the logic that whitelists the host's IP addr.

pkg/sidecar/docker_reactor.go
```diff
@@ -18,6 +18,16 @@ import (
...
 
+// PublicAddr points to an IP address in the public range. It helps us discover
+// the IP address of the gateway (i.e. the Docker host) on the control network
+// (the learned route will be via the control network because, at this point,
+// the only network that's attached to the container is the control network).
+//
+// Sidecar doesn't whitelist traffic to public addresses, but it special-cases
+// traffic between the container and the host, so that pprof, metrics and other
+// ports can be exposed to the Docker host.
+var PublicAddr = net.ParseIP("1.1.1.1")
```
coryschwartz: This is the IP address of Cloudfliar DNS. I don't have a problem with that, just pointing that out.

pkg/tgwriter/tgwriter.go
```diff
@@ -44,6 +44,10 @@ type Error struct {
 }
 
 func (tgw *TgWriter) Write(p []byte) (n int, err error) {
+	if p == nil {
```
coryschwartz: Does this prevent errors from popping up?

pkg/sidecar/docker_reactor.go
raulk: Yep, that's ok. This will not perform any lookup of any kind, it just matches against the kernel routing tables.

pkg/tgwriter/tgwriter.go
raulk: It suppresses empty writes from making it onto the wire, which I found can happen when Flush is called. This was leading to the intermittent bug with the `describe` command.

PR #681, opened 2020-03-14
add EnsureImage and BuildImage

coryschwartz:
This is a primative for any infrastructure containers we want to add to the local:docker runner.

I think we could *probably* use this for the golang builder as well with very few changes.

For example, if we want to emulate the experience of the prometheus-operator on local docker (and we do) then we can use this to create a pre-configured prometheus scraper.


coryschwartz:
Actually, I decided to go ahead and edit the builder to use this.

coryschwartz:
@Robmat05 @raulk this is ready to review whenever you get the chance

In the same style as the other docker Ensure<Blah> methods, here is EnsureImage, which is the  equivalent of  `docker image ls | grep name || docker build -t name /some/where`

And BuildImage, which is the equivalent of `docker build -t name /some/where`

pkg/docker/image.go
```diff
@@ -0,0 +1,87 @@
...
+	"strings"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/client"
+	"github.com/docker/docker/pkg/archive"
+
+	"go.uber.org/zap"
+)
+
+type EnsureImageOpts struct {
```
raulk: Let's call this type `BuildImageOpts`?

pkg/docker/image.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		return err
+	}
+	defer buildResponse.Body.Close()
+	return PipeOutput(buildResponse.Body, os.Stdout)
+}
+
+// EnsureImage
+// Create a docker image from build context.
+// If an image with the requested tag already exists, don't re-create it.
```
raulk: These godocs seem like notes you wrote for yourself rather than documentation for API users? I'd try with:

> EnsureImage checks if the specified image exists. If it does, it returns immediately with `created=false`. Otherwise, it builds the image using the supplied `BuildImageOpts` parameters, and returns `created=true`, along with any errors that may have occurred during the process.

pkg/docker/image.go
```diff
@@ -0,0 +1,87 @@
...
+		ForceRemove:    true,
+		PullParent:     true,
+		Dockerfile:     "/Dockerfile",
+		Tags:           []string{name},
+	}
+}
+
+// buildImage
+// Use a Dockerfile and supporting files to build a docker image.
+// Think `docker build /path/to/build`
```
raulk: These godocs don't follow the standard style to document a Go API; looks like loose notes. Mind revisiting?

pkg/docker/image.go
```diff
@@ -0,0 +1,87 @@
...
+	// Find out if we have any images with a RepoTag which matches the name of the image.
+	// the RepoTag will be something like "name:latest" and I want to match any that have "name"
+	listOpts := types.ImageListOptions{All: true}
+	images, err := client.ImageList(ctx, listOpts)
+	if err != nil {
+		return false, err
+	}
+	for _, image := range images {
+		for _, rt := range image.RepoTags {
+			if strings.Contains(rt, opts.Name) {
```
raulk: https://golang.org/pkg/strings/#HasPrefix?

pkg/docker/image.go
```diff
@@ -0,0 +1,87 @@
...
+				log.Info("found existing image")
```
raulk: Let's add the image name to this log statement and below.

coryschwartz:
I reverted the last two commits because it was addinga  lot of extra code that isn't relevant to this. 

https://github.com/ipfs/testground/pull/689

PR #683, opened 2020-03-14
local docker prometheus

coryschwartz:
This is not yet complete, but you can see where it's going.

The prometheus endpoint comes pre-configured with a scraper for the pushgateway, just like it does with the prometheus-operator on kuberntes. It accomplishes this by building a new image with the desired configuration built in.

This builds upon https://github.com/ipfs/testground/pull/681 and depends on it.

Right now there's something I'm doing something wrong with setting up the routes, and I'll have to fix that later.

issue: https://github.com/ipfs/testground/issues/660

### DIFF
--- a/infra/docker/testground-prometheus/Dockerfile
+++ b/infra/docker/testground-prometheus/Dockerfile
@@ -1,0 +1,3 @@
+FROM prom/prometheus
+
+COPY prometheus.yml /etc/prometheus.yml
--- a/pkg/runner/local_docker.go
+++ b/pkg/runner/local_docker.go
@@ -588,7 +588,7 @@ func ensureControlNetwork(ctx context.Context, cli *client.Client, log *zap.Suga
 		ctx,
 		log, cli,
 		"testground-control",
-		true,
+		false,
 		network.IPAMConfig{
 			Subnet:  controlSubnet,
 			Gateway: controlGateway,
@@ -642,7 +642,8 @@ func ensureInfraContainer(ctx context.Context, cli *client.Client, log *zap.Suga
 			Image: imageName,
 		},
 		HostConfig: &container.HostConfig{
-			NetworkMode: container.NetworkMode(NetworkID),
+			NetworkMode:     container.NetworkMode(NetworkID),
+			PublishAllPorts: true,
 		},
 		PullImageIfMissing: pull,
 	})
--- a/pkg/runner/local_docker.go
+++ b/pkg/runner/local_docker.go
@@ -102,11 +102,12 @@ func (r *LocalDockerRunner) Healthcheck(fix bool, engine api.Engine, writer io.W
 	}
 
 	var (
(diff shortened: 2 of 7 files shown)


raulk:
This was merged without review. That's not right. Let's give the team the chance review first @coryschwartz. cc @Robmat05.

pkg/sidecar/docker_reactor.go
```diff
@@ -19,17 +19,25 @@ import (
 )
 
 type DockerReactor struct {
-	redis   net.IP
+	routes  []net.IP
 	manager *docker.Manager
 }
 
 func NewDockerReactor() (Reactor, error) {
```
raulk: How is the k8s reactor dealing with this new route?

infra/docker/testground-prometheus/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM prom/prometheus
```
raulk: We spoke about this, and I was reluctant to adding yet another moving piece that the user had to remember to build/make. I see you've decided to go this way anyway. What are the reasons? Shouldn't this be accompanied with a change in the `Makefile`?

pkg/runner/local_docker.go
```diff
@@ -542,24 +634,25 @@ func newDataNetwork(ctx context.Context, cli *client.Client, log *zap.SugaredLog
 	return id, subnet, err
 }
 
-// ensureRedisContainer ensures there's a testground-redis container started.
-func ensureRedisContainer(ctx context.Context, cli *client.Client, log *zap.SugaredLogger, controlNetworkID string) (id string, err error) {
+// ensure container is started
+func ensureInfraContainer(ctx context.Context, cli *client.Client, log *zap.SugaredLogger, containerName string, imageName string, NetworkID string, pull bool) (id string, err error) {
```
raulk: `networkID` (lower case).

pkg/runner/local_docker.go
```diff
@@ -496,7 +588,7 @@ func ensureControlNetwork(ctx context.Context, cli *client.Client, log *zap.Suga
 		ctx,
 		log, cli,
 		"testground-control",
-		true,
+		false,
```
raulk: Why the change?

Robmat05:
ok, putting this back into Review and assigning to CR team mate

infra/docker/testground-prometheus/Dockerfile
coryschwartz: We don't make this with the Makefile, I have EnsureImage building this at runtime. This is different than the bind mount, but this way we don't have to ask the user to place a config file in a specific location.

We could probably actually move the local docker sidecar to this method, and have nothing to build with the Makefile.

I should have left this open for review, rather than after the fact. I've been in a "just merge it" mode, this week, sorry I didn't give enough opportunity to review before,

PR #688, opened 2020-03-16
dont be a terminal

coryschwartz:
Don't try to send fancy terminal output.

1. It doesn't work that well. Docker command output is sometimes multi-line progress bars, which get squashed when output the way we print it now.

2. Even if it did work well, we might not want it. The output should be scroll-able -- more like a log, and should not include terminal escape sequences, cursor movements, beeps, colors, clears, etc.

### DIFF
--- a/pkg/docker/output.go
+++ b/pkg/docker/output.go
@@ -16,7 +16,7 @@ Loop:
 	for dec := json.NewDecoder(r); ; {
 		switch err := dec.Decode(&msg); err {
 		case nil:
-			_ = msg.Display(w, true)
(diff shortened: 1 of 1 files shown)


coryschwartz:
When terminal is set to false, the output is less jumpy -- multi-line progress bar outputs are not over-writing each other, for example. 

I would point out, however, that it doesn't seem like produces the *whole* output. When the command is complete, and the progress bar should be full, this is not always displayed.

## before:
Server output with terminal true

Client output with terminal true


## after:
Server output with terminal false

Client output with terminal false

PR #689, opened 2020-03-16
gomod 1.14

coryschwartz:
I think this needs to be merged now that we are on 1.14 in CI. 

See: https://github.com/ipfs/testground/pull/638


.travis.yml
```diff
@@ -38,6 +38,7 @@ jobs:
       script:
         - make tidy
         - git diff --exit-code
+        - go mod download
```
raulk: Why?

manifests/placebo.toml
```diff
@@ -52,7 +52,7 @@ instances = { min = 1, max = 250, default = 1 }
...
   some_param = { type = "int", desc = "some param", unit = "peers" }
 
-# seq 3
-[[testcases]]
-name = "panic"
-instances = { min = 1, max = 250, default = 1 }
+# # seq 3
+# [[testcases]]
+# name = "panic"
+# instances = { min = 1, max = 250, default = 1 }
```
raulk: Why?

coryschwartz: I intended to mention this in handover, but this is because integration tests fail when this is uncommented. 

Let me open an issue for this.

PR #690, opened 2020-03-16
Revert "build with go 1.14 in CI (#638)"

coryschwartz:
This reverts commit 8d58fcab86336cc3bcaf9785e58a23940573f863.

Running into some errors, that I think are caused by the CI go versions.

https://github.com/ipfs/testground/pull/689 to upgrade to 1.14 in more places, but it's currently failing CI also, so for now I want to revert this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ cache:
     - $HOME/gopath/pkg/mod
 
 go:
-  - 1.14.x
+  - 1.13.x
 
 services:
   - docker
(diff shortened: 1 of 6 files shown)


coryschwartz:
Closing, unnecessary. The problem I was seeing wasn't related.

Robmat05:
Closing this ticket as ‘discarded’ given that it was un-necessary and nothing else is pending to-do

PR #694, opened 2020-03-17
reduce failure message output

coryschwartz:
This changes what would happen in case the pushgateway is unavailable.

Instead of printing a message every five seconds, which is the current behavior, decrease from a number of max failures, and after the max failures is exhausted only then print a message that metrics are unavailable.

### DIFF
--- a/sdk/runtime/runner.go
+++ b/sdk/runtime/runner.go
@@ -23,6 +23,7 @@ func Invoke(tc func(*RunEnv) error) {
 	// interval. When this "push_interval" is changed, you may want to change the scrape interval
 	// on the pushgateway
 	pushStopCh := make(chan struct{})
+	pushAllowedFailures := 3
 	go func() {
 		pushInterval := 5 * time.Second
 		for {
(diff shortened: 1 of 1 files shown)


sdk/runtime/runner.go
```diff
@@ -23,14 +23,18 @@ func Invoke(tc func(*RunEnv) error) {
...
+	pushAllowedFailures := 3
 	go func() {
 		pushInterval := 5 * time.Second
 		for {
 			select {
 			case <-time.After(pushInterval):
 				err := runenv.MetricsPusher.Add()
 				if err != nil {
-					runenv.RecordMessage("error during periodic metric push: %w", err)
+					pushAllowedFailures -= 1
```
raulk: We should match against the actual error to take this path. If the error is different (e.g. pushgateway is actually reachable but failing), we probably want to record it repeatedly.

Robmat05:
Is this being worked on or can it be re assigned the v0.4 milestone

coryschwartz:
I think I'm just going to cancel this PR.

@raulk is right that that the way I've done this is not a good way to do this. Hopefully there aren't that many error messages anyway, and if we *do* want to reduce how verbose the error messages are, are displayed, we should do it differently than this. Raul's suggestion to count depending on the type of error would be a good one.

PR #697, opened 2020-03-17
sync: optimise barrier logic to avoid polling.

raulk:
Closes https://github.com/ipfs/testground/issues/663.

Inspired by the [RedissonCountDownLatch](https://github.com/redisson/redisson/blob/master/redisson/src/main/java/org/redisson/RedissonCountDownLatch.java).

When setting a Barrier on a state, we:

1. Subscribe to a channel `<root>/states/<state_name>/barriers/<desired_value>`.
2. GET the value, if >= `desired_value`, cancel the subscription and return.
3. Else, wait until a message is received in the subscription. When a message is received, perform another GET to confirm consistency, cancel the subscription, and return.

When signalling entry on a state, we:

1. `INCR` the value of the state.
2. Publish the new value to channel: `<root>/states/<state_name>/barriers/<new_value>`.

### DIFF
--- a/sdk/sync/redis_bench_test.go
+++ b/sdk/sync/redis_bench_test.go
@@ -18,8 +18,8 @@ func BenchmarkBarrier(b *testing.B) {
 	for n := 0; n < b.N; n++ {
 		state := State(fmt.Sprintf("yoda-%d", n))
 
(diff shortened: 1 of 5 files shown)


laurentsenta:
Closing, I believe this is outdated with the new sync service, we dropped the polling.

PR #699, opened 2020-03-17
Explicit AWS region in install script

coryschwartz:
If you have specified the ~~AWS_ZONE~~ AWS_REGION to something other than the default, kops will create the cluster that is undetectable later. Lets specify --region on all aws commands.

### DIFF
--- a/infra/k8s/install.sh
+++ b/infra/k8s/install.sh
@@ -56,7 +56,7 @@ echo
 
 echo "Install EFS..."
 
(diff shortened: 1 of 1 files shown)


coryschwartz:
right. I mean regions.

kops zones are aws regions, im mixing my terminology. I'll change the title to regions for searchability.

Robmat05:
Per @coryschwartz this can be added to v0.4 sprint and prioritized at sprint kickoff

raulk:
Given that we're merging this to master, it'll be released with Milestone v0.3. So resetting the milestone.

PR #700, opened 2020-03-17
Remove unnecessary shell

coryschwartz:
@aschmahmann Does this work on your system?

### DIFF
--- a/infra/k8s/README.md
+++ b/infra/k8s/README.md
@@ -78,6 +78,9 @@ Where:
 
 You might want to add them to your `rc` file (`.zshrc`, `.bashrc`, etc.), or to an `.env.sh` file that you source.
 
+In addition to the initial cluster setup, these variables should be accessible to the daemon. If these variables are
+manually set or you source them manually, you should make sure to do so before starting the testground daemon.
+
 ```sh
 # NAME needs be a subdomain of an existing Route53 domain name. The testground team uses .testground.ipfs.team, which is set up by our Terraform configs.
 export NAME=<desired kubernetes cluster name (cluster name must be a fully-qualified DNS name (e.g. mycluster.myzone.com)>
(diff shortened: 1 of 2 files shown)


coryschwartz:
context:

Even though running the command "kops validate cluster" works, running "sh -c kops validate cluster" doesn't. I think the use of of another shell is not needed.

Robmat05:
Is this being worked on? If so is it critical for the v0.3 sprint or should it roll to v0.4?

Robmat05:
Per @aschmahmann: I tested this process live with @Cory Schwartz today and it's fine. Although there needs to be a README somewhere that you need to have your k8s variables accessible to the shell running the testground daemon

infra/k8s/README.md
```diff
@@ -78,6 +78,9 @@ Where:
 
 You might want to add them to your `rc` file (`.zshrc`, `.bashrc`, etc.), or to an `.env.sh` file that you source.
 
+In addition to the initial cluster setup, these variables should be accessible to the daemon. If these variables are
```
nonsense: This is not ideal, because now we have config in `.env.toml` and also in environment variables, and all are required. We should consolidate config for `testground` at one place - either `.env.toml` or environment variables, but having both required is not user-friendly.

WDYT?

nonsense: Currently this seems to be a problem for `kops` and `kubectl` (since we use default context there, and this has been fine).

Wouldn't `kops validate cluster --name <NAME>` solve this, if we pick up NAME (cluster name) from `.env.toml` ? Then we need to make sure `.env.toml` is configured correctly, but we do that anyways (for example for `region`). At least we wouldn't expect users to also have to set environment variables.

raulk: @nonsense mind opening a separate issue for that?

raulk: I agree we need to converge.

raulk: This diff just describes the status quo.

nonsense:
Added an issue to fix the config: https://github.com/ipfs/testground/issues/713 and merging this, as this is strictly an improvement to master.

PR #704, opened 2020-03-18
wait for prometheus-pushgateway to be available

coryschwartz:
Fixes https://github.com/ipfs/testground/issues/685

Fixes https://github.com/ipfs/testground/issues/598

Fixes https://github.com/ipfs/testground/issues/637

### DIFF
--- a/sdk/runtime/runner.go
+++ b/sdk/runtime/runner.go
@@ -1,6 +1,7 @@
 package runtime
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"net"
@@ -12,7 +13,9 @@ import (
 	"net/http"
 	_ "net/http/pprof"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/prometheus/client_golang/prometheus/push"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


sdk/runtime/runner.go
```diff
@@ -20,11 +21,33 @@ func Invoke(tc func(*RunEnv) error) {
...
 	go func() {
 		pushInterval := 5 * time.Second
+
+		// Wait until the pushgateway is ready.
+		runenv.RecordMessage("Waiting for pushgateway to become accessible.")
+		var resbuf []byte
+		for {
+			select {
+			case <-time.After(pushInterval):
+				resp, err := http.Get("http://prometheus-pushgateway:9091/-/ready")
```
coryschwartz: I *definitely* want to get rid of the hard-coded urls everywhere. Should turn this into an env variable, like the REDIS_HOST, I think.

raulk: Yes, go for it.

* `cluster:k8s` and `local:docker` runners would inject `prometheus-pushgateway:9091`
* `local:exec` should inject `localhost:9091`.

raulk: Oh, nice. I hadn't realised that pushgateway had a `ready` endpoint.

raulk: How does this interact with the defer below, which actually forces a final push? We should probably capture the error there and make it less scary, with a nicer message like: `metrics push failed during instance shutdown; prometheus metrics from this instance may not be available; err: %s`

coryschwartz: it does print the final message " error during end metric push...", right now. But of course that could be changed.

coryschwartz: I created https://github.com/ipfs/testground/issues/711 to change this into a environment-delivered parameter.

PR #707, opened 2020-03-18
consistent path for testground outputs

nonsense:
Currently for `cluster:k8s` we create a tar with the `/outputs/RUN_ID/COMPOSITION/...` structure, whereas other runners create it with `/RUN_ID/COMPOSITION/...`.

This PR is fixing this.

### DIFF
--- a/pkg/runner/cluster_k8s.go
+++ b/pkg/runner/cluster_k8s.go
@@ -495,9 +495,12 @@ func (c *ClusterK8sRunner) CollectOutputs(ctx context.Context, input *api.Collec
 		VersionedParams(&v1.PodExecOptions{
 			Container: "collect-outputs",
 			Command: []string{
-				"sh",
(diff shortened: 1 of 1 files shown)


pkg/runner/cluster_k8s.go
```diff
@@ -496,17 +495,16 @@ func (c *ClusterK8sRunner) CollectOutputs(ctx context.Context, input *api.Collec
 		VersionedParams(&v1.PodExecOptions{
 			Container: "collect-outputs",
 			Command: []string{
-				"tar",
-				"-czf",
-				"-",
-				outputPath,
+				"sh",
```
coryschwartz: We can avoid the extra shell
`tar -C /outputs -czf - input.RunID`

nonsense: Fixed.

PR #709, opened 2020-03-18
fix(sidecar) ignore "route exists" errors when adding redis routes

Stebalien:
The alternative would be to add the new routes after removing the old ones. But that would leave us with no routes.


Stebalien:
Nevermind. I'm pretty sure this is wrong. We probably need to _replace_ the route.

PR #710, opened 2020-03-18
add build and push step timers

nonsense:
These 2 steps are rather important for our feedback loop, so it'd be nice to log how long they take and build intuition in case something is wrong, once we start building more complex compositions.

### DIFF
--- a/pkg/build/golang/docker.go
+++ b/pkg/build/golang/docker.go
@@ -223,10 +223,14 @@ func (b *DockerGoBuilder) Build(ctx context.Context, in *api.BuildInput, output 
 		BuildOpts: &opts,
 	}
 
+	buildStart := time.Now()
+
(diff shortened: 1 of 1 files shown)


pkg/build/golang/docker.go
```diff
@@ -223,11 +223,15 @@ func (b *DockerGoBuilder) Build(ctx context.Context, in *api.BuildInput, output
...
 	}
 
+	buildStart := time.Now()
+
 	err = docker.BuildImage(ctx, cli, &imageOpts)
 	if err != nil {
 		return nil, err
 	}
 
+	log.Debugw("build completed", "took", time.Since(buildStart))
```
raulk: I think these should be `Infow`. They're not overwhelming or repetitive, and very valuable.

nonsense: Changed in https://github.com/ipfs/testground/pull/714

PR #712, opened 2020-03-18
harden redis and sync service.

raulk:
Fixes https://github.com/ipfs/testground/issues/708.

* increase default sync service redis pool size from `runtime.NumCPU()*10` (default) to `8192` to prevent starvation.
    * Starvation is very probable in the sidecar, given that we use the XREAD primitive, which blocks a connection.
    * The sidecar is especially exposed to this, as it is a single process that accumulates a lot of `XREAD` blocked conns to receive network change requests from instances.
    * `8192` is a suitable value because on `local:docker`, we won't run as many containers, and on `cluster:k8s`, sidecars only handle containers colocated on the node.
* increase ulimits on infra containers to 1M open file descriptors.
* expose pprof port of sidecar; enable GC traces for debuggability.
* upgrade to go-redis v7.2.0.

additionally: harden throttling logic in local:docker runner.

### DIFF
--- a/pkg/runner/local_docker.go
+++ b/pkg/runner/local_docker.go
@@ -519,7 +519,7 @@ func (r *LocalDockerRunner) Run(ctx context.Context, input *api.RunInput, ow io.
 
 	log.Infow("starting containers", "count", len(containers))
 
-	g, _ := errgroup.WithContext(ctx)
+	g, gctx := errgroup.WithContext(ctx)
 	for _, id := range containers {
 		id := id
 		f := func() error {
@@ -532,8 +532,9 @@ func (r *LocalDockerRunner) Run(ctx context.Context, input *api.RunInput, ow io.
 			if err == nil {
 				log.Debugw("started container", "id", id)
 				select {
-				case started <- id:
+				case <-gctx.Done():
 				default:
+					started <- id
 				}
 			}
 			return err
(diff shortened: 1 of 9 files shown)


raulk:
@aschmahmann @jacobheun This should fix all your woes with the local:docker runner, the sync service, and the sidecar.

pkg/runner/local_docker.go
```diff
@@ -525,7 +531,10 @@ func (r *LocalDockerRunner) Run(ctx context.Context, input *api.RunInput, ow io.
 			err := cli.ContainerStart(ctx, id, types.ContainerStartOptions{})
 			if err == nil {
 				log.Debugw("started container", "id", id)
-				started <- id
+				select {
+				case started <- id:
+				default:
```
Stebalien: panic?

raulk: I don't think so, this is to mitigate a potential race between a `ContainerStart`s that were in progress, and have _just_ returned in success, just after the context was cancelled and we closed the `started` channel.

raulk: Basically we want to discard this container ID because it would anyway be deleted by `deleteContainers`, which deletes all _created_ containers.

Stebalien: Can we select on the context instead?

raulk: I guess if the select is on the context returned by errgroup.WithContext(), yes. I'll do that.

PR #718, opened 2020-03-19
remove initRunner. add initPool

nonsense:
Fixes https://github.com/ipfs/testground/issues/680

---

This PR is replacing `initRunner` with `initPool` - the idea is that we have a global pool of connections to the Kubernetes API.

I moved the pod resource requirements away from the global state in the Kubernetes runner, because we should be able to supply different resource requirements per instance for various testplans.


nonsense:
> Looks better. Approving with a question -
> 
> Is there any reason not to always initPool in init()? you could opportunistically initialize if defaultKubernetesConfig is properly filled, and avoid hving to try the init in every method.

`init()` is going to run for all testground instances, whereas `initPool` will lazy-init only if someone tries to use Kubernetes.

PR #722, opened 2020-03-19
Enhance the DHT test

aschmahmann:


raulk:
I'm merging this branch to master, so I can migrate this test plan to sync v2. In the next week, project-specific test plans will spin off outside this repo, and the upcoming testground org, into dedicated repositories in the organisations of the projects they test (in this case, I'm envisioning a libp2p/test-plans repo).

sdk/runtime/runner.go
```diff
@@ -45,7 +45,14 @@ func Invoke(tc func(*RunEnv) error) {
 	ctx, cancel := context.WithCancel(context.Background())
 
 	setupHTTPListener(runenv)
-	metricsDoneCh := setupMetrics(ctx, runenv)
+	pushGatewayEnabled := false
```
nonsense, line 48: Raising this to attention of the team, in case anyone is still using it - @coryschwartz @raulk

pkg/runner/cluster_k8s.go
```diff
@@ -700,7 +700,7 @@ func (c *ClusterK8sRunner) monitorTestplanRunState(ctx context.Context, ow *rpc.
 		default:
 		}
 
-		if time.Since(start) > 10*time.Minute {
+		if time.Since(start) > 100*time.Minute {
```
aschmahmann, line 703: @raulk @nonsense this was a change I had made since it seemed useful at the time. Worth keeping it like this until we have a config option for it?

sdk/runtime/runenv.go
```diff
@@ -301,6 +301,21 @@ func (re *RunParams) IntParam(name string) int {
...
+		return -1.0
+	}
+
+	f, err := strconv.ParseFloat(v, 32)
+	if err != nil {
+		panic(err)
+	}
+	return f
+}
+
```
aschmahmann, line 318: @raulk @nonsense a heads up that this was added in. I'm not currently using it, but every so often I think "maybe I want a float parameter"... happy to remove it if nobody wants it.

sdk/runtime/runner.go
aschmahmann: yep, this means the gateway is currently disabled, so we should either remove this or make it configurable.

pkg/runner/cluster_k8s.go
nonsense: I will add a config in `.env.toml` for it.

vyzo: we want this too! We want to run tests for longer than 10m.

PR #725, opened 2020-03-20
disable transparent huge pages support in kernel

nonsense:
This is disabling THP, as we are running Redis on one of these hosts:

```
1:M 20 Mar 2020 12:55:31.928 # WARNING you have Transparent Huge Pages (THP) support enabled in your kernel. This will create latency and memory usage issues with Redis. To fix this issue run the command 'echo never > /sys/kernel/mm/transparent_hugepage/enabled' as root, and add it to your /etc/rc.local in order to retain the setting after a reboot. Redis must be restarted after THP is disabled.
```

### DIFF
--- a/infra/k8s/kops-weave/dummy.yml
+++ b/infra/k8s/kops-weave/dummy.yml
@@ -16,8 +16,18 @@ spec:
       labels:
         name: dummy
     spec:
+      volumes:
+      - name: host-sys
+        hostPath:
+          path: /sys
+      initContainers:
+      - name: disable-thp
+        image: busybox
+        volumeMounts:
+          - name: host-sys
+            mountPath: /host-sys
+        command: ["sh", "-c", "echo never >/host-sys/kernel/mm/transparent_hugepage/enabled"]
       containers:
       - name: dummy
         command: ["/bin/sleep", "3650d"]
         image: governmentpaas/curl-ssl
-


Robmat05:
This PR relates to #727 in which we ran the benchmarks in docker. This ticket involves scaling this process using K8s

nonsense:
> Do we not need echo never > /sys/kernel/mm/transparent_hugepage/defrag too?

Redis is not complaining about this one, as far as I can tell. If it comes up, we can add it.

raulk:
@nonsense we'll be better served to do it now. If it does manifest in the future, it will burn a lot of time to debug and pinpoint, and we'll curse this exact moment.

nonsense:
> @nonsense we'll be better served to do it now. If it does manifest in the future, it will burn a lot of time to debug and pinpoint, and we'll curse this exact moment.

The way I read this flag, it should not have any effect, when you have disabled THP.

It is not mentioned in the warning or in most tutorials I find about configuring the kernel for Redis.

Why do you suggest we change it? Could you point me to any docs that talk about both these values? I really don't want to be tuning params without understanding what we are tuning.

raulk:
https://github.com/docker-library/redis/issues/55#issuecomment-368346882
https://github.com/docker-library/redis/issues/55#issuecomment-317584715

Issue #325, opened 2020-01-07
TestSidecar seems to be flaky


nonsense:
#### Version information:
```
* 1e8bf85 - (HEAD -> master, origin/master) codecov set informational flag to true (#324) (17 minutes ago) <Anton Evangelatov>
```

#### Description:

Test failed on one Travis runner and passed on another - for a PR that has not touched any sidecar code - https://github.com/ipfs/testground/pull/275 - and rebased on master on `1e8bf85`

Will report if I see a similar test fail again.

```
=== RUN   TestSidecar
2020-01-07T16:34:23.152Z	INFO	client/client.go:31	testground client initialized	{"addr": "127.0.0.1:38565"}
2020-01-07T16:34:23.152Z	INFO	server/server.go:81	daemon listening	{"addr": "127.0.0.1:38565"}
Step 1/25 : ARG GO_VERSION=1.13.4
Step 2/25 : FROM golang:${GO_VERSION}-buster
 ---> ed081345a3da
Step 3/25 : ARG GO_IPFS_VERSION
 ---> Using cache
 ---> 53fc7df0f44b
Step 4/25 : ARG TESTPLAN_EXEC_PKG
 ---> Using cache
 ---> 6b863b60478d
Step 5/25 : ARG GO_PROXY=direct
 ---> Using cache
 ---> 6b1b152d21cc
Step 6/25 : ENV TESTPLAN_EXEC_PKG ${TESTPLAN_EXEC_PKG}
 ---> Using cache
 ---> 668cff757454
Step 7/25 : ENV PLAN_DIR /plan/
 ---> Using cache
 ---> ec2b6e14b05c
Step 8/25 : RUN if [ -n "${GO_IPFS_VERSION}" ]; then echo Install IPFS ${GO_IPFS_VERSION} && cd /tmp && wget https://dist.ipfs.io/go-ipfs/v${GO_IPFS_VERSION}/go-ipfs_v${GO_IPFS_VERSION}_linux-amd64.tar.gz && tar xf go-ipfs_v${GO_IPFS_VERSION}_linux-amd64.tar.gz; fi
 ---> Using cache
 ---> bbf0d83c8053
Step 9/25 : RUN touch /tmp/delete_me
 ---> Using cache
 ---> 96008c84c00d
Step 10/25 : COPY /plan/go.mod ${PLAN_DIR}
 ---> 2c1af6d4f88f
Step 11/25 : COPY /sdk/sync/go.mod /sdk/sync/go.mod
[...]

PR #731, opened 2020-03-23
disable TestSidecar

nonsense:
`TestSidecar` is flaky, we have a bug issue for it for a while: https://github.com/ipfs/testground/issues/325


Robmat05:
test disabled. Waiting on #325 to become prioritized into later sprint.

PR #733, opened 2020-03-24
local exec redis exporter

coryschwartz:
This is the local:exec equivilent for https://github.com/ipfs/testground/pull/730

With this change, running the local exec runner will be more difficult to use, requiring more third party software to be in the PATH, installed by the user somehow.

From an ease-of-use perspective, we could download third party software for the user.  It would be possible to download grafana, redis, redis_exporter, prometheus, and prometheus_pushgateway into the user's home directory under `~/.testground/thridparty` or some similar location, but this presents its own set of problems to detect the OS and download the appropriate package.

I'm including an entry in the README to point users toward where to download this software on their own.

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -141,9 +141,13 @@ You should see a bunch of logs that describe the steps of the test, from:
 
 In order to run the tests with the local:exec runner, there are a few things that must be taken care of first.
 1. install required test software
-  * redis (redis.io)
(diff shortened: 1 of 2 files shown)


Robmat05:
per @coryschwartz #734 is a superior solution to this PR, so this solution has been unmerged and will be discarded/tabled

coryschwartz:
Closed in favor of the other solution

PR #734, opened 2020-03-24
DRY out the healthcheck api; use containers for local:exec dependencies; wire context to healthchecks

coryschwartz:
Fix: https://github.com/ipfs/testground/issues/720

Rather than separate RunChecks() and RunFixes, I have gone for just RunChecks with fix bool. This is so I can copy the errcheck code already in the local exec runner.

(updated from a comment I made later on): 

A list of some of the things going on here to look at during review.

- change in the healthcheck API which uses Checker and Fixer. This is similar, but not exactly what was suggested in https://github.com/ipfs/testground/issues/720

```
Helper.Enlist("name", Checker, Fixer)
Helper.RunChecks(context, true/false)
```

- Common operations can use the same Helper/Fixer constructor with different parameters.

- The container helpers and fixers are complex to provide enough flexibility to give different options... if you have any suggestions about how to make this simpler, I am up for it. I need  a second pair of eyes to tell me what looks confusing to them. I suspect those could be made simpler.

- Containers have explicit port number exposure into the default host network namespace.

- local:docker and local:exec healthchecks are mostly identical.

- the local:exec runner uses docker containers for everything but the runner.

- The sidecar is just another container to be built. No need to `make docker-ipfs-testground`

- ~~concurrent docker builds. No more waiting for docker images to download one at a time.~~
    - sequential builds so healthcheck dependencies can be easily managed.

Better than: https://github.com/ipfs/testground/pull/733
No more healthchecks like this: https://github.com/ipfs/testground/pull/730/
Closes: https://github.com/ipfs/testground/issues/720 and https://github.com/ipfs/testground/issues/660


pkg/runner/healthcheck_common.go
```diff
@@ -0,0 +1,79 @@
...
+	RunChecks(ctx context.Context, fix bool)
+}
+
+type toDoElement struct {
+	Name    string
+	Checker Checker
+	Fixer   Fixer
+}
+
+type ErrgroupHealthcheckHelper struct {
```
coryschwartz: I thought there might be some conditions under which concurrent fixes might be inappropriate so there might be multiple Healthchecked interface implementations. In this one, the implementation of RunChecks closely resembles the change to local_exec runner. 
See https://github.com/ipfs/testground/pull/703 and  https://github.com/ipfs/testground/pull/733
My intention is to move that logic here, easily convert the local exec runner to use this.

I suspect the docker runner can use this helper also

coryschwartz:
What we decided in the testground weekly meeting today is that the local:exec runner can use docker for thirdparty infra.

This would require knowing (or at least inspecting) the port assignments for indra containers, but the corrective action can be the same regardless of which local runner is used. -- ensureContainer.

coryschwartz:
I have added Checker and Fixer for docker containers.

The checker and fixer are based on methods from the local_docker, and those can be removed from there once this is swapped out. 

Tomorrow I should be able to get a working implementation I think for both local runners.

pkg/docker/container.go
```diff
@@ -72,7 +72,7 @@ func EnsureContainer(ctx context.Context, log *zap.SugaredLogger, cli *client.Cl
 
 		err := cli.ContainerStart(ctx, ci.ID, types.ContainerStartOptions{})
 		if err != nil {
-			log.Errorw("starting container failed", "container_id", container.ID)
+			log.Errorf("starting container failed", "container_id %s: %v", ci.ID, err)
```
coryschwartz: Not related to this change, but when I encountered this, `container.ID` caused a nil pointer panic.

pkg/runner/local_exec.go
```diff
@@ -283,3 +180,46 @@ func (*LocalExecutableRunner) ConfigType() reflect.Type {
 func (*LocalExecutableRunner) CompatibleBuilders() []string {
 	return []string{"exec:go"}
 }
+
```
coryschwartz: I thought there was a ticket to implement TerminateAll on local:exec somewhere, and wasn't done because the errgroup generally killed subprocesses anyway. Now, there are docker containers to take care of, so lets do it.

pkg/runner/healthcheck_common.go
```diff
@@ -0,0 +1,227 @@
...
+		return ci.State.Running
+
+	}
+}
+
+// DefaultContainerFixer returns a Fixer, a method which when executed will ensure the named
+// container exists with some default paramaters which are appropriate for infra containers.
+// Unless containers require special consideration, this should be considered the sensible default
+// fixer for docker containers.
+func DefaultContainerFixer(ctx context.Context, log *zap.SugaredLogger, cli *client.Client, containerName string, imageName string, networkID string, portSpecs []string, pull bool, customHostConfig *container.HostConfig, cmds ...string) Fixer {
```
coryschwartz: This is different than the old `ensureInfraContainer` in a couple of ways.
1. It allows you to pass custom build options so it can be used with the sidecar.
2. the port mapping is explicit, rather than random, so containers have predictable port numbers in the default netns and can therefore be used by the local:exec runner.

coryschwartz:
A list of some of the things going on here to look at during review.

- change in the healthcheck API which uses Checker and Fixer. This is similar, but not exactly what was suggested in https://github.com/ipfs/testground/issues/720

```
Helper.Enlist("name", Checker, Fixer)
Helper.RunChecks(context, true/false)
```

- Common operations can use the same Helper/Fixer constructor with different parameters.

- The container helpers and fixers are complex to provide enough flexibility to give different options... if you have any suggestions about how to make this simpler, I am up for it. I need  a second pair of eyes to tell me what looks confusing to them. I suspect those could be made simpler.

- Containers have explicit port number exposure into the default host network namespace.

- local:docker and local:exec healthchecks are mostly identical.

- the local:exec runner uses docker containers for everything but the runner.

- The sidecar is just another container to be built. No need to `make docker-ipfs-testground`

- concurrent docker builds. No more waiting for docker images to download one at a time.


Better than: https://github.com/ipfs/testground/pull/733
No more healthchecks like this: https://github.com/ipfs/testground/pull/730/
Closes: https://github.com/ipfs/testground/issues/720 and https://github.com/ipfs/testground/issues/660

pkg/docker/container.go
nonsense: `func (s *SugaredLogger) Errorf(template string, args ...interface{}) {`

`Errorf` is taking template as a first argument and then args after that. If `container.ID` is nil, maybe just use `opts.ContainerName` in the `log.Errorw` or fix the `log.Errorf` arguments.

coryschwartz:
By the way, there is a race condition currently with concurrent builds, which is the cause for the build failures in CI. I merged this with the logging changes in https://github.com/ipfs/testground/pull/756 and the race condition appears to have gone. There isn't really a dependency between the two PRs, but merge that one before this one to prevent the race condition the CI is currently complaining about.

pkg/docker/container.go
coryschwartz: Changed from ci.ID to opts.ContainerName.

.travis.yml
```diff
@@ -18,7 +18,6 @@ go:
 
 services:
   - docker
-  - redis-server
```
raulk: ?

.travis.yml
```diff
@@ -38,11 +37,9 @@ jobs:
       script:
         - make tidy
         - git diff --exit-code
-        - go mod download
```
raulk: ?

.travis.yml
```diff
@@ -38,11 +37,9 @@ jobs:
...
         - echo 'travis_fold:start:go.mod.graph' && echo 'go mod graph' && go mod graph && echo 'travis_fold:end:go.mod.graph'
         - pushd .. && go get github.com/golangci/golangci-lint/cmd/golangci-lint@v1.23.6 && popd
         - make lint           # run a bunch of code checkers/linters in parallel
-        - make test-build     # build the project
```
raulk: ?

pkg/docker/container.go
```diff
@@ -72,7 +72,7 @@ func EnsureContainer(ctx context.Context, log *zap.SugaredLogger, cli *client.Cl
 
 		err := cli.ContainerStart(ctx, ci.ID, types.ContainerStartOptions{})
 		if err != nil {
-			log.Errorw("starting container failed", "container_id", container.ID)
+			log.Errorf("starting container failed", "container_id %s: %v", opts.ContainerName, err)
```
raulk: I don't think this will have the right effect. I'd suggest you keep `Errorw` and add an `"error", err` tuple.

pkg/runner/healthcheck_common.go
```diff
@@ -0,0 +1,256 @@
...
+
+func (hh *ErrgroupHealthcheckHelper) RunChecks(ctx context.Context, fix bool) error {
+	eg, _ := errgroup.WithContext(ctx)
+	hh.report.Checks = make([]api.HealthcheckItem, len(hh.toDo))
+	hh.report.Fixes = make([]api.HealthcheckItem, len(hh.toDo))
+
+	for i, li := range hh.toDo {
+		i, li := i, li
+		hcp := *li
+		eg.Go(func() error {
```
raulk: Parallel checks won't work. Some depend on others, e.g. the control network being available, so that we can start the sidecar, redis, etc. Let's run them sequentially, in the order they were enlisted. Parallelism won't buy us much anyway; the checks should be relatively lightweight.

raulk: To express these dependencies, we might need to add a `dependsOn` parameter ([]string) when enlisting a check.

pkg/runner/healthcheck_common.go
```diff
@@ -0,0 +1,256 @@
...
+	"github.com/docker/go-connections/nat"
+	"github.com/docker/go-units"
+	"github.com/ipfs/testground/pkg/api"
+	"github.com/ipfs/testground/pkg/docker"
+
+	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
+)
+
+type Checker func() bool
```
raulk: The check can error out, which does not denote that the check failed, but that it was inoperative altogether. I think this function needs to return `(bool, error)`.

pkg/runner/healthcheck_common.go
```diff
@@ -0,0 +1,256 @@
...
+type Fixer func() error
```
raulk: Same, just for consistency with the `Checker`, I'd make this a `(bool, error)`, where the bool indicates if the fix succeeded.

pkg/runner/healthcheck_common.go
```diff
@@ -0,0 +1,256 @@
...
+	return func() error {
+		_, _, err := docker.EnsureContainer(ctx, log, cli, &ensure)
+		return err
+	}
+}
+
+// CustomContainerFixer returns a Fixer, a method which when executed will ensure the named
+// container exists. Unlike the DefaultContainerFixer, a custom image may be built for the
+// container.
+func CustomContainerFixer(ctx context.Context, log *zap.SugaredLogger, cli *client.Client, buildCtx string, containerName string, imageName string, networkID string, portSpecs []string, pull bool, customHostConfig *container.HostConfig, cmds ...string) Fixer {
```
raulk: Such long argument lists can get difficult to work with. Can we encapsulate these in a struct, please? Same for `DefaultContainerFixer`.

pkg/runner/healthcheck_common.go
```diff
@@ -0,0 +1,256 @@
...
+		if err != nil {
+			return false
+		}
+		return fi.IsDir()
+	}
+}
+
+// DirExistsFixer returns a Fixer, a method which when executed will create a directory and
+// any parent directories as appropriate.
+func DirExistsFixer(path string) Fixer {
```
raulk: I really like the util functions. Nice!

.travis.yml
coryschwartz: Yep! Redis started by the both docker and local exec runners, so the tests will pass without redis-server running on the host.

.travis.yml
coryschwartz: The make test-build was to build the sidecar, which is no longer needed because the healthcheck will build it

pkg/runner/healthcheck_common.go
coryschwartz: That is a good point i hadn't thought of.

I was trying to speed up builds by enabling multiple-downloads, 

I hadn't encountered an error running it this way, but I can see that you are clearly right. The dependency graph is not addressed by starting all in parallel. Well, that's easy enough to change. I'll swap this out for a less complicated serial HealthcheckHelper.

pkg/runner/healthcheck_common.go
coryschwartz: Sure. This definitely grew more complex as I started adding to it. Can swap this out with a struct.

pkg/runner/healthcheck_common.go
coryschwartz: If only they were all as simple as this one.

.travis.yml
raulk: The local:exec runner starts the Redis server if it's available on the PATH of the host. You mean to say that the Redis server is always on the path, and all that the `redis-server` entry does here is start it?

.travis.yml
raulk: Still, I'd rather make the CI more obvious. If I had a crystal ball, I'm pretty sure it would show me us going crazy in the future because we don't know where the sidecar is being built.

pkg/docker/container.go
```diff
@@ -70,7 +70,7 @@ func EnsureContainer(ctx context.Context, ow *rpc.OutputWriter, cli *client.Clie
 
 		err := cli.ContainerStart(ctx, ci.ID, types.ContainerStartOptions{})
 		if err != nil {
-			log.Errorw("starting container failed", "container_id", container.ID)
+			log.Errorw("starting container failed", "ContainerName", opts.ContainerName, "error", err)
```
raulk: Fields on log statements elsewhere are snake-cased.

pkg/runner/healthcheck_common.go
raulk: @coryschwartz ?

pkg/runner/local_exec.go
```diff
@@ -284,3 +179,45 @@ func (*LocalExecutableRunner) ConfigType() reflect.Type {
...
+func (*LocalExecutableRunner) TerminateAll(ctx context.Context, ow *rpc.OutputWriter) error {
+	ow.Info("terminate local:docker requested")
```
raulk: `local:exec`.

pkg/runner/local_infra.go
```diff
@@ -0,0 +1,130 @@
+package runner
+
+import (
+	"context"
+	"github.com/docker/docker/client"
+	"github.com/ipfs/testground/pkg/rpc"
```
raulk: Import order, in groups:

- stdlib
- testground
- third parties

https://github.com/golang/go/wiki/CodeReviewComments#imports

pkg/runner/local_infra.go
```diff
@@ -0,0 +1,130 @@
...
+	"path/filepath"
+)
+
+func healthcheck_common_local_infra(hcHelper HealthcheckHelper, ctx context.Context, ow *rpc.OutputWriter, cli *client.Client, controlNetworkID string, srcdir string, workdir string) {
```
raulk: Should be upper camel cased.

pkg/runner/healthcheck_common.go
```diff
@@ -0,0 +1,283 @@
...
+	"os/exec"
+
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/api/types/network"
+	"github.com/docker/docker/client"
+	"github.com/docker/go-connections/nat"
+	"github.com/docker/go-units"
+	"github.com/ipfs/testground/pkg/api"
+	"github.com/ipfs/testground/pkg/docker"
+	"github.com/ipfs/testground/pkg/rpc"
```
raulk: Import order, in groups:

- stdlib
- testground
- third parties

https://github.com/golang/go/wiki/CodeReviewComments#imports

pkg/runner/healthcheck_common.go
```diff
@@ -0,0 +1,283 @@
...
+// Each runner may have required elements -- infrastructure, etc. which should be checked prior to
+// running plans. Individual checks are registered to the HealthcheckHelper using the Enlist()
+// method. With all of the checks enlisted, execute the checks, and optionally fixes, using the
+// RunChecks() method. The details of how the checks are performed is implementation specific.
+// Typically, the checker and fixer passed to the enlist method will be closures. These methods will
+// be called when RunChecks is executed.
+type HealthcheckHelper interface {
+	Enlist(name string, c Checker, f Fixer)
+	RunChecks(ctx context.Context, fix bool) error
+}
```
raulk: I don't think we'll ever need multiple implementations of the healthchecker interface. If down the pike we decide to parallelise, we can do so without an alternative impl. I'd rm this interface, to simplify.

pkg/runner/healthcheck_common.go
```diff
@@ -0,0 +1,283 @@
...
+func (hh *SequentialHealthcheckHelper) Enlist(name string, c Checker, f Fixer) {
+	hh.toDo = append(hh.toDo, &toDoElement{name, c, f})
+}
+
+func (hh *SequentialHealthcheckHelper) RunChecks(ctx context.Context, fix bool) error {
+	for _, li := range hh.toDo {
+		// Check succeeds.
+		succeed, err := li.Checker()
+		if err != nil {
+			return err
```
raulk: Should use `api.HealthcheckStatusAborted` -- it's for internal errors.

pkg/runner/healthcheck_common.go
```diff
@@ -0,0 +1,283 @@
...
+				Status:  api.HealthcheckStatusOK,
+				Message: fmt.Sprintf("%s: OK", li.Name),
+			})
+			continue
+		}
+		// Checker failed, Append the failure to the check report
+		hh.report.Checks = append(hh.report.Checks, api.HealthcheckItem{
+			Name:    li.Name,
+			Status:  api.HealthcheckStatusFailed,
+			Message: fmt.Sprintf("%s: FAILED. Fixing: %t", li.Name, fix),
```
raulk: We're losing information we output in the current version. This is a bit of a regression. The message is redundant with the Status. Above and here, you output OK/FAILED, and the name of the fix. All that information is already in the struct. The message field is intended for message that explain WHY a healthcheck failed.

coryschwartz:
@raulk it's not clear to me whether we need the Ulimit resources on containers that are just started rather than built. 

I've left it on the sidecar since it is a go 1.14 build, but the other are either just downloaded and run or have a simple configuration layer added. Feel free to add the Ulimit to the other containers.

I've simplified the complex Default/custom container fixer in favor of simpler container and image.

Fixers AND and OR allow for multiple Fixers to be executed.

I have also added  Checkers ANY and ALL to be used for combinational predicates and a negation checker NOT, although I'm not using those anywhere.

.travis.yml
raulk: Closing the loop here. The answer is that the local:exec runner now runs prerequisites (redis, pushgateway, prometheus, etc.) as docker containers.

.travis.yml
raulk: I am going to revert this.

.travis.yml
raulk: I am going to revert this. If the sidecar fails building, we want the build system to tell us.

pkg/api/healthchecker.go
```diff
@@ -28,6 +28,9 @@ var (
 	// HealthcheckStatusOmitted indicates that a healthcheck or a fix was not
 	// carried out due to previous errors.
 	HealthcheckStatusOmitted = HealthcheckStatus("omitted")
+	// HealthcheckStatusUnnecessary indicates that a check or fix was not
+	// needed.
+	HealthcheckStatusUnnecessary = HealthcheckStatus("unnecessary")
```
raulk, line 33: Added this for fixes that were not necessary because the check was successful.

pkg/docker/container.go
```diff
@@ -80,34 +90,32 @@ func EnsureContainer(ctx context.Context, ow *rpc.OutputWriter, cli *client.Clie
 
 	log.Infow("container not found; creating")
 
-	if opts.PullImageIfMissing {
-		out, err := cli.ImagePull(ctx, opts.ContainerConfig.Image, types.ImagePullOptions{})
+	switch opts.ImageStrategy {
```
raulk, line 93: I've refactored this. We _don't_ want two different fixers:
- BuildImageFixer
- ContainerStartFixer

The reason is that a healthcheck could fail because the container exists but it's not started (e.g. testground-sidecar or our prometheus). The way it was implemented before, the image build would've ran in that case too.

pkg/docker/image.go
```diff
@@ -84,3 +76,28 @@ func EnsureImage(ctx context.Context, ow *rpc.OutputWriter, client *client.Clien
 	}
 	return true, err
 }
+
+// FindImage looks for an image with name `name` in our local daemon.
+//
+// If found, it returns the image summary and true.
+// If absent, it returns a nil image summary, and false.
+// If an internal error occurs, it returns a nil image summary, false, and a non-nil error.
+func FindImage(ctx context.Context, ow *rpc.OutputWriter, client *client.Client, name string) (*types.ImageSummary, bool, error) {
```
raulk, line 85: Added this function to the Docker package, to then reuse/compose it.

pkg/engine/engine.go
```diff
@@ -501,7 +501,7 @@ func (e *Engine) DoTerminate(ctx context.Context, runner string, ow *rpc.OutputW
 		return err
 	}
 
-	ow.Infof("all jobs terminated on runner: ", runner)
+	ow.Infof("all jobs terminated on runner: %s", runner)
```
raulk, line 504: was a bug

pkg/runner/local_exec.go
raulk: I'll take care of this.

pkg/runner/local_exec.go
```diff
@@ -284,3 +178,45 @@ func (*LocalExecutableRunner) ConfigType() reflect.Type {
...
+
+	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	if err != nil {
+		return err
+	}
+
+	// Build two separate queries: one for infrastructure containers, another
+	// for test plan containers. The former, we match by container name. The
+	// latter, we match by the `testground.purpose` label, which we apply to all
+	// plan containers managed by testground label.
```
raulk, line 193: This is copy pasta.

pkg/runner/local_exec.go
```diff
@@ -284,3 +178,45 @@ func (*LocalExecutableRunner) ConfigType() reflect.Type {
...
+
+	// Build query for runner infrastructure containers.
+	infraOpts := types.ContainerListOptions{}
+	infraOpts.Filters = filters.NewArgs()
+	infraOpts.Filters.Add("name", "prometheus-pushgateway")
+	infraOpts.Filters.Add("name", "testground-goproxy")
```
raulk, line 199: The goproxy container does not belong to the runner, but to the builder.

pkg/runner/local_exec.go
raulk: Indeed. Since we use exec.CommandContext, the plan processes should be killed. But terminate commands are used to recover from broken states. To play it safe, we may want to find all processes started by the daemon process, and send them SIGKILL.

pkg/runner/local_docker.go
```diff
@@ -787,80 +452,6 @@ func newDataNetwork(ctx context.Context, cli *client.Client, rw *rpc.OutputWrite
...
-			Image:      "ipfs/testground:latest",
-			Entrypoint: []string{"testground"},
-			Cmd:        []string{"sidecar", "--runner", "docker", "--pprof"},
-			Env:        []string{"REDIS_HOST=testground-redis", "GODEBUG=gctrace=1"},
-		},
-		HostConfig: &container.HostConfig{
-			PublishAllPorts: true,
-			PortBindings:    nat.PortMap{"6060": []nat.PortBinding{{HostIP: "0.0.0.0", HostPort: "0"}}},
-			NetworkMode:     container.NetworkMode(controlNetworkID),
-			// To lookup namespaces. Can't use SandboxKey for some reason.
```
raulk, line 837: @coryschwartz when you copied this over to the fixer configuration, you removed comments. Let's not remove comments that are good, please.

pkg/runner/local_docker.go
```diff
@@ -787,80 +452,6 @@ func newDataNetwork(ctx context.Context, cli *client.Client, rw *rpc.OutputWrite
...
-			PidMode: "host",
-			// We need _both_ to actually get a network namespace handle.
-			// We may be able to drop sys_admin if we drop netlink
-			// sockets that we're not using.
-			CapAdd: []string{"NET_ADMIN", "SYS_ADMIN"},
-			// needed to talk to docker.
-			Mounts: []mount.Mount{{
-				Type:   mount.TypeBind,
-				Source: dockerSock,
```
raulk, line 846: You also hardcoded this to "/var/run/docker.sock". Not everybody runs Linux or the default docker configuration. Please be careful not to cause any regressions with your refactors, @coryschwartz!

pkg/runner/local_docker.go
```diff
@@ -83,371 +84,80 @@ type LocalDockerRunner struct {
...
+					{Name: "nofile", Hard: InfraMaxFilesUlimit, Soft: InfraMaxFilesUlimit},
+				},
+			},
+			RestartPolicy: container.RestartPolicy{
+				Name: "unless-stopped",
+			},
+		},
+		ImageStrategy: docker.ImageStrategyBuild,
+		BuildImageOpts: &docker.BuildImageOpts{
+			Name:     "testground-sidecar:latest",
```
coryschwartz, line 148: small bug, the image built is called testground-sidecar:latest, but the image the container looks for is called ipfs/testground. Changed to ipfs/testground for both.

PR #735, opened 2020-03-24
remote build for testplans

nonsense:
Addresses: https://github.com/ipfs/testground/issues/747

---

TODO:
- [x] deploy `testground daemon` on k8s
- [x] decouple testplans from testground source (https://github.com/ipfs/testground/issues/629). testplans should be independent from testground daemon. testground daemon shouldn’t know about all possible test plans.
- [x] fix testground daemon to client streaming - https://github.com/ipfs/testground/issues/306 - if we don't do that it becomes even more clumbersome to check logs in both daemon and client, as now daemon is on k8s
- [x] build docker images in k8s (currently with docker-in-docker (dind))
- [x] give access to `testground-daemon` to kubernetes (kubectl, etc.)
- [x] rebuild and redeploy script for `testground-daemon`
- [x] fix healthchecks - we no longer have `kops` context on k8s cluster. we have `kops` context locally, so this healthcheck has to be removed/updated.
- [ ] add goproxy to host with `testground-daemon` pod

NICE TO HAVE:
- [ ] dockerhub repo for images (ideally versioned)
- [ ] discover automatically `testground daemon` on k8s from local `testground client` with minimal config (for now `kubectl port-forward service/testground-daemon 8042:8042` does the trick)
- [ ] differential sync of testplan source code from testground client to testground daemon. maybe `ksync` (https://github.com/ksync/ksync) or?
---

Open questions:
- [ ] testground sdk?

---

As a side-effect this is also fixing https://github.com/ipfs/testground/issues/572


coryschwartz:
to summarize -- 
running the daemon as a kubernetes pod with host docker socket mounted.

This would permit the daemon to create networks, volumes, containers on the host.

Maybe we can go with a docker-in-docker approach.

I was able to get external networking using docker:dind image without creating containers on the host, although my tests were not very thorough.

raulk:
I'll incorporate the latest changes removing dind to #849, and then close this PR in favour of landing that one. Thanks!

raulk:
The changes in this PR have been merged into #849.

PR #737, opened 2020-03-25
move testground services to control network

nonsense:
See `docs/NETWORKING.md`

control network -> flannel
data network -> weave

Generally we should not add auxiliary services to the data network - it is used by the test plans.

### DIFF
--- a/infra/k8s/install.sh
+++ b/infra/k8s/install.sh
@@ -114,8 +114,9 @@ kubectl apply -f ./efs/rbac.yaml \
               -f $EFS_MANIFEST_SPEC
 
 # monitoring and redis.
-echo "installing helm infrastructure"
-helm install testground-infra ./testground-infra
+echo "Installing helm infrastructure"
+helm install --wait --timeout 2m testground-infra ./testground-infra
+sleep 10
 
 echo "Install Weave, CNI-Genie, s3bucket DaemonSet, Sidecar Daemonset..."
 echo
--- a/infra/k8s/testground-infra/values.yaml
+++ b/infra/k8s/testground-infra/values.yaml
@@ -17,6 +17,9 @@
 #   See https://github.com/helm/charts/blob/master/stable/prometheus-operator/README.md#L193
 #   for an explanation of this option.
 prometheus-operator:
+  podAnnotations: {
+    cni: "flannel"
+  }
   alertmanager:
     enabled: false
   grafana:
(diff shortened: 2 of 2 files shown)


infra/k8s/testground-infra/values.yaml
```diff
@@ -17,6 +17,9 @@
 #   See https://github.com/helm/charts/blob/master/stable/prometheus-operator/README.md#L193
 #   for an explanation of this option.
 prometheus-operator:
+  podAnnotations: {
+    cni: "flannel"
+  }
   alertmanager:
     enabled: false
   grafana:
```
nonsense: I don't see how to move Grafana to the control network, but at least test plans don't need access to it.

nonsense:
@coryschwartz alternatively we could install all these services prior to installing weave, and then we don't have through the trouble of specifying `pod annotations`.

infra/k8s/install.sh
```diff
@@ -114,8 +114,9 @@ kubectl apply -f ./efs/rbac.yaml \
               -f $EFS_MANIFEST_SPEC
 
 # monitoring and redis.
-echo "installing helm infrastructure"
-helm install testground-infra ./testground-infra
+echo "Installing helm infrastructure"
+helm install --wait --timeout 2m testground-infra ./testground-infra
```
nonsense: While this is fine and would create all the services on the control network (after all `weave` is not installed at that time yet), we have to consider the `healthchecks` and `fixes` and being able to restart/upgrade those services.

coryschwartz: The only reason why put this up here is weave-service-monitor.yaml. has a CRD which doesn't exist until the operator is installed (the serviceMonitor)

What are we actually waiting for with the timeouts? Does the timeout just ensure the services are created on the right network?

nonsense: The `timeout` and the `wait`, are waiting for the actual resources to be created and running, before we continue.

If we continue to the next step, `weave` gets installed prior to all these resources, and then when we don't have `podAnnotations` set on them, they end up on the `weave` network, which is wrong.

nonsense:
Merging this so that `master` works again.

We can think and improve the situation of pods without explicit pod annotations being attached to the data network in a follow up issue/PR.

coryschwartz:
Related:
https://github.com/ipfs/testground/pull/740 which accomplishes the same thing as this PR  plus redis value changes

PR #740, opened 2020-03-25
[bug]restore redis values

coryschwartz:
After merging https://github.com/ipfs/testground/pull/723 there were a few issues noticed.

This PR will restore redis values modifications that were lost when I moved redis into a subchart of the testground-infra chart.

### DIFF
--- a/infra/k8s/testground-infra/values.yaml
+++ b/infra/k8s/testground-infra/values.yaml
@@ -55,6 +55,7 @@ prometheus-operator:
         serviceMonitorNamespaceSelector:
           any: true
 
+
 # Changes from defaults:
 # * override "fullname" so it is resolves with http://prometheus-pushgateway
 #   This matches the behavior from the kops addon, but we don't need to do this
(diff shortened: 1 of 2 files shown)


infra/k8s/testground-infra/values.yaml
```diff
@@ -85,3 +92,26 @@ redis:
...
+  master:
+    extraFlags:
+    - "--maxclients 100000"
+    podAnnotations:
+       cni: flannel
+    resources:
+      requests:
+        memory: 13000Mi
+        cpu: 6500m
+  sysctlImage:
```
nonsense: I noticed that this was previously applied to `slave` pods, so it was essentially redundant change from my part.

Do you know if it is necessary now, specially when we have the `securityContext` at the top? We should look into the actual chart, and see if this is applied or not.

coryschwartz: According to the comments in the helm chart, a sufficiently high version of kubernetes makes this redundant, and I removed the sysctlImage in a later commit.

PR #742, opened 2020-03-25
engine: deduplicate group builds; perform only unique builds.

### LINKED ISSUE #741: de-duplicate builds when using compositions
## Problem

Each composition group triggers a build. This is wasteful because the source and output artifacts will be the same as long as the upstream dependencies are not being overridden.

Moreover, all builds run concurrently. So neither golang nor Docker are able to leverage caches.

## Solution

Canonicalise and deduplicate builds, so that only unique builds are run. Demultiplex the build artifacts onto the groups.

raulk:
Fixes #741.

cc @yusefnapora @aschmahmann

### DIFF
--- a/pkg/api/composition.go
+++ b/pkg/api/composition.go
@@ -3,6 +3,7 @@ package api
 import (
 	"fmt"
 	"math"
+	"strings"
 
 	"github.com/go-playground/validator/v10"
 )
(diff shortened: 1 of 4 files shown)


pkg/api/composition.go
```diff
@@ -107,6 +108,18 @@ type Build struct {
 	Dependencies Dependencies `toml:"dependencies" json:"dependencies"`
 }
 
+// BuildKey returns a composite key that identifies this build, suitable for
+// deduplication.
+func (b Build) BuildKey() string {
+	var sb strings.Builder
+	sb.WriteString(fmt.Sprintf("selectors=%s;", strings.Join(b.Selectors, ",")))
```
coryschwartz: Maybe these can be sorted? It may not be obvious that selectors ["foo", "bar"] is different than ["bar", "foo"]

raulk: Yeah! This and also the dependencies need to be sorted for a canonical key. Good catch!

PR #743, opened 2020-03-26
Refactor/remove thirdparty helm

coryschwartz:
Like https://github.com/ipfs/testground/pull/740 this PR addresses comments to https://github.com/ipfs/testground/pull/723 which is already merged.

In https://github.com/ipfs/testground/pull/723, thirdparty helm charts were added, which introduced a large number of files into source control.

This PR removes those files.

`update-testground-infra.sh` is intended to be an in-place update for changing values. Manually, through CD, gitops, etc. It would be okay just to remove this.


nonsense:
@coryschwartz I thought about what you said, that these charts could change without us knowing in the current setup - we could lock the versions with a `requirements.yaml` file and `dependencies` field - https://helm.sh/docs/topics/charts/#chart-dependencies

Then we get the same guarantee as the go mod and we only upgrade the charts when we decide, and obviously not at the time that a user runs the install script.

PR #753, opened 2020-03-27
do not query logs if many testplan instances

nonsense:
It doesn't make sense to query for logs from 10k pods at the end of a test run, to get the last 2-3 lines of logs, specially when we store them on EFS, and `collect` them at the end of the run.


pkg/runner/cluster_k8s.go
```diff
@@ -256,33 +256,35 @@ func (c *ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow io.W
 		return nil, err
 	}
 
-	var gg errgroup.Group
+	if input.TotalInstances <= 500 {
```
nonsense: Obviously this value is up for discussion, or we should flat-out remove this part.

coryschwartz: I would give a :+1: for just removing this section as well.

PR #754, opened 2020-03-27
bump up resource requests for pushgateway

nonsense:
Not sure if this will be good enough to support 10k instances, but it is better than the defaults:

```
    Requests:
      cpu:        100m
```

### DIFF
--- a/infra/k8s/testground-infra/values.yaml
+++ b/infra/k8s/testground-infra/values.yaml
@@ -75,6 +75,10 @@ prometheus-operator:
       spec:
         serviceMonitorNamespaceSelector:
           any: true
+    resources:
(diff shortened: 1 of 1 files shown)


coryschwartz:
@nonsense does it make sense to you to have a few single-tenant hosts? You could one host for redis and redis_exporter, one host for prom/push/etc. and this way we can set the resources for these services fairly high without worry about bumping off plan pods.

nonsense:
@coryschwartz it definitely makes sense. At the same time Redis is not multithreaded so we are probably also wasting resources.

Then again I'd rather we error on the side of over-provisioning than under-provisioning.

Let's add this as a next step.

PR #756, opened 2020-03-27
daemon/client: mirror logs from daemon to client.

### LINKED ISSUE #306: testground daemon to client streaming is broken/weird
Top is client, bottom is daemon. I should be getting all structured output in the client. Instead, I'm getting what's arguably the most uninteresting part only -- the docker build logs.

image

@nonsense @hacdias let's chat about this.

raulk:
Fixes #306.

---

* This refactors the `TgWriter` to create and hold a zap.Logger that outputs to two `WriteSyncer`s: `stdout` and the HTTP response.

* Log statements written on the HTTP response make it over as `progress` messages.

* All operations on the daemon side have been refactored to log via the `TgWriter`, which will create the log event once and tee it into two writers. I discarded using [`zapcore.NewTee()`](https://godoc.org/go.uber.org/zap/zapcore#NewTee), because that would process the log event _twice_, and incur in more overhead than necessary.

## TODO

- [x] Rename `TgWriter` to something nicer.
- [x] Move it to another package: `rpc`. ~, maybe `tgio` or an existing package.~
- [x] Name the `TgWriter` variables consistently across the board.

### DIFF
--- a/main.go
+++ b/main.go
@@ -33,8 +33,6 @@ func main() {
 }
 
 func configureLogging(c *cli.Context) {
-	logging.TerminalMode()
-
 	// The LOG_LEVEL environment variable takes precedence.
 	if level := os.Getenv("LOG_LEVEL"); level != "" {
 		var l zapcore.Level
--- a/pkg/rpc/chunk.go
+++ b/pkg/rpc/chunk.go
@@ -1,0 +1,23 @@
+package rpc
+
+// ChunkType defines the type of a chunk.
+type ChunkType rune
+
+const (
+	ChunkTypeProgress ChunkType = 'p'
+	ChunkTypeResult   ChunkType = 'r'
+	ChunkTypeError    ChunkType = 'e'
+)
+
+// Chunk is a response chunk sent from the Testground daemon to the Testground
+// client. For a given request, clients should expect between 0 to `n`
+// `progress` chunks, and exactly 1 `result` or `error` chunk before EOF.
+type Chunk struct {
+	Type    ChunkType   `json:"t"` // progress or result or error
+	Payload interface{} `json:"p,omitempty"`
+	Error   *Error      `json:"e,omitempty"`
+}
+
+type Error struct {
+	Msg string `json:"m"`
+}
--- a/pkg/rpc/writer.go
+++ b/pkg/rpc/writer.go
@@ -1,4 +1,4 @@
-package tgwriter
+package rpc
 
 import (
 	"encoding/json"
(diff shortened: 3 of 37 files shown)


main.go
```diff
@@ -33,8 +33,6 @@ func main() {
 }
 
 func configureLogging(c *cli.Context) {
-	logging.TerminalMode()
```
coryschwartz: :+1:

pkg/rpc/writer.go
```diff
@@ -0,0 +1,136 @@
...
+	ow := &OutputWriter{
+		SugaredLogger:  logger.Sugar(),
+		out:            httpWriter,
+		progressWriter: progressWriter,
+	}
+
+	// we need to wire this back for the lock.
+	progressWriter.ow = ow
+	return ow
+}
```
raulk: Here's where most of the meat is.

main.go
```diff
@@ -33,8 +33,6 @@ func main() {
 }
 
 func configureLogging(c *cli.Context) {
```
coryschwartz: This method can be entirely removed, can it not?

coryschwartz:
Currently, if the client disconnects, then the test is canceled and the server issues an error that it cannot write to the client. If there is a large volume of data streaming to the client and overwhelms the network connection between client and server, I would expect the test to continue. By sending more data to the client, it becomes more likely that something like this could happen. But we can deal with that separately. This is an improvement.

main.go
coryschwartz: Nevermind. -vv on the daemon will send debug messages to the client.

coryschwartz:
Should docker build information (the ones printed using PipeOutput()) be switched to the OutputWriter as well?

pkg/rpc/chunk.go
```diff
@@ -0,0 +1,23 @@
+package rpc
```
coryschwartz: Why this package name?  I see RPC and I think "remote procedure call" not log message forwarding

raulk: This package contains objects to deal with incoming remote procedure calls. It’s used by the client and the server. The Chunk is part of the RPC interface, and the OutputWriter handles the output generated by the call.

raulk:
> Should docker build information (the ones printed using PipeOutput()) be switched to the OutputWriter as well? It does after all implement io.Writer already.

Yep, I’ll do that.

raulk:
> Currently, if the client disconnects, then the test is canceled and the server issues an error that it cannot write to the client. If there is a large volume of data streaming to the client and overwhelms the network connection between client and server, I would expect the test to continue. By sending more data to the client, it becomes more likely that something like this could happen. But we can deal with that separately. This is an improvement.

Which test are you referring to?

coryschwartz:
> > Which test are you referring to?

Sorry, I accidentally edited your comment rather than quote reply. You were asking me to clarify what I was talking regarding the client disconnections canceling tests.

I just mean if I am running a test plan and testground client disconnects from testground server, then the server cancels the run. It already does this, but I am thinking if server and client are not running on the same machine the network between them may be too slow to transfer messages to the client.  A message is frequently 100-200 bytes in length. If we have 10k nodes, sending a few messages per second, the network connectivity required between client and server could be pretty high.

Would be nice if the server could buffer the messages and then send them later when the client reconnects. But maybe we can implement that later when/if it ever becomes a problem.

PR #759, opened 2020-03-31
[wip] graceful shutdown 

coryschwartz:
This PR will  probably just be canceled, just want to put it up to demonstrate


coryschwartz:
Canceling, with pushgateway no longer with us.

PR #762, opened 2020-03-31
testground collect: add io.Writer for binary tgz outputs

nonsense:
Fixes regression in `testground collect` and also fixes https://github.com/ipfs/testground/issues/546.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -193,11 +193,7 @@ func ParseCollectResponse(r io.ReadCloser, file io.Writer) (CollectResponse, err
 			}
 
 			_, err = file.Write(m)
-			if err != nil {
-				return err
-			}
-			resp.Exists = true
-			return nil
+			return err
 		},
 		func(result interface{}) error {
 			resp.Exists = result.(bool)
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -159,7 +159,10 @@ func parseGeneric(r io.ReadCloser, fnProgress, fnBinary, fnResult func(interface
 			return fnResult(chunk.Payload)
 
 		case rpc.ChunkTypeBinary:
-			fnBinary(chunk.Payload)
+			err := fnBinary(chunk.Payload)
+			if err != nil {
+				return err
+			}
 
 		default:
 			return errors.New("unknown message type")
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -159,7 +159,7 @@ func parseGeneric(r io.ReadCloser, fnProgress, fnBinary, fnResult func(interface
 			return fnResult(chunk.Payload)
 
 		case rpc.ChunkTypeBinary:
-			return fnBinary(chunk.Payload)
+			fnBinary(chunk.Payload)
 
 		default:
 			return errors.New("unknown message type")
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -129,7 +129,7 @@ func (c *Client) Healthcheck(ctx context.Context, r *HealthcheckRequest) (io.Rea
 	return c.request(ctx, "POST", "/healthcheck", bytes.NewReader(body.Bytes()))
 }
 
-func parseGeneric(r io.ReadCloser, fnProgress, fnResult func(interface{}) error) error {
+func parseGeneric(r io.ReadCloser, fnProgress, fnBinary, fnResult func(interface{}) error) error {
 	var chunk rpc.Chunk
 	var once sync.Once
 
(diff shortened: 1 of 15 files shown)


pkg/daemon/outputs.go
```diff
@@ -27,10 +29,19 @@ func (srv *Daemon) outputsHandler(engine api.Engine) func(w http.ResponseWriter,
...
+		result := false
+		defer tgw.WriteResult(result)
+
+		var file bytes.Buffer
+		fw := bufio.NewWriter(&file)
+
+		err = engine.DoCollectOutputs(r.Context(), req.Runner, req.RunID, tgw, fw)
 		if err != nil {
-			log.Errorw("collect outputs error", "err", err.Error())
+			log.Warnw("collect outputs error", "err", err.Error())
```
nonsense: Demoting this to `Warnw` because this happens when we request a non-existing run id.

pkg/daemon/outputs.go
```diff
@@ -27,10 +29,21 @@ func (srv *Daemon) outputsHandler(engine api.Engine) func(w http.ResponseWriter,
...
-		err = engine.DoCollectOutputs(r.Context(), req.Runner, req.RunID, tgw)
+		result := false
+		defer func() {
+			tgw.WriteResult(result)
+		}()
+
+		var file bytes.Buffer
+		fw := bufio.NewWriter(&file)
+
+		err = engine.DoCollectOutputs(r.Context(), req.Runner, req.RunID, tgw, fw)
```
raulk: We shouldn't need this, nor the `io.Writer` parameter in `DoCollectOutputs`. You can create a `OutputWriter.BinaryWriter()` that returns an `io.Writer` that wires writes to the underlying `OutputWriter.WriteBinary`. See `OutputWriter.StdoutWriter()` and `OutputWriter.InfoWriter()` for inspiration.

pkg/client/client.go
```diff
@@ -174,12 +180,40 @@ func printProgress(progress interface{}) error {
...
+			m, err := base64.StdEncoding.DecodeString(payload.(string))
+			if err != nil {
+				return err
+			}
+
+			_, err = file.Write(m)
+			if err != nil {
+				return err
+			}
+			resp.Exists = true
```
raulk: Can probably do away with this, since we're setting `resp.Exists` based on what the server tells us in the Result.

pkg/daemon/list.go
```diff
@@ -20,7 +20,7 @@ func (srv *Daemon) listHandler(engine api.Engine) func(w http.ResponseWriter, r
 		plans := engine.TestCensus().ListPlans()
 		for _, tp := range plans {
 			for _, tc := range tp.TestCases {
-				_, err := tgw.Write([]byte(tp.Name + "/" + tc.Name + "\n"))
+				_, err := tgw.WriteProgress([]byte(tp.Name + "/" + tc.Name + "\n"))
```
raulk: All of this is technically part of the result, not progress. Ideally we'd create a `strings.Builder`, write to it, then write the full string as the Result via `WriteResult`.

pkg/rpc/writer.go
```diff
@@ -20,46 +21,56 @@ import (
...
 	logger := logging.NewLogger(writeSyncer).With(zap.String("req_id", r.Header.Get("X-Request-ID")))
 
 	ow := &OutputWriter{
-		SugaredLogger:  logger.Sugar(),
-		out:            httpWriter,
-		progressWriter: progressWriter,
+		SugaredLogger: logger.Sugar(),
+		out:           httpWriter,
+		pw:            progressWriter,
+		bw:            binaryWriter,
```
raulk: No need to store it, just create it and return it from `OutputWriter#BinaryWriter()`.

pkg/rpc/writer.go
```diff
@@ -122,16 +133,54 @@ func (ow *OutputWriter) StdoutWriter() io.Writer {
...
+		pw:            ow.pw,
 	}
 }
 
+func (ow *OutputWriter) WriteProgress(b []byte) (n int, err error) {
+	return ow.pw.Write(b)
+}
+
+func (ow *OutputWriter) WriteBinary(b []byte) (n int, err error) {
+	msg := Chunk{Type: ChunkTypeBinary, Payload: base64.StdEncoding.EncodeToString(b)}
```
raulk: No need to convert the `[]byte` to a base64 string manually. The `json` package already does this for us: https://golang.org/pkg/encoding/json/#Marshal.

pkg/client/client.go
nonsense: true.

pkg/daemon/list.go
nonsense: I see what you mean. Technically we were using `Write` previously, which was directly going to the `progressWriter`, so here we keep the same semantics.

How we define `progress` and `result` in the `list` cmd is up to us, so you could say that all of it is the Result, or individual Progress messages.

pkg/rpc/writer.go
nonsense: Does it really matter? If we use the binary writer multiple times, we would get the same writer right now, and not create multiple values... I am not sure if it makes any difference.

nonsense:
Merged so that folks get a working `collect` if they pick up `master` now. If you want we can address the other 2 comments in follow up PRs, but I think they are very minor and just a matter of opinion, not really bugs.

PR #763, opened 2020-03-31
miscellaneous fixes

raulk:
* fix incorrect messages.
* log when we're performing a healthcheck as part of a build or run.
* print kops output when healthcheck fails.

### DIFF
--- a/pkg/runner/cluster_k8s.go
+++ b/pkg/runner/cluster_k8s.go
@@ -295,9 +295,10 @@ func (*ClusterK8sRunner) ID() string {
 
 func (c *ClusterK8sRunner) healthcheckK8s() (k8sCheck api.HealthcheckItem) {
 	k8sCheck = api.HealthcheckItem{Name: "k8s", Status: api.HealthcheckStatusOK, Message: "k8s cluster is running"}
-	err := exec.Command("kops", "validate", "cluster").Run()
-	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/runner/cluster_k8s.go
```diff
@@ -295,9 +295,10 @@ func (*ClusterK8sRunner) ID() string {
 
 func (c *ClusterK8sRunner) healthcheckK8s() (k8sCheck api.HealthcheckItem) {
 	k8sCheck = api.HealthcheckItem{Name: "k8s", Status: api.HealthcheckStatusOK, Message: "k8s cluster is running"}
-	err := exec.Command("kops", "validate", "cluster").Run()
-	if err != nil {
-		k8sCheck = api.HealthcheckItem{Name: "k8s", Status: api.HealthcheckStatusFailed, Message: fmt.Sprintf("k8s cluster validation failed: %s", err)}
+	cmd := exec.Command("kops", "validate", "cluster")
```
nonsense: We need to get rid of this - once the `testground daemon` goes remote on Kubernetes, we no longer have access to `kops`, as `kops` is a local utility that creates a Kubernetes cluster, and not something that we have on the Kubernetes clusters.

raulk: @nonsense mind opening an issue to track that?

PR #769, opened 2020-04-01
collect cmd: fix when run as --collect

nonsense:
`collect` is still broken when called from `testground run ... --collect`.

### DIFF
--- a/cmd/collect.go
+++ b/cmd/collect.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io"
 	"os"
 
 	"github.com/ipfs/testground/pkg/client"
@@ -56,12 +55,21 @@ func collectCommand(c *cli.Context) error {
 		return err
 	}
 
-	req := &client.OutputsRequest{
-		Runner: runner,
-		RunID:  id,
+	err = collect(ctx, api, runner, id, output)
+	if err != nil {
+		return err
 	}
 
-	resp, err := api.CollectOutputs(ctx, req)
+	return nil
+}
+
+func collect(ctx context.Context, cl *client.Client, runner string, runid string, outputFile string) error {
+	req := &client.OutputsRequest{
+		Runner: runner,
+		RunID:  runid,
+	}
+
+	resp, err := cl.CollectOutputs(ctx, req)
 	if err != nil {
 		if err == context.Canceled {
 			return fmt.Errorf("interrupted")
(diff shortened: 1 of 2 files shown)


cmd/collect.go
```diff
@@ -56,12 +55,21 @@ func collectCommand(c *cli.Context) error {
 		return err
 	}
 
+	err = collect(ctx, api, runner, id, output)
```
nonsense, line 58: Extracting this previously didn't make sense as we were literally getting a file as the response from the API.

However now the contract has changed and we actually get a JSON response from the `collect` API endpoint, so makes sense to extract together with the parsing part.

cmd/collect.go
```diff
@@ -85,16 +93,10 @@ func collectCommand(c *cli.Context) error {
 	}
 
 	if !cr.Exists {
-		logging.S().Errorw("no such testplan run", "run_id", id, "runner", runner)
+		logging.S().Errorw("no such testplan run", "run_id", runid, "runner", runner)
 		return nil
 	}
 
-	_, err = io.Copy(file, resp)
```
nonsense, line 92: No-op, this should have been removed yesterday.

cmd/collect.go
```diff
@@ -85,16 +93,10 @@ func collectCommand(c *cli.Context) error {
 	}
 
 	if !cr.Exists {
-		logging.S().Errorw("no such testplan run", "run_id", id, "runner", runner)
+		logging.S().Errorw("no such testplan run", "run_id", runid, "runner", runner)
```
raulk, line 96: This will leave a 0kb file behind, no? Shouldn't we delete it here?

cmd/run.go
```diff
@@ -192,8 +191,8 @@ func doRun(c *cli.Context, comp *api.Composition) (err error) {
 	logging.S().Infof("finished run with ID: %s", rout.RunID)
 
 	// if the `collect` flag is not set, we are done, just return
-	collect := c.Bool("collect")
-	if !collect {
+	collectBool := c.Bool("collect")
```
raulk, line 194:
```suggestion
	collectOpt := c.Bool("collect")
```

cmd/run.go
```diff
@@ -202,27 +201,10 @@ func doRun(c *cli.Context, comp *api.Composition) (err error) {
...
-		}
-		return err
-	}
-	defer file.Close()
-
-	_, err = io.Copy(file, rc)
+	err = collect(ctx, cl, comp.Global.Runner, rout.RunID, collectFile)
 	if err != nil {
 		return err
 	}
```
raulk, line 207: Can just return the error now.

cmd/collect.go
nonsense: Added a delete on error.

PR #773, opened 2020-04-01
fix the baseline/subtree test case.

raulk:


plans/benchmarks/benchmarks.go
```diff
@@ -300,9 +316,12 @@ func SubtreeBench(runenv *runtime.RunEnv) error {
 				t := prometheus.NewTimer(tst.Summary)
 				b := <-ch
 				t.ObserveDuration()
-				if !bytes.Equal(tst.Data, b) {
+				if strings.Compare(*b, *tst.Data) != 0 {
 					return fmt.Errorf("received unexpected value")
 				}
+				if i%500 == 0 {
+					runenv.RecordMessage("received %d items", i)
```
nonsense, line 323: Please add test case id as well, which is ignored right now.

plans/benchmarks/benchmarks.go
```diff
@@ -284,14 +287,27 @@ func SubtreeBench(runenv *runtime.RunEnv) error {
...
+
 	case "receive":
 		runenv.RecordMessage("i am a subscriber")
 
 		// if we are receiving, wait for the publisher to be done.
 		<-watcher.Barrier(ctx, handoff, int64(1))
 
 		for _, tst := range tests {
-			ch := make(chan []byte, 1)
+			ch := make(chan *string, 100)
```
nonsense, line 310: Why 100? Ideally we don't want to consume more than necessary, and want to block the producer (in this case `watcher.Subscribe`.

I think 1 is fine.

PR #775, opened 2020-04-01
speed up install.sh

nonsense:
Considering https://github.com/ipfs/testground/pull/768 we no longer need the `helm wait`.

We are confident that the default network is flannel, so no need to wait for resources to come up in order to continue with next steps.

Redis is required for some of the next DaemonSets, but they are implemented with `initContainers` that wait for it, so we are good there.

### DIFF
--- a/infra/k8s/install.sh
+++ b/infra/k8s/install.sh
@@ -114,12 +114,11 @@ kubectl apply -f ./efs/rbac.yaml \
               -f $EFS_MANIFEST_SPEC
 
 # monitoring and redis.
-echo "installing helm infrastructure"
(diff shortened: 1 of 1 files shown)


nonsense:
The failure on travis has nothing to do with this PR and is addressed in another PR.

nonsense:
> If the CRD isn't created before weaver-service-monitor.yml is applied, there will be a failure for an unknown resource.

Wouldn't k8s to try to create it again after that Error? We also have `prometheus` failing the first time to start, due to missing config (which is created by the ConfigMap), but it starts up fine after retry?

PR #782, opened 2020-04-01
docker filters: apply regex filtering

coryschwartz:
Fixes https://github.com/ipfs/testground/issues/770

compare with https://github.com/ipfs/testground/pull/771 when reviewing

I have added regex filters to the EnsureContainer and EnsureVolume.

### DIFF
--- a/pkg/docker/volume.go
+++ b/pkg/docker/volume.go
@@ -37,9 +37,16 @@ func EnsureVolume(ctx context.Context, log *zap.SugaredLogger, cli *client.Clien
 		return nil, false, err
 	}
 
-	if len(volumes.Volumes) > 0 {
+	switch l := len(volumes.Volumes); l {
+	case 0:
+		break
+	case 1:
 		log.Info("found existing volume")
 		return volumes.Volumes[0], false, err
+	default:
+		err := fmt.Errorf("unexpected number of volumes returned by docker; expected 0 or 1: got: %d; while searching for: %s", l, opts.Name)
+		log.Error(err)
+		return nil, false, err
 	}
 
 	log.Infof("creating new docker volume")
--- a/pkg/docker/volume.go
+++ b/pkg/docker/volume.go
@@ -2,6 +2,7 @@ package docker
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/filters"
@@ -30,16 +31,15 @@ func EnsureVolume(ctx context.Context, log *zap.SugaredLogger, cli *client.Clien
 	log.Debug("checking state of volume")
 
 	// Check whether volume exists.
-	volumes, err := cli.VolumeList(ctx, filters.Args{})
+	exactName := fmt.Sprintf("^%s$", opts.Name)
+	volumes, err := cli.VolumeList(ctx, filters.NewArgs(filters.Arg("name", exactName)))
 	if err != nil {
 		return nil, false, err
 	}
 
-	for _, v := range volumes.Volumes {
-		if v.Name == opts.Name { // We found a match, volume exists.
-			log.Info("found existing volume")
-			return v, false, nil
-		}
+	if len(volumes.Volumes) > 0 {
+		log.Info("found existing volume")
+		return volumes.Volumes[0], false, err
 	}
 
 	log.Infof("creating new docker volume")
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 notifications:
   email: false
 
+dist: bionic
+
 language: go
 
 env:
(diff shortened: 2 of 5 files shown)


nonsense:
LGTM

nonsense:
@coryschwartz on a second thought, something must be wrong with this code, or with our tests, and I am not sure it is the tests. I see the following:

```
=== RUN   TestAbortedTestShouldFailDocker
...
Apr  2 10:31:53.989391	[34mINFO[0m	starting new container	{"req_id": "3cc8d957", "container_name": "testground-grafana", "id": "0f8ddc94506af991e62f3125072ea8787cc704c0b51da57ce6e337fd5c3f90ef"}
Apr  2 10:31:54.583323	[34mINFO[0m	started container	{"req_id": "3cc8d957", "container_name": "testground-grafana", "id": "0f8ddc94506af991e62f3125072ea8787cc704c0b51da57ce6e337fd5c3f90ef"}
...
Apr  2 10:32:59.897764	[33mWARN[0m	engine run error: healthcheck fixes failed; aborting:
Checks:
- grafana-container: failed; grafana container: non-existent
Fixes:
- grafana-container: failed; failed to create grafana container: Error response from daemon: Conflict. The container name "/testground-grafana" is already in use by container "0f8ddc94506af991e62f3125072ea8787cc704c0b51da57ce6e337fd5c3f90ef". You have to remove (or rename) that container to be able to reuse that name.
```

Looks like we creates a `grafana` container `0f8d` at `10:31:53` and we didn't detect it properly at `10:32:59` and tried to create it again (with the same name), which resulted in a `conflict`.

There are a few problems with that:
- [ ] tests don't clean up after they run, so containers stay behind (we know about this, and there is an issue)
- [ ] not sure why testground is not detecting that there is such a container already existing. (this happens not only for `grafana`, I didn't leave all the logs here for brevity)

coryschwartz:
I'll see if I can find out what's going on with that today, yeah looks like somomething is wrong. `make test` does pass on my machine -- will investigate today.

nonsense:
@coryschwartz tests also seem to be consistent for me locally, but Travis seems to be consistently failing.

If we merge this, we have to at least disable the broken test.

pkg/docker/volume.go
```diff
@@ -30,16 +31,15 @@ func EnsureVolume(ctx context.Context, log *zap.SugaredLogger, cli *client.Clien
...
 	}
 
-	for _, v := range volumes.Volumes {
-		if v.Name == opts.Name { // We found a match, volume exists.
-			log.Info("found existing volume")
-			return v, false, nil
-		}
+	if len(volumes.Volumes) > 0 {
+		log.Info("found existing volume")
+		return volumes.Volumes[0], false, err
```
raulk, line 42: Logically speaking, this would be sound if our length check was for == 1. But in the current form, it's unsound, because we allow for N volumes to be returned, yet assume the first one is the correct one. We should either:

1. add an assertion check that len == 1 and error if it's higher, or
2. be lenient as before.

It's a good practice to be defensive when dealing with third party libraries, especially Docker... 🙃

raulk:
My guess is that somewhere along history, Docker changed the filtering logic, and we are running a more lenient version on our development machines that matches against the container name **with** and **without** a leading slash, whereas CI is running a stricter version that matches against the actual container name.

On my machine:

```
⟩ docker version
Client: Docker Engine - Community
 Version:           19.03.8
⟩ docker inspect testground-redis -f '{{.Name}}'
/testground-redis
⟩ docker container ps --filter="name=^/testground-redis\$"
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS                     NAMES
bb55ac5d61e5        redis               "docker-entrypoint.s…"   40 minutes ago      Up 40 minutes       0.0.0.0:32802->6379/tcp   testground-redis
⟩ docker container ps --filter="name=^testground-redis\$"
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS                     NAMES
bb55ac5d61e5        redis               "docker-entrypoint.s…"   40 minutes ago      Up 40 minutes       0.0.0.0:32802->6379/tcp   testground-redis
```

Both match, which makes NO SENSE if the regex was being evaluated solely against the actual container name (with a leading slash). So there's some trickery introduced in newer Docker versions.

For the record, the Docker version in Travis is:

image

I'm going to update it as per https://docs.travis-ci.com/user/docker/#installing-a-newer-docker-version, and I'll get back to you.

Regardless, I will modify our matching logic to match against the slash, as that is more likely to work in all cases.

raulk:
The Travis instructions didn't work. I ended up with the same version: 
image

Probably because Ubuntu hasn't updated the docker-ce package in their repos, and Travis is using an old version of Ubuntu.

raulk:
This works on CI now.

PR #784, opened 2020-04-01
ulimits local:docker

coryschwartz:
Docker component of https://github.com/ipfs/testground/issues/781

apply higher ulimits to plan containers

### DIFF
--- a/pkg/conv/conversions.go
+++ b/pkg/conv/conversions.go
@@ -84,20 +84,21 @@ func ToEnvVar(input map[string]string) []v1.EnvVar {
 	return out
 }
 
-// ToUlimits takes a slice of strings formatted in a particular way. It returns a new slice
-// with the strings parsed as Ulimits such that they can be used by docker. If any of the ulimit
-// strings are improperly formatted, they will be ignored.
-// Each element in the input should be formatted in one of the following ways:
-// nofile=512:1024
-// nofile=1024
-// cpu=2:4
-// cpu=6
-func ToUlimits(input []string) (out []*units.Ulimit, err error) {
+// ToUlimits converts a slice of strings following the Docker ulimit format, to
+// the appropriate type. If parsing fails, this function shortcircuits and
+// returns an error.
+//
+// See
+// https://docs.docker.com/engine/reference/commandline/run/#set-ulimits-in-container---ulimit
+// for more info on format.
+func ToUlimits(input []string) ([]*units.Ulimit, error) {
+	out := make([]*units.Ulimit, 0, len(input))
 	for _, s := range input {
 		parsed, err := units.ParseUlimit(s)
-		if err == nil {
-			out = append(out, parsed)
+		if err != nil {
+			return nil, err
 		}
+		out = append(out, parsed)
 	}
-	return
+	return out, nil
 }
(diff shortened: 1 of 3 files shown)


pkg/runner/local_docker.go
```diff
@@ -56,6 +56,8 @@ type LocalDockerRunnerConfig struct {
 	// Background avoids tailing the output of containers, and displaying it as
 	// log messages (default: false).
 	Background bool `toml:"background"`
+	// number of open files
+	Ulimits []string `toml: "ulimits"`
```
coryschwartz, line 60: I am somewhat torn about having the option as a slice -- by making this option into a slice rather than just a string, maybe csv, this makes the command-line flag pretty awkward. I would recommend just using the toml file to specify the toml array.

pkg/runner/local_docker.go
```diff
@@ -548,6 +551,11 @@ func (r *LocalDockerRunner) Run(ctx context.Context, input *api.RunInput, ow *rp
 				},
 			}
 
+			ulimits, err := conv.ToUlimits(cfg.Ulimits)
```
coryschwartz, line 554: This applies the configured ulimits to plan contaienrs, but not for "infrastructure" containers, redis, etc. because those containers are created during healthcheck fix actions and the input configuration isn't passed there. I suspect we will want that to be configured by the same option, which will require passing the RunInput composition.

pkg/conv/conversions.go
```diff
@@ -82,3 +83,21 @@ func ToEnvVar(input map[string]string) []v1.EnvVar {
...
+// cpu=2:4
+// cpu=6
+func ToUlimits(input []string) (out []*units.Ulimit, err error) {
+	for _, s := range input {
+		parsed, err := units.ParseUlimit(s)
+		if err == nil {
+			out = append(out, parsed)
+		}
+	}
+	return
```
raulk, line 102: Naked returns are **dangerous**. This is not doing what you think it's doing. The assignment above (`parsed, err := units.ParseUlimit(s)`) is creating a new local `err` variable, not assigning to the return variable. Therefore, the returned error is always nil.

Also we should short-circuit if we detect an error, no reason why to continue processing the rest of the input.

pkg/runner/local_docker.go
```diff
@@ -56,6 +56,8 @@ type LocalDockerRunnerConfig struct {
 	// Background avoids tailing the output of containers, and displaying it as
 	// log messages (default: false).
 	Background bool `toml:"background"`
+	// number of open files
```
raulk, line 59: Godoc is inaccurate; ulimits restrict more things than just file descriptors.

pkg/runner/local_docker.go
```diff
@@ -64,6 +66,7 @@ var defaultConfig = LocalDockerRunnerConfig{
 	KeepContainers: false,
 	Unstarted:      false,
 	Background:     false,
+	Ulimits:        []string{""},
```
raulk, line 69: The intention was to set a high default limit (1Mi), as suggested in https://github.com/ipfs/testground/issues/781.

pkg/runner/local_docker.go
raulk: Infrastructure containers are not run-scoped; they have their own configuration.

raulk:
Waiting for CI.

PR #785, opened 2020-04-01
ulimits kubernetes using kops

coryschwartz:
Kubernetes component of https://github.com/ipfs/testground/issues/781

This took me longer than I had hoped to figure out how to do this, and this is an undocumented kops feature, so I want to give an explanation.

All of the yaml stanzas are serialized into structs, the docker configuration struct is here: https://godoc.org/k8s.io/kops/pkg/apis/kops#DockerConfig

As you DefaultUlimit takes a slice/yaml-list of strings which are formatted with the same function as other docker ulimit strings. This is familiar.: https://github.com/docker/go-units/blob/master/ulimit.go#L66

Pretty simple!

To verify that this is working, log into one of the kops hosts, and have a look at any of the docker containers and see the result:

```
# docker inspect bf24b493fb8b | grep -A5 Ulimit
            "Ulimits": [
                {
                    "Name": "nofile",
                    "Hard": 1048576,
                    "Soft": 1048576
                }
```

### DIFF
--- a/infra/k8s/cluster.yaml
+++ b/infra/k8s/cluster.yaml
@@ -12,6 +12,9 @@ spec:
   channel: stable
   cloudProvider: aws
   configBase: ${KOPS_STATE_STORE}/${NAME}
+  docker:
+    defaultUlimit:
+    - "nofile=1048576:1048576"
   etcdClusters:
   - cpuRequest: 200m
     etcdMembers:
--- a/infra/k8s/cluster.yaml
+++ b/infra/k8s/cluster.yaml
@@ -14,7 +14,7 @@ spec:
   configBase: ${KOPS_STATE_STORE}/${NAME}
   docker:
     defaultUlimit:
-    - "nofile=1048576:1048576"
+    - "nofile=2097152:2097152"
   etcdClusters:
   - cpuRequest: 200m
     etcdMembers:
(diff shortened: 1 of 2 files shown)


nonsense:
@coryschwartz we are getting the same values today in all containers, without this config.

I am not sure where this is setup by default in `kops` or in the underlying AMI.

infra/k8s/cluster.yaml
```diff
@@ -12,6 +12,9 @@ spec:
   channel: stable
   cloudProvider: aws
   configBase: ${KOPS_STATE_STORE}/${NAME}
+  docker:
+    defaultUlimit:
+    - "nofile=1048576:1048576"
```
nonsense, line 17: @coryschwartz maybe use a different value, just so that we can confirm we can actually change it.

Currently on previous clusters we have:
```
➜  testground git:(disable-pushgateway) k exec -it dummy-tt89z sh
/ # ulimit -a
-f: file size (blocks)             unlimited
-t: cpu time (seconds)             unlimited
-d: data seg size (kb)             unlimited
-s: stack size (kb)                8192
-c: core file size (blocks)        unlimited
-m: resident set size (kb)         unlimited
-l: locked memory (kb)             64
-p: processes                      1048576
-n: file descriptors               1048576
-v: address space (kb)             unlimited
-w: locks                          unlimited
-e: scheduling priority            0
-r: real-time priority             0
/ #
```

nonsense:
I really can't tell if we have the `ulimit` within containers right now due to this change, or because it used to be `1048576`.

PR #792, opened 2020-04-02
remove redis persistence

nonsense:


nonsense:
Marking this as blocked, as we don't really want to disable the probes. This was an attempt to understand what's going on with Redis when running gossipsub tests.

It is still not clear what the underlying issue is, but it doesn't appear to be the probes.

---

I still think we don't need to enable persistence.

infra/k8s/testground-infra/values.yaml
```diff
@@ -132,6 +132,12 @@ redis:
       - name: net.netfilter.nf_conntrack_max
         value: "100000"
   master:
+    persistence:
+      enabled: false
+    livenessProbe:
+      enabled: false
+    readinessProbe:
+      enabled: false
```
raulk, line 140: Looks like these probes were actually right. These adjustments should fix the root cause: https://github.com/ipfs/testground/pull/798 Then we can keep the probes enabled, and only merge the persistence disabling.

nonsense:
Closing as we already applied `persistence: false`, and we don't want to apply the rest of the changes.

PR #797, opened 2020-04-03
increase conntrack from 100k to 1mln

nonsense:
https://www.projectcalico.org/when-linux-conntrack-is-no-longer-your-friend/

https://morganwu277.github.io/2018/05/26/Solve-production-issue-of-nf-conntrack-table-full-dropping-packet/

---

```
Recommended size: CONNTRACK_MAX = RAMSIZE (in bytes) / 16384 / (ARCH / 32)
```

For c5.2xlarge 1mln is probably too much, but we should be fine with at least 300k.


nonsense:
Closing in favour of https://github.com/ipfs/testground/pull/798

PR #798, opened 2020-04-03
update linux networking stack configuration.

raulk:
After a lot of investigation, let me explain what I've done here, parameter by parameter.

**`net.core.somaxconn = 131072`**

Bumped up to 128Ki -- for a good binary value. Note: this is the number of incoming sockets awaiting to be accepted, not the maximum number of connections we can maintain. It's a misnomer...

**`net.netfilter.nf_conntrack_max = 1048576`**

Let's use binary values.

**`net.ipv4.tcp_max_syn_backlog = 131072`**

This is the maximum number of incoming connections in SYN state. I don't think it needs to be so high. I was going to suggest 64Ki, but it doesn't hurt us to have a higher value, since our workloads are very bursty. So I rounded up to the next meaningful binary value.

**`net.core.netdev_max_backlog = 524288`**

This is the number of frames that can be outstanding before being consumed from the networking device. We can afford for this to be higher.

**`net.core.rmem_max = 131072`**
**`net.core.wmem_max = 131072`**

Maximum read and write buffer sizes in bytes. Setting them to 128KiB. We are interested in supporting a large number of concurrent connections, but at the same time we want this number to be similar to the defaults found in the wild. Rationale: we want our test workloads to get realistic congestion control. A quick Google search reveals the defaults in Linux environments tend to be 128KiB.

**`net.ipv4.tcp_mem = 262144 524288 1048576`**

These are the low, pressure, and max watermarks for total memory usage the TCP stack -- in memory pages (not bytes). Each page is 4KiB, so our max is 4GiB.

**`net.ipv4.tcp_rmem = 4096 65536 131072`**
**`net.ipv4.tcp_wmem = 4096 65536 131072`**

These are the low, pressure, and max watermarks for buffers. We allow a max of 128KiB. Under memory pressure, buffers will be theoretically pruned to their lowest (4KiB), which would allow 1Mi minimally sized buffers, and because each connection users two buffers, we're looking at a max of 512Ki connections.

### DIFF
--- a/infra/k8s/cluster.yaml
+++ b/infra/k8s/cluster.yaml
@@ -49,6 +49,7 @@ spec:
       - net.ipv4.tcp_max_syn_backlog
       - net.ipv4.tcp_tw_recycle
       - net.ipv4.tcp_tw_reuse
+      - net.ipv4.tcp_mem
       - net.ipv4.tcp_rmem
       - net.ipv4.tcp_wmem
       - net.ipv4.tcp_max_orphans
@@ -93,17 +94,18 @@ spec:
     content: |
       #!/bin/sh
       cat <<EOT >> /etc/sysctl.d/999-testground.conf
-      net.core.somaxconn = 100000
-      net.netfilter.nf_conntrack_max = 1000000
-      net.ipv4.tcp_max_syn_backlog = 100000
-      net.core.netdev_max_backlog = 100000
+      net.core.somaxconn = 131072
+      net.netfilter.nf_conntrack_max = 1048576
+      net.ipv4.tcp_max_syn_backlog = 65536
+      net.core.netdev_max_backlog = 524288
       net.ipv4.ip_local_port_range = 1024 65535
       net.ipv4.tcp_tw_recycle = 1
       net.ipv4.tcp_tw_reuse = 1
-      net.core.rmem_max = 134217728
-      net.core.wmem_max = 134217728
-      net.ipv4.tcp_rmem = 4096 12582912 67108864
-      net.ipv4.tcp_wmem = 4096 12582912 67108864
+      net.core.rmem_max = 131072
+      net.core.wmem_max = 131072
+      net.ipv4.tcp_mem = 262144 524288 1048576
+      net.ipv4.tcp_rmem = 4096 65536 131072
+      net.ipv4.tcp_wmem = 4096 65536 131072
       EOT
   image: kope.io/k8s-1.17-debian-stretch-amd64-hvm-ebs-2020-01-17
   machineType: c5.2xlarge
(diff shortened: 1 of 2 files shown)


coryschwartz:
The way these are written to sysctl.conf means they won't take effect until reboot, right?

infra/k8s/cluster.yaml
```diff
@@ -93,17 +94,18 @@ spec:
...
       net.ipv4.tcp_tw_reuse = 1
-      net.core.rmem_max = 134217728
-      net.core.wmem_max = 134217728
-      net.ipv4.tcp_rmem = 4096 12582912 67108864
-      net.ipv4.tcp_wmem = 4096 12582912 67108864
+      net.core.rmem_max = 131072
+      net.core.wmem_max = 131072
+      net.ipv4.tcp_mem = 262144 524288 1048576
+      net.ipv4.tcp_rmem = 4096 65536 131072
+      net.ipv4.tcp_wmem = 4096 65536 131072
```
coryschwartz, line 108: Should we end these with `sysctl -p` to apply this file?

nonsense:
> The way these are written to sysctl.conf means they won't take effect until reboot, right?

Yes, I think so, we re-create a cluster after changing them.

infra/k8s/cluster.yaml
nonsense: I don't think so, why? They are in the `sysctl.d` directory and they get picked up on reboot.

nonsense:
FYI `net.netfilter.nf_conntrack_max = 1048576` is already set on our hosts, and changing it in our `999-testground.conf` doesn't have any effect. Something runs after that and sets it.

nonsense:
@raulk please change `nf_conntrack_max` at `infra/k8s/testground-infra/values.yaml`  as well

infra/k8s/cluster.yaml
coryschwartz: Right, but you wouldn't reboot after this usually.

nonsense: These values are applied to the host, you can check them when you ssh, and then `sysctl -a`.

I don't think we need a reboot, after they've been set in the `userdata`.

nonsense:
Having spoke to @coryschwartz I changed my mind, and now I also think we can just set these on the `redis` container, no need to set them host-wide.

raulk:
@nonsense I cooked up these values to be applied cluster-wide. They are important too on the client-side and on the VMs they will be running test plans.

nonsense:
@raulk client-side is basically the `testplan` pods (`cluster_k8s.go`), so @coryschwartz is right we can just apply them per `pod`.

Kubuxu:
Increase `net.core.netdev_budget` too.
Probably also install netdata on it and check for warnings there.

nonsense:
@coryschwartz let's agree that this PR is about the values (as we are already updating the sysctl on the host on `master` in this way), and address the way we apply sysctls in a follow up PR.

nonsense:
@Kubuxu thanks for the suggestion. @raulk already told me about netdata and it was very useful, we are using it!

Will review what this `netdev_budget` is set to on our hosts right now and see about updating it in a follow up PR.

nonsense:
@coryschwartz 

> https://github.com/kubernetes/kops/blob/master/docs/cluster_spec.md#docker

I don't see any `sysctl` in the `docker` section of `spec`.

> https://pkg.go.dev/k8s.io/kops/pkg/apis/kops?tab=doc#DockerConfig

I also don't see any sysctl settings here in the `DockerConfig`

---

So I am not sure how exactly you suggest we apply the sysctls.

---

There is a `sysctlParameters` in `spec`, but these don't seem to be propagated to containers started in Kubernetes either (for example `net.core.somaxconn` stays at 128, when starting a new container).

PR #799, opened 2020-04-03
fix: close response and read full

Stebalien:
1. Read may just read a single byte.
2. Make sure to actually close the body to avoid leaking.

### DIFF
--- a/sdk/runtime/runner.go
+++ b/sdk/runtime/runner.go
@@ -156,7 +156,11 @@ func setupMetrics(ctx context.Context, runenv *RunEnv) (doneCh chan error) {
 				if err != nil {
 					continue
 				}
-				_, _ = resp.Body.Read(b)
+				_, err = io.ReadFull(resp.Body, b)
+				resp.Body.Close()
+				if err != nil {
+					continue
+				}
 				if string(b) == "OK" {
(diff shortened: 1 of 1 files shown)


sdk/runtime/runner.go
```diff
@@ -156,7 +156,11 @@ func setupMetrics(ctx context.Context, runenv *RunEnv) (doneCh chan error) {
 				if err != nil {
 					continue
 				}
-				_, _ = resp.Body.Read(b)
+				_, err = io.ReadFull(resp.Body, b)
```
nonsense, line 159: Good catch.

sdk/runtime/runner.go
```diff
@@ -156,7 +156,11 @@ func setupMetrics(ctx context.Context, runenv *RunEnv) (doneCh chan error) {
...
+				resp.Body.Close()
+				if err != nil {
```
nonsense, line 161: I think this was breaking the functionality for some weird reason. I remember applying it, and then we had to revert.

raulk: It was breaking because we were not properly managing `io.EOF`. `io.ReadFull` does that for us, so this patch shouldn't cause a regression.

Issue #779, opened 2020-04-01
k8s node gets into `NotReady` state possibly due to OOM


nonsense:
When running:

```
testground run single benchmarks/subtree \
    --builder=docker:go \
    --runner=cluster:k8s \
    --build-cfg bypass_cache=true \
    --build-cfg push_registry=true \
    --build-cfg registry_type=aws \
    --run-cfg keep_service=false \
    --instances=5000 --collect
```

I started getting the following warnings in the `testground daemon`:

```
Apr  1 15:33:51.302130  WARN    got error when trying to fetch pod logs {"req_id": "7c479e0e", "runner": "cluster:k8s", "run_id": "b7a100d42ada", "err": "Get https://172.20.55.188:10250/containerLogs/default/tg-benchmarks-b7a100d42ada-sin
gle-282/tg-benchmarks-b7a100d42ada-single-282?follow=true&sinceSeconds=1000: net/http: TLS handshake timeout"}
```

The node 172.20.55.188 was in `NotReady` state:

```
ip-172-20-55-188.eu-central-1.compute.internal   NotReady   node     74m   v1.17.3   172.20.55.188   3.127.220.195    Debian GNU/Linux 9 (stretch)   4.9.0-11-amd64   docker://19.3.4
```

Apart from `benchmark` testplans, we also have the following on that node:

```
dummy-m8zkl                                            1/1     Running            0          72m     100.96.7.2      ip-172-20-55-188.eu-central-1.compute.internal   <none>           <none>
prometheus-testground-infra-prometheu-prometheus-0     3/3     Running            2          72m     100.96.7.3      ip-172-20-55-188.eu-central-1.compute.internal   <none>           <none>
testground-infra-prometheus-node-exporter-n64vt        1/1     Running            0          72m     172.20.55.188   ip-172-20-55-188.eu-central-1.compute.internal   <none>           <none>
testground-sidecar-9vbsl                               1/1     Running            0          72m     172.20.55.188   ip-172-20-55-188.eu-central-1.compute.internal   <none>           <none>
```

Running a `kubectl describe` on the node, results in:

```
  (Total limits may be over 100 percent, i.e., overcommitted.)
  Resource                    Requests      Limits
  --------                    --------      ------
  cpu                         6570m (82%)   3440m (43%)
  memory                      6684Mi (43%)  6744Mi (43%)
  ephemeral-storage           0 (0%)        0 (0%)
  attachable-volumes-aws-ebs  0             0
Events:
[...]

PR #801, opened 2020-04-06
Fix/separate k8s infrastructure from plans

coryschwartz:
fix: https://github.com/ipfs/testground/issues/780

This creates a separate autoscaling group (tginfra) on which redis, prometheus, prometheus-pushgateway, and grafana are isolated away from other nodes. I created a label `testground.nodetype` 

In this diff, the WORKER_NODES environment variable corresponds only for the autoscaling group with the label `testground.nodetype=plan` 

Sidecar and plan pods are restricted to `testground.nodetype=plan`

testground infrastructure is restricted to `testground.nodetype=infra`

### DIFF
--- a/infra/k8s/cluster.yaml
+++ b/infra/k8s/cluster.yaml
@@ -144,12 +144,55 @@ spec:
       net.ipv4.tcp_rmem = 4096 12582912 67108864
       net.ipv4.tcp_wmem = 4096 12582912 67108864
       EOT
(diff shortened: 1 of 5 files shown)


coryschwartz:
@nonsense 

I added just two nodes in tginfra, but I haven't tested what the difference in performance actually is when separated this way.

Robmat05:
fixes pods issue called out in #779

PR #816, opened 2020-04-07
k8s healthcheck checker api

coryschwartz:
Implement part one of https://github.com/ipfs/testground/issues/805

Remove the old healthcheck methods, and the broken "fakefixes" logic, and in its place use a more generic kubernetes pod checker to verify whether the number of pods is correct.


pkg/healthcheck/checkers.go
```diff
@@ -76,6 +79,45 @@ func CheckDirectoryExists(path string) Checker {
...
+			return false, fmt.Sprintf("command failed `%s` (%v)", cmd, err), nil
+		}
+		// command returns with a zero exit code.
+		return true, fmt.Sprintf("command completed successfully `%s`", cmd), nil
+	}
+}
+
+// CheckK8sPods returns a checker which verifies the number of pods found matches the number
+// expected. If Listing the pods returns an error, the error is returned. The boolean value returned
+// by the check follows whether the number of pods observed in the list matches the expected count.
```
coryschwartz, line 104: This method intentionally doesn't check the Status of the pods to verify they are all running, only that they exist in the correct number.

My rational for this is that kubernetes is generally self-correcting, so the fix action for a pod not being started is to do nothing and the scheduler will eventually succeed. Unlike local runners, testground is not the scheduler/supervisor for pods on k8s, and can be of no assistance when it comes to restarting pods once they are in the control of the scheduler.

As it is, the Fixer which would fix the problem detected by CheckK8sPods  (not yet written) would apply a new pod configuration through the API, or using kubectl apply, or helm install, etc.

On the other hand, perhaps it would be worthwhile to know before starting that the right number of pods exist. I could imagine a Fixer which watched/waited for pods to become ready. Waiting on pods to become ready is a different category of problem, and I want to distinguish between them.

pkg/runner/cluster_k8s.go
```diff
@@ -293,138 +293,60 @@ func (*ClusterK8sRunner) ID() string {
...
-	}
-
-	redisCheck = api.HealthcheckItem{Name: "redis", Status: api.HealthcheckStatusOK, Message: "redis service is running"}
-	return
-}
-
-func (c *ClusterK8sRunner) healthcheckSidecar() (sidecarCheck api.HealthcheckItem) {
-	sidecarCheck = api.HealthcheckItem{Name: "sidecar", Status: api.HealthcheckStatusFailed, Message: "sidecar service is not reachable"}
+func (c *ClusterK8sRunner) Healthcheck(ctx context.Context, engine api.Engine, ow *rpc.OutputWriter, fix bool) (*api.HealthcheckReport, error) {
+	// TODO how does one pass the context to k8s API calls?
```
nonsense, line 297: I don't think you can send `ctx` to k8s API calls, I think you can only set a `timeout` param. Would be interested to hear if you find anything.

pkg/runner/cluster_k8s.go
```diff
@@ -293,138 +293,60 @@ func (*ClusterK8sRunner) ID() string {
 	return "cluster:k8s"
 }
 
-func (c *ClusterK8sRunner) healthcheckK8s() (k8sCheck api.HealthcheckItem) {
-	k8sCheck = api.HealthcheckItem{Name: "k8s", Status: api.HealthcheckStatusOK, Message: "k8s cluster is running"}
-	cmd := exec.Command("kops", "validate", "cluster")
```
nonsense, line 298: I like that we are removing this - it will be a pain (and most probably a mistake) to move `kops validate cluster` to k8s, once the daemon is on the remote cluster.

pkg/runner/cluster_k8s.go
```diff
@@ -293,138 +293,60 @@ func (*ClusterK8sRunner) ID() string {
...
-	})
-	if err != nil {
-		sidecarCheck.Message = err.Error()
-		return
-	}
-	if len(pods.Items) != nodes {
-		sidecarCheck.Message = fmt.Sprintf("expected %d sidecar pods. found %d.", nodes, len(pods.Items))
-		return
-	}
+	hh.Enlist("kops validate",
```
nonsense, line 313: We should do this healthcheck in a different way - just check that we have a number of healthy nodes maybe?

It'd be nice if we don't make assumptions on the Kubernetes cluster in Testground - users should be able to spin it up as they see fit - for example if there is a way to run the networking part in EKS or GKE, it should be possible to run Testground there as well.

pkg/runner/cluster_k8s.go
```diff
@@ -293,138 +293,60 @@ func (*ClusterK8sRunner) ID() string {
...
+		hc.CheckCommandStatus(ctx, "kops", "validate", "cluster"),
+		hc.NotImplemented(),
+	)
 
-	for _, pod := range pods.Items {
-		if pod.Status.Phase != "Running" {
-			return
-		}
-	}
+	hh.Enlist("efs pod",
```
nonsense, line 318: With this we are locked in AWS as well, and we might want to revise in the future, how to make it configurable.

However with the `kops` validation above we will most definitely have issues when we try to move the daemon on the cluster.

pkg/runner/cluster_k8s.go
nonsense: That could very well be off scope for this PR, so just ignore if you don't want to fix now!

coryschwartz:
Per our conversation, implements the same checkers which already exist, and there will be changes needed in the future to support running the testground daemon on k8s or in environments other than AWS, but these can be tackled separately.

PR #825, opened 2020-04-08
Influx metrics -- spike pushing metrics immediately

coryschwartz:
Alternative to #826.

---

One possible strategy for https://github.com/ipfs/testground/issues/755 

maybe push into influx db directly from plans.
But this would work better if influxdb was accessible in the cluster, much as we do with prometheus/pushgateway now. Runs like this:

```
./testground run single benchmarks/all \
--builder docker:go \
--runner cluster:k8s \
--run-cfg influx_url=https://us-west-2-1.aws.cloud2.influxdata.com \
--run-cfg influx_token=XXXXXXXXXX  \
--run-cfg influx_org=YYYYYYYYYYYY \
--run-cfg influx_bucket=ZZZZZZZZZZZ \
--instances 100 \
--collect
```

As you might expect this fails completely when the sidecar is active, although this would work alright if configured to use influxdb in the cluster or if we configured the sidecar to permit this traffic. However, if the sidecar is disabled, this does push metrics into an influxdb cloud.

### DIFF
--- a/plans/benchmarks/benchmarks.go
+++ b/plans/benchmarks/benchmarks.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
 	"math"
 	"math/rand"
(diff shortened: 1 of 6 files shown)


plans/benchmarks/benchmarks.go
```diff
@@ -31,8 +55,10 @@ func StartTimeBench(runenv *runtime.RunEnv) error {
 	elapsed := time.Since(runenv.TestStartTime)
 	emitTime(runenv, "Time to start", elapsed)
 
-	gauge := runenv.M().NewGauge(runtime.GaugeOpts{Name: "start_time", Help: "time from plan scheduled to plan booted"})
-	gauge.Set(float64(elapsed))
+	w := metricsWriter(runenv)
+	w.WritePoint(makePoint(runenv, "time to start", map[string]interface{}{"start_time": elapsed}))
```
nonsense, line 59: I think we should do a few back-of-the-envelope type calculations here, so that we have an idea of how much data we are talking about and whether this scales with 1 node of InfluxDB.

What is the payload size of one point in this format? How many of these can we collect in InfluxDB if we want a 7-day retention period (anything less than that is mostly useless I think, we should probably even strive for 30-60 days).

For operational metrics, I think it would be valuable to also have aggregates on the client side. Just yesterday I was running tests with 10k nodes issuing a single measurement on every one of the 5000 writes and reads and we are talking about a single measurement.

In contrast the library mentioned at https://github.com/ipfs/testground/pull/605 (go-metrics) is doing a client side aggregation, and this works just fine for operational metrics, and scales with the number of measurements and flush interval, and NOT the number of time we call `emitTime`.

nonsense: Related to the data / sizing estimations:

1. https://docs.influxdata.com/influxdb/v1.7/guides/hardware_sizing/#single-node-or-cluster

2. https://docs.influxdata.com/influxdb/v1.7/guides/hardware_sizing/#influxdb-oss-guidelines

---

If we assume that we are going to have 10k clients on a given run, then we are left with 25 writes per second per testplan instance, which is a bit useless if we are going to be writing out individual data points I think. Even thinking about how that could work at the end of the testplan, just processing all the data points collected during the run, and I am still not convinced it'd work.

---

If we however aggregate every `flush interval`, this reduces the load orders of magnitude.

plans/benchmarks/benchmarks.go
```diff
@@ -31,8 +55,10 @@ func StartTimeBench(runenv *runtime.RunEnv) error {
...
+	w.Flush()
```
raulk, line 60: My concern with the "immediate send" approach is that:

a) we don't _really_ need the data to be available _immediately_.
b) Influxdb will have to cope with bursty writes, coming from as many as 10k clients.
c) sending data continuously can affect the performance of the test case itself, as well as the network.

Related: I don't think we should be flushing the batch on every single write.

raulk:
Thanks for doing this spike -- it help us set course for a definitive solution.

We ended up implementing a different design that differentiates between various types of observations:

- lifecycle events.
- diagnostics.
- results.

See https://github.com/testground/sdk-go/pull/3/ for a more detailed description.

PR #826, opened 2020-04-08
influx collect -- spike deferring metric push until after test completes

coryschwartz:
I am not confident that this works flawlessly, 

One possible strategy for https://github.com/ipfs/testground/issues/755

Compared to https://github.com/ipfs/testground/pull/825 this more closely resembles the strategy laid out in https://github.com/ipfs/testground/issues/755

This pushes metrics from the collections. It does this by unmarshaling the log files into their original objects and marshaling them again into influxdb points, then pushing them into influxdb.

Just as before, the collections tarball is downloaded for the user to look at offline, but the metrics are extracted and also pushed into influx. 

```
./testground run single benchmarks/all \
--builder docker:go \
--runner cluster:k8s \
--run-cfg influx_url=https://us-west-2-1.aws.cloud2.influxdata.com \
--run-cfg influx_token=XXXXXXXXXX  \
--run-cfg influx_org=YYYYYYYYYYYY \
--run-cfg influx_bucket=ZZZZZZZZZZZ \
--instances 100 \
--collect
```

This code could be moved pretty easily to another place, but as a proof of concept, this is functional

### DIFF
--- a/cmd/collect.go
+++ b/cmd/collect.go
@@ -55,13 +55,18 @@ func collectCommand(c *cli.Context) error {
 		return err
 	}
 
-	return collect(ctx, api, runner, id, output)
+	return collect(ctx, c, api, runner, id, output)
 }
 
-func collect(ctx context.Context, cl *client.Client, runner string, runid string, outputFile string) error {
+func collect(ctx context.Context, c *cli.Context, cl *client.Client, runner string, runid string, outputFile string) error {
+	comp, err := createSingletonComposition(c)
+	if err != nil {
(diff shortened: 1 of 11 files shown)


raulk:
Thanks for doing this spike -- it help us set course for a definitive solution.

We ended up implementing a different design that differentiates between various types of observations, each of which gets collected differently:

- lifecycle events.
- diagnostics.
- results.

See https://github.com/testground/sdk-go/pull/3/ for a more detailed description.

PR #827, opened 2020-04-08
apply resource limits to sidecar

nonsense:


nonsense:
This is necessary as a result of https://github.com/ipfs/testground/pull/820/files - `requests` cannot be lower than `limits`.

PR #829, opened 2020-04-08
pod resources per composition group

nonsense:
Addresses https://github.com/ipfs/testground/issues/789

TODO:
- [ ] ~check memory as well, not just CPU~

### DIFF
--- a/pkg/api/composition.go
+++ b/pkg/api/composition.go
@@ -60,8 +60,8 @@ type Metadata struct {
 }
 
 type Resources struct {
-	Memory string `toml:"memory"`
-	CPU    string `toml:"cpu"`
+	Memory string `toml:"memory" json:"memory"`
+	CPU    string `toml:"cpu" json:"cpu"`
 }
 
 type Group struct {
@@ -69,7 +69,7 @@ type Group struct {
 	ID string `toml:"id" json:"id"`
 
 	// Resources requested for each pod from the Kubernetes cluster
-	Resources Resources `toml:"resources"`
+	Resources Resources `toml:"resources" json:"resources"`
 
 	// Instances defines the number of instances that belong to this group.
 	Instances Instances `toml:"instances" json:"instances"`
--- a/pkg/api/composition.go
+++ b/pkg/api/composition.go
@@ -59,9 +59,17 @@ type Metadata struct {
 	Author string `toml:"author" json:"author"`
 }
 
+type Resources struct {
+	Memory string `toml:"memory"`
+	CPU    string `toml:"cpu"`
+}
+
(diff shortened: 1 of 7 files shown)


nonsense:
I consider implementing `check memory as well, not just CPU`, but we really don't need that now, and this whole functionality will be redundant once we add an autoscaler and/or try to introduce spot instances.

The issue right now is that, if we don't add resource check (which is not race-free, and in general not 100% accurate), developers can try to schedule big workloads, and they would be blocked in `Pending` state on Kubernetes, without much feedback to Testground.

---

This is strictly better than `master`, and I think is good enough for now.

pkg/api/composition.go
```diff
@@ -59,10 +59,18 @@ type Metadata struct {
 	Author string `toml:"author" json:"author"`
 }
 
+type Resources struct {
+	Memory string `toml:"memory"`
+	CPU    string `toml:"cpu"`
```
raulk, line 64: Missing json tag.

pkg/api/composition.go
```diff
@@ -59,10 +59,18 @@ type Metadata struct {
...
+}
+
 type Group struct {
 	// ID is the unique ID of this group.
 	ID string `toml:"id" json:"id"`
 
+	// Resources requested for each pod from the Kubernetes cluster
+	Resources Resources `toml:"resources"`
```
raulk, line 72: Missing json tag.

nonsense:
> If we add this in the composition schema, we should apply in local:docker also, and emit a warning when set on local:exec. Otherwise, the feature is uneven and it's not clear which runners support it.

By `emit a warning`, you mean emit a warning that these values are ignored, right?

For `local:exec` they would totally be ignored as we have no isolation.

For `local:docker` do we want to apply resource restrictions at all?

nonsense:
@raulk I added a warning for `local:exec` and `local:docker` in case resources are set in the composition file for a given group.

pkg/runner/cluster_k8s.go
```diff
@@ -886,11 +881,28 @@ func (c *ClusterK8sRunner) maxPods(podResourceCPU resource.Quantity) (int, error
...
+
+		neededCPUs += podCPU * float64(g.Instances)
+	}
+
+	if (availableCPUs * utilisation) > neededCPUs {
+		return true, nil
+	}
 
-	return pods, nil
+	ow.Warnw("not enough resources on cluster", "availableCPUs", availableCPUs, "neededCPUs", neededCPUs, "utilisation", utilisation)
```
raulk, line 904: nit: lower snake case for logging fields.

PR #833, opened 2020-04-08
reproduce redis readiness probe failure ; with aggregate op metrics to influxdb

nonsense:
This is my WIP `benchmarks` testplan, which is simulating 10k `sync service` subscribers, all of which read 5000 entries from a given subtree.

Here are screenshots from the operational metrics and the performance of the XREAD (as measured from the client - the `subscribers`)

Screenshot 2020-04-08 at 00 25 49
Screenshot 2020-04-07 at 23 11 06

You can clearly see the spike in XREAD calls if we send a swarm of tcp connections to Redis without adding a random delay after the barrier is released.

cc @raulk @coryschwartz


plans/benchmarks/benchmarks.go
```diff
@@ -314,27 +338,74 @@ func SubtreeBench(runenv *runtime.RunEnv) error {
...
+
+					if strings.Compare(*b, *tst.Data) != 0 {
+						metrics.GetOrRegisterResettingTimer("subscriber.test.iteration.fail", nil).UpdateSince(now)
+						return fmt.Errorf("received unexpected value")
+					}
+
+					if i%500 == 0 {
+						runenv.RecordMessage("received %d items (series: %s)", i, tst.Name)
+					}
+					metrics.GetOrRegisterResettingTimer("subscriber.test.iteration.ok", nil).UpdateSince(now)
```
nonsense, line 395: @coryschwartz following my comment on your `influxdb metrics` PR - we have to think if we want to support that kind of measurements in the SDK. I personally find them useful, but I am fine if we don't.

However I am not sure it will be feasible for 1000 testplan instances, each running 10 tests (concurrently), going over 5000 iterations, to send individual measurements to InfluxDB without aggregating. This is just a single measurement, I imagine we would want to have 10-100 measurements per testplan.

In any case it is both experimenting with both approaches.

Robmat05:
This will stay in prgress till sync services refactor is complete

PR #843, opened 2020-04-13
wait longer for cluster validate

nonsense:
Sometimes download `kops` dependencies on the `master` takes a lot of time...

### DIFF
--- a/infra/k8s/install.sh
+++ b/infra/k8s/install.sh
@@ -51,13 +51,8 @@ kops create -f $CLUSTER_SPEC
 kops create secret --name $NAME sshpublickey admin -i $PUBKEY
 kops update cluster $NAME --yes
 
-# wait for worker nodes and master to be ready
-kops validate cluster --wait 10m
-if [ $? -ne 0 ]
-then
-	echo "cluster was not ready after 10 minutes."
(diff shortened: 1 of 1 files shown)


nonsense:
In this particular failure, we ended up waiting for minutes to download protokube:

https://github.com/kubernetes/kops/blob/master/upup/pkg/fi/cloudup/urls.go#L54

which results in:

```
protokubeImage:
  hash: d507311f110c69caa58ea887edbb34efcdd2dd56864a666d4cb82bc3a69c99a6
  name: protokube:1.17.0-beta.1
  sources:
  - https://artifacts.k8s.io/binaries/kops/1.17.0-beta.1/images/protokube.tar.gz
  - https://github.com/kubernetes/kops/releases/download/v1.17.0-beta.1/images-protokube.tar.gz
  - https://kubeupv2.s3.amazonaws.com/kops/1.17.0-beta.1/images/protokube.tar.gz
```

Note that the first URL is dead.

nonsense:
I wouldn't call this `P0` - we rely on a lot of software and services (CDNs, github, AWS, etc.) that other companies provide, and sometimes this bites us - no way around it, we just live with it.

PR #844, opened 2020-04-13
rm TEST_CASE_SEQ as a selector; replace with test case name.

raulk:
Fixes https://github.com/ipfs/testground/issues/466.

* Removes `TEST_CASE_SEQ` as an env variable, including its bindings in the `RunEnv`.
* Test plans now use `TEST_CASE` as a string selector for the test case to run.
* New `runtime.InvokeMap(map[string]runtime.TestCaseFn)` function for supplying a `string => func` map to select the test case from.

### DIFF
--- a/sdk/runtime/runner.go
+++ b/sdk/runtime/runner.go
@@ -36,10 +36,24 @@ var PushgatewayEndpoints = []string{"prometheus-pushgateway:9091", "localhost:90
 // invoked. If we were unable to start the listener, this value will be "".
 var HTTPListenAddr string
 
-// Invoke runs the passed test-case and reports the result.
-func Invoke(tc func(*RunEnv) error) {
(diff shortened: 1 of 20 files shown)


sdk/runtime/runner.go
```diff
@@ -36,10 +36,24 @@ var PushgatewayEndpoints = []string{"prometheus-pushgateway:9091", "localhost:90
 // invoked. If we were unable to start the listener, this value will be "".
 var HTTPListenAddr string
 
+type TestCaseFn func(env *RunEnv) error
+
+// InvokeMap takes a map of test case names and their functions, and calls the
+// matched test case, or panics if the name is unrecognised.
+func InvokeMap(cases map[string]TestCaseFn) {
```
coryschwartz, line 43: This is quite different; we'll have to update the gitbook with this as well!

PR #845, opened 2020-04-14
make: clean docker env, build image, install testground

nonsense:
An improved Makefile so that we have a simpler way to re-compile and restart Testground locally, and avoid issues like: https://github.com/ipfs/testground/pull/841#issuecomment-613337422

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,16 @@ define eachmod
 	@find . -type f -name go.mod -print0 | xargs -I '{}' -n1 -0 bash -c 'dir="$$(dirname {})" && echo "$${dir}" && cd "$${dir}" && $(1)'
 endef
 
+all: clean docker-ipfs-testground install-testground
+
+clean:
+	docker rm -f testground-sidecar || true
+	docker rm -f testground-redis || true
+	docker rm -f testground-goproxy || true
+
+install-testground:
+	go install ./...
+
 pre-commit:
 	python -m pip install pre-commit --upgrade --user
 	pre-commit install --install-hooks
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -6,6 +6,16 @@ define eachmod
...
 endef
 
+all: clean docker-ipfs-testground install-testground
+
+clean:
+	docker rm -f testground-sidecar || true
+	docker rm -f testground-redis || true
+	docker rm -f testground-goproxy || true
+
+install-testground:
```
raulk, line 16: A few comments:

* We can safely drop `-testground` suffix; it's redundant.

* Also, let's rename the `docker-ipfs-testground` target to `docker`.

* `clean` is a misnomer here; `make clean` is commonly used to clean build targets, not to delete docker containers. Let's look for another name.

nonsense: I was hesitant to change `docker-ipfs-testground` as it has been used for ages, but we are in v0.4 so probably best to introduce better naming now, than later.

Updated these, let me know if you think they are better.

Makefile
```diff
@@ -6,11 +6,21 @@ define eachmod
...
+
+install:
+	go install ./...
+
 pre-commit:
 	python -m pip install pre-commit --upgrade --user
 	pre-commit install --install-hooks
 
-docker-ipfs-testground:
+build-sidecar-image:
```
raulk, line 23: This doesn't just build the sidecar image, it builds testground entirely. Let's call this `make docker`.

We also need to update `.travis.yml`

PR #846, opened 2020-04-14
cleanup: remove docker-swarm

nonsense:
This PR is:
1. Removing `infra/docker-swarm` as we no longer use it.
2. Renaming `infra/docker` to `infra/local-docker`, so that it is a bit more explicit what these Dockerfiles are used for.
3. Removing `infra/docker/testground-redis-exporter` as I see that we are using directly `bitnami/redis-exporter` in our code. @coryschwartz I think this is not used.
4. Removing `s3fs`.


infra/docker/testground-redis-exporter/Dockerfile
```diff
@@ -1,5 +0,0 @@
-FROM bitnami/redis-exporter
```
nonsense, line 1: @coryschwartz I think this file is not used, so I am removing it.

coryschwartz: Yes, I think this was added unintentionally.

PR #847, opened 2020-04-14
add pod ip to testplan env vars

nonsense:
This will be helpful when debugging issues with networking on Kubernetes.

cc @vyzo

### DIFF
--- a/pkg/runner/cluster_k8s.go
+++ b/pkg/runner/cluster_k8s.go
@@ -213,6 +213,15 @@ func (c *ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow *rpc
 				Value: cfg.LogLevel,
 			})
 		}
+
+		env = append(env, v1.EnvVar{
+			Name: "POD_IP",
+			ValueFrom: &v1.EnvVarSource{
+				FieldRef: &v1.ObjectFieldSelector{
+					FieldPath: "status.podIP",
+				}},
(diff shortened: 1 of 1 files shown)


coryschwartz:
I was not aware you could create environment variables for information post creation, but this does work.

Just for my own benefit, looking at a pod with `kubectl describe pod` displays the follwing in the environment

```
POD_IP:                      (v1:status.podIP)
```

and POD_IP contains the ip address of the eth0 interface.


However this environment variable would not be available on other runners, so any plan written to use this would only be useful for debugging kubernetes. Why not just get the IP address in code so it can be used for more runners?

nonsense:
> I was not aware you could create environment variables for information that is only available post creation, but this does work.

This is called `the Downward API` - https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#the-downward-api

---

> However this environment variable would not be available on other runners, so any plan written to use this would only be useful for debugging kubernetes. Why not just get the IP address in code so it can be used for more runners?

Yes, the idea is to use it for debugging purposes. I think there was a bug with parsing of IPv4 addresses, so libp2p was `detecting IPs` that didn't really match the IP of the interface.

---

You could get this by Go code as well (specially if you have multiple interfaces on the container, as we do), but it is handy to know the POD_IP as well if necessary. This is strictly for debugging purposes.

Issue #750, opened 2020-03-27
client-side test plan management


raulk:
Once we implement #747, in order to achieve #749, we'll need to provide some CLI goodness to manage test plans at the client side.

Supported sources: Git repos and filesystem.

## Potential CLI

  - `testground plan add --name foo --from git://...`
  - `testground plan add --name foo --from file://...`
  - `testground plan switch --name foo --to branch_a`
  - `testground plan switch --name foo --to master`
  - `testground plan remove --name foo`

## Notes

- `switch` only works on git origins.
- `add` for git origins would clone the upstream into managed source directory under `$TESTGROUND_HOME/plans_src`.
- `add` for local filesystem origins would symlink into `$TESTGROUND_HOME/plans_src`.

raulk:
@Robmat05 why did you close this as discarded?

Robmat05:
I meant to close #745, but these tickets were right next to each other and this was a keystroke error. ugh. Nice catch, thanks

PR #849, opened 2020-04-14
the big refactor: shipping source, manifests, TESTGROUND_HOME, SDK linkage

raulk:
This PR introduces a major refactor in the testground architecture.

## Shipping the test plan source to the daemon

The daemon no longer knows about specific test plans or SDKs, and no longer discovers manifests. The client now submits the source of the test plan to the daemon. We do this through a `multipart/related` request, with three parts:

 * Part 1 (Content-Type: application/json): composition json.
 * Part 2 (Content-Type: application/zip): test plan source.
 * Part 3 (optional, Content-Type: application/zip): linked sdk, see below.

This allows the daemon to be completely agnostic of test plans, and for test plans to be hosted anywhere, not compulsorily on this repo.

## TESTGROUND_HOME

All activity of testground, both client and daemon, revolves around the `TESTGROUND_HOME` directory. It can be specified through the corresponding env variable (`$TESTGROUND_HOME`), or we automatically fall back to `$HOME/testground`. We create the directory and its inner layout automatically, if it doesn't exist.

The inner layout of this directory is as follows:

* `$TESTGROUND_HOME/plans` => where plans are to be hosted. One directory per plan. They can be git working copies, symlinks to local directories, etc. In commands, we reference the plan by its directory name.
* `$TESTGROUND_HOME/sdks` => where SDKs are hosted. This is mostly used by testground developers, as described below, to test changes to an SDK.
* `$TESTGROUND_HOME/data/outputs` => runners deposit test plan outputs here. Of internal use.
* `$TESTGROUND_HOME/data/work` => work directory, for unpacking files, staging source, etc. Of internal use.

The `EnvConfig` now has a `Dirs()` method that returns a `Directories` struct from which testground components can obtain the full path of the above directories.

## Manifests

Manifests are now only a client-side artefact.

Manifests are purely human/machine readable descriptions so that users know what test cases reside in test plan, and which parameters each accepts. They are test plan metadata, useful for building UIs and powering porcelain.

Manifests must now sit on the root of a test plan. As explained above, a test plan can now live anywhere. The only requirement is that it is accompanied by a `manifest.toml` file.

## Describe and list commands

Since the daemon no longer has knowledge of test plans, these commands are now client-side only. They operate on `$TESTGROUND_HOME/plans`.

The `describe` command now only accepts test plans, and enumerates all test cases. The logic to accept test cases has been dropped, as it was poorly implemented.

## SDK linkage

We no longer do any magic `replace`s. Test plans are compiled against a concrete version of the SDK, as determined in their `go.mod`.

However, testground developers tend to modify the SDK itself. And it's extremely handy to be able to modify the SDK locally, and ship that SDK for linkage/replacement/override into the test plan when the build is performed.

This is achieved via a new option: `--link-sdk=<value>`, where `<value>` can be:

1. an absolute path.
2. a path relative to `$PWD`.
3. a path relative to `$TESTGROUND_HOME/sdks`.

(in that order of precedence).

An example of (3) in action would be:

* Inside `$TESTGROUND_HOME/sdks`, I have a symbolic link `sdk-go` that points to my local working copy of the Go SDK.
* When I build/run a test plan, I can supply `--link-sdk=sdk-go`, and the client will pack the SDK source and ship it to the daemon, in part 3 of the multipart request, as explained above.
* The daemon will unpack that SDK, and apply the relevant replace directives (in Go builders) to override the plan's dependency with the shipped SDK.

## Nomenclature of test plans and test cases

Previously, the format to refer to a test case was: `testplan/testcase`. This PR changes the nomenclature to `path/to/test/plan:testcase`.

Everything before the colon is the plan’s location under `$TESTGROUND_HOME/plans`. In the aforementioned case, we expect to find a plan with a `manifest.toml` exactly at: `$TESTGROUND_HOME/path/to/test/plan`. And we expect the `manifest.toml` to contain a test case named `testcase1`.

This new logic makes it possible to unambiguously refer to test plans from within multi-test-plan repos, without confounding slashes for path segments and as a separator between testplan and testcase.

Rationale: I predict folks will host multiple plans in a single repo, e.g. `libp2p/test-plans` may contain `dht`, `pubsub`, etc. test plans.

With #750, we'll add that repo with `testground plan add libp2p git:github.com/libp2p/test-plans.git`, and we'll import multiple test plans at once.

## TODO

- [X] README.
- [X] ~tests.~ (we'll do a test-oriented sprint)

---

Fixes https://github.com/ipfs/testground/issues/747
Fixes https://github.com/ipfs/testground/issues/629


pkg/config/loader.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/ipfs/testground/pkg/logging"
+)
+
+const (
+	EnvTestgroundHomeDir = "TESTGROUND_HOME"
+
+	DefaultListenAddr = "localhost:8042"
+)
+
+func (e *EnvConfig) Load() error {
```
raulk, line 19: The `*EnvConfig` now has a `Load()` method that loads the configuration, applies any env variables, and fallbacks.

pkg/daemon/build.go
```diff
@@ -36,3 +50,94 @@ func (srv *Daemon) buildHandler(engine api.Engine) func(w http.ResponseWriter, r
 		tgw.WriteResult(out)
 	}
 }
+
+func consumeBuildRequest(r *http.Request, dir string) (*api.BuildRequest, string, string, error) {
```
raulk, line 54: This is where the server consumes the multipart build request.

pkg/runner/local_common.go
```diff
@@ -30,28 +29,33 @@ func localCommonHealthcheck(ctx context.Context, hh *healthcheck.Helper, cli *cl
...
+	// 			PortBindings: exposed,
+	// 			NetworkMode:  container.NetworkMode(controlNetworkID),
+	// 		},
+	// 		ImageStrategy: docker.ImageStrategyBuild,
+	// 		BuildImageOpts: &docker.BuildImageOpts{
+	// 			Name:     "testground-prometheus:latest",
+	// 			BuildCtx: filepath.Join(srcdir, "infra/local-docker/testground-prometheus"),
+	// 		},
+	// 	}),
+	// )
```
raulk, line 55: Since we're dropping prometheus for test plan monitoring, I'm inclined to drop it for local executors, as we'd probably use other tools to monitor them.

pkg/runner/local_docker.go
```diff
@@ -149,11 +149,16 @@ func (r *LocalDockerRunner) Healthcheck(ctx context.Context, engine api.Engine,
...
+		// TODO(raulk): not sure what to do with this; it's not really useful
+		//  beyond the first run anyway. Since we are not distributing binaries
+		//  we can safely assume that the user will anyway have to build testground.
+		//  If the build instructions consist in running a make target that also
+		//  builds the sidecar, we would've omitted the need for this entirely.
+		// ImageStrategy: docker.ImageStrategyBuild,
+		// BuildImageOpts: &docker.BuildImageOpts{
+		// 	Name:     "ipfs/testground:latest",
+		// 	BuildCtx: engine.EnvConfig().Home,
+		// },
```
raulk, line 161: I'm planning to remove this @coryschwartz.

pkg/config/dirs.go
```diff
@@ -0,0 +1,27 @@
...
+
+type Directories struct {
+	home string
+}
+
+func (d Directories) Home() string {
+	return d.home
+}
+
+func (d Directories) Plans() string {
```
raulk, line 13: @coryschwartz for #750, you'd need to operate in this directory.

raulk:
@coryschwartz thanks for the thorough review! I'll address those comments tomorrow.

* The change of behaviour in `list` is definitely a regression; I'll fix that. 
* Re: no longer building the sidecar, I think it's reasonable to ask the user to run `make docker` or `make install` when they clone the repo, instead of `go install`. We'd take care of building the image in those make targets instead. WDYT?

pkg/daemon/build.go
```diff
@@ -36,3 +50,94 @@ func (srv *Daemon) buildHandler(engine api.Engine) func(w http.ResponseWriter, r
...
+	if err = json.NewDecoder(p).Decode(&req); err != nil {
+		return nil, "", "", fmt.Errorf("failed to json decode request body: %w", err)
+	}
+
+	var (
+		planzip *os.File
+		sdkzip  *os.File
+	)
+
+	// 2a. Read test plan source archive.
```
coryschwartz, line 96: There is substantial storage redundancy here. For each run, a directory is created which contains the compressed and uncompressed copies of the plan being executed.

plans/benchmarks/manifest.toml
```diff
@@ -7,27 +7,27 @@ source_path = "file://${TESTGROUND_SRCDIR}/plans/benchmarks"
 builder = "exec:go"
```
coryschwartz, line 7: line 4: Remove source_path right?

pkg/build/golang/docker.go
```diff
@@ -174,30 +135,36 @@ func (b *DockerGoBuilder) Build(ctx context.Context, in *api.BuildInput, ow *rpc
...
-	cmd := exec.CommandContext(ctx, "go", append([]string{"mod", "edit"}, replaces...)...)
-	cmd.Dir = plandst
-	cmd.Stdout = os.Stdout
-	cmd.Stderr = os.Stderr
-	err = cmd.Run()
-	if err != nil {
-		return nil, fmt.Errorf("unable to add replace directives to go.mod; %w", err)
+	if sdksrc != "" {
+		replaces = append(replaces,
+			"-replace=github.com/ipfs/testground/sdk/sync=../sdk/sync",
```
coryschwartz, line 144: I think we should remove these replace lines.

When the sdk is provided by the client, I can't say for certain that the SDK is `github.com/ipfs/testground/...` Even if it was, it is possible that whatever local edits the user has made might not be satisfied with just these replace lines.

What I am thinking is that the SDK should be renamed to `replaces` or some such. This way, the user can add replacements in their own go mod to replace the SDK or whatever other code they want to add replacements. The contract should be that this directory will sit next to the plans directory and your replace lines will be used as you have written them. What do you think of this?

coryschwartz:
>     * Re: no longer building the sidecar, I think it's reasonable to ask the user to run `make docker` or `make install` when they clone the repo, instead of `go install`. We'd take care of building the image in those make targets instead. WDYT?

I'm just looking to reduce the number of gotchas that might prevent this from working right off the bat. Someone might hear about testground, run `go get` and then think it's broken and give up without any further research. -- we've talked before about error messages pointing the user to what steps they need to take for other things, this could be one of those. "container not found. build it with make docker and try again"

pkg/config/loader.go
```diff
@@ -0,0 +1,93 @@
...
+		e.dirs.SDKs(),
+		e.dirs.Work(),
+	} {
+		if err := ensureDir(d); err != nil {
+			return fmt.Errorf("failed to check/create directory %s: %w", d, err)
+		}
+	}
+
+	// parse the .env.toml file, if it exists.
+	f := filepath.Join(e.dirs.Home(), ".env.toml")
```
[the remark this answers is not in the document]
raulk: SGTM. It does contain credentials, though. But now that this is moved outside the source repo and into $TESTGROUND_HOME, it's less dangerous. Previously it was easy to accidentally commit and push it.

pkg/build/golang/docker.go
raulk: > When the sdk is provided by the client, I can't say for certain that the SDK is github.com/ipfs/testground/...

Not sure I follow. The SDK always has a fixed import path. That's how test plans import it (`github.com/ipfs/testground/sdk/{runtime,sync}` pre-refactor, `github.com/ipfs/testground/sdk` after #858).

And this is a Go builder, so it knows how to apply that override in Go. A JS builder will know the import path of its SDK and will use npm/package.json fiddling appropriately.

raulk: > What I am thinking is that the SDK should be renamed to replaces or some such. This way, the user can add replacements in their own go mod to replace the SDK or whatever other code they want to add replacements. The contract should be that this directory will sit next to the plans directory and in the same way it is on the client's local TESTGROUND_HOME. This way, the user can write their own replace entries if they want to provide customizations to the SDK or any other component, and their go.mod will make sense on the client side and when the daemon does the builds.

The SDK really has a special standing here, because you code a test plan against an SDK.

Overriding arbitrary dependencies is useful for when you run a test plan against version X or Y of an upstream dependency, e.g. go-libp2p v0.2.2 vs v0.3.0.

To make that functionality really useful, we need to allow for overriding dependencies **at the group level**, so you can compose hybrid test runs with groups A and B, where the former hits one version, and the latter hits another.

We already have that functionality: users can specify upstream dependency overrides via the `groups.build.dependencies` composition field. See it in action here with the DHT test plan: https://github.com/ipfs/testground/blob/master/compositions/all-cypress-k8s.toml#L126-L132

The only shortcoming is that the user cannot ship arbitrary local code for that replace, it needs to live under a commit in the origin repo of that dependency.

That _might_ be useful; I'm just now sure how much. Let's stash the idea for now, and not intertwine it with SDK replaces, which are plan-wide, not group-scoped.

raulk:
> I'm just looking to reduce the number of gotchas that might prevent this from working right off the bat.

So instead of instructing the user to run `go install .`, we just instruct them to run `make install`, which runs the docker build as well as the go install.

pkg/config/loader.go
```diff
@@ -0,0 +1,93 @@
...
+	if _, err := os.Stat(f); err == nil {
+		// try to load the optional .env.toml file
+		_, err = toml.DecodeFile(f, e)
+		if err != nil {
+			return fmt.Errorf("found .env.toml at %s, but failed to parse: %w", f, err)
+		}
+		logging.S().Infow(".env.toml loaded from %s", f)
```
nonsense, line 73: `Infof`

Also, `testground daemon` now complains with:

```
Apr 16 12:14:45.923955  DPANIC  Ignored key without a value.    {"ignored": "/Users/nonsense/testground/.env.toml"}
Apr 16 12:14:45.923946  INFO    .env.toml loaded from %s
```

I am not sure what this DPANIC is about, because I don't have keys without values in my `.env.toml`.

raulk: It's coming from zap, due to the wrong log format.

raulk: Fixed in https://github.com/ipfs/testground/pull/849/commits/6d6296cd161a91c89b1f7f624cc519abbca814de

raulk:
> > I'm just looking to reduce the number of gotchas that might prevent this from working right off the bat.
> 
> So instead of instructing the user to run `go install .`, we just instruct them to run `make install`, which runs the docker build as well as the go install.

@coryschwartz #802 introduces an improved Makefile, such that we now go install testground AND build the docker container with a simple `make` command.

pkg/runner/local_docker.go
```diff
@@ -231,23 +233,23 @@ func (r *LocalDockerRunner) Run(ctx context.Context, input *api.RunInput, ow *rp
...
-			odir := filepath.Join(r.outputsDir, input.TestPlan.Name, input.RunID, g.ID, strconv.Itoa(i))
+			odir := filepath.Join(r.outputsDir, input.TestPlan, input.RunID, g.ID, strconv.Itoa(i))
 			err = os.MkdirAll(odir, 0777)
 			if err != nil {
 				err = fmt.Errorf("failed to create outputs dir %s: %w", odir, err)
 				break
 			}
 
-			name := fmt.Sprintf("tg-%s-%s-%s-%s-%d", input.TestPlan.Name, testcase.Name, input.RunID, g.ID, i)
+			name := fmt.Sprintf("tg-%s-%s-%s-%s-%d", input.TestPlan, input.TestCase, input.RunID, g.ID, i)
```
coryschwartz, line 243: We need to filter characters docker doesn't allow. 

Right now if I start testground with this incantation
```
testground run single --builder docker:go --runner local:docker plans/placebo:ok -i 1

```

I get an error 
```
Apr 16 18:53:21.228605	WARN	engine run error: Error response from daemon: Invalid container name (tg-plans/placebo-ok-fa6c8935f543-single-0), only [a-zA-Z0-9][a-zA-Z0-9_.-] are allowed	{"req_id": "2a8d699c"}
```

Which is caused by the `/` character

raulk: Thanks for reporting. The problem is that we were sending the client-side path of the plan to the server.

As of https://github.com/ipfs/testground/pull/849/commits/9c6aaffa9c88a3e4b62c3a29de9cbb86d7539164, we now replace it with the name of the plan as per the manifest.

We should still replace illegal characters for Docker/k8s, but that bug was already present before this PR. And it's not too bad to fail this way.

PR #859, opened 2020-04-15
remove k8s scripts

coryschwartz:
Step 3 of https://github.com/ipfs/testground/issues/855

Deleting k8s, so it can be added in the new repo


coryschwartz:
The local_docker can be removed as well if we plan on depreciating custom docker builds

nonsense:
@coryschwartz as soon as you are done moving to the new repo, just merge this PR, so that we don't have discrepancies in versioning ;)

coryschwartz:
They are up, in https://github.com/testground/infra

PR #860, opened 2020-04-16
add pinger

coryschwartz:
Tester for https://github.com/ipfs/testground/issues/348

Simply test whether instances are reachable over the control and data networks and report an error when something doesn't work right.


plans/verify/main.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	runtime.InvokeMap(testcases)
+}
+
+func setupNetwork(ctx context.Context, runenv *runtime.RunEnv) (*sync.Client, error) {
+	client := sync.MustBoundClient(ctx, runenv)
+	return client, client.WaitNetworkInitialized(ctx, runenv)
+}
+
+func isControlNet(nw string) bool {
```
coryschwartz, line 27: I know I can parse and check if the address is in range rather than using HasPrefix. 

Is there a better way for determining the range for k8s control networks? It might not always be this /24,

nonsense: @coryschwartz please extract those as `const`s at the top of the file.

plans/verify/main.go
```diff
@@ -0,0 +1,109 @@
...
+				runenv.RecordFailure(err)
+				return err
+			}
+			for _, addr := range addrs {
+				runenv.RecordMessage("publishing %s", addr)
+				client.Publish(ctx, netTopic, addr.String())
+			}
+		}
+		client.Publish(ctx, netTopic, ".")
+		runenv.RecordMessage("networks published. ready to be tested.")
```
nonsense, line 76:
```suggestion
		runenv.RecordMessage("published my addresses from all networks to sync service. ready to be tested.")
```

plans/verify/main.go
```diff
@@ -0,0 +1,109 @@
...
+			if err != nil {
+				runenv.RecordFailure(err)
+				return err
+			}
+			for _, addr := range addrs {
+				runenv.RecordMessage("publishing %s", addr)
+				client.Publish(ctx, netTopic, addr.String())
+			}
+		}
+		client.Publish(ctx, netTopic, ".")
```
nonsense, line 75: Please extract `"."` into a var and name it somehow - `endOfStream` or `endOfNetworks` or something like that.

nonsense:
@coryschwartz LGTM. I added a few comments on stuff I think you could improve, and the tests/linter/go mod tidy needs to pass.

plans/verify/main.go
```diff
@@ -0,0 +1,109 @@
...
+		runenv.RecordMessage("networks published. ready to be tested.")
+		client.SignalEntry(ctx, "target-ready")
+
+	case pingmode:
+		runenv.RecordMessage("ping mode. waiting for target networks.")
+		<-client.MustBarrier(ctx, "target-ready", 1).C
+		runenv.RecordMessage("starting ping")
+		nwCh := make(chan string)
+		client.Subscribe(ctx, netTopic, nwCh)
+		for network := <-nwCh; network != "."; network = <-nwCh {
```
nonsense, line 85: `"."` here as well.

PR #861, opened 2020-04-16
add limit range to default namespace

nonsense:
@coryschwartz I wonder if the way we are adding the LimitRange now, doesn't apply to every namespace in k8s, or just to `default`. I guess we don't want to limit containers in the `kube-system` namespace, but only in the `default` namespace.

Related: https://github.com/ipfs/testground/issues/856

### DIFF
--- a/infra/k8s/install.sh
+++ b/infra/k8s/install.sh
@@ -60,7 +60,7 @@ echo
 echo "Install default container limits"
 echo
 
-kubectl apply -f ./limit-range/limit-range.yaml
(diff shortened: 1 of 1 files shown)


nonsense:
The limits we add don't seem to be in effect for `kube-system`, so I guess this PR is not needed.

For example, `weave` reports:

```
    Restart Count:  0
    Requests:
      cpu:      10m
    Readiness:  http-get http://127.0.0.1:6784/status delay=0s timeout=1s period=10s #success=1 #failure=3
```

PR #862, opened 2020-04-16
increase weave cpu requests

nonsense:


nonsense:
I think 1 vCPU is a bit too much, but let's be on the safe side, and scale down if that's the case.

PR #867, opened 2020-04-17
Blackhole the control network

coryschwartz:
With this change, we have the following routing table

```
 ip route

default via 24.0.0.0 dev eth1
24.0.0.0 dev eth1 scope link
30.180.0.0/16 dev eth1 scope link  src 30.180.8.7
blackhole 100.64.0.0/10
100.64.0.10 via 100.96.11.1 dev eth0
100.96.7.5 via 100.96.11.1 dev eth0
blackhole 100.96.11.0/24
```

### DIFF
--- a/pkg/sidecar/k8s_reactor.go
+++ b/pkg/sidecar/k8s_reactor.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/containernetworking/cni/libcni"
 	"github.com/vishvananda/netlink"
+	"github.com/vishvananda/netlink/nl"
 	"github.com/vishvananda/netns"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


coryschwartz:
I'm torn about whether to blackhole routes on the docker reactor as well. Already on the docker route configuration doesn't allow connectivity, but they aren't explicitly null routed. 

@nonsense @raulk what do you think?

nonsense:
@raulk this is what the `pinger` (https://github.com/ipfs/testground/pull/860 PR) is supposed to be. I will review and test on Monday as well.

nonsense:
> Mind creating a test plan that verifies the new behaviour? Ideally a test plan that fails with the previous behaviour and passes with this PR.

This has been address in the `pinger` test, so merging this.

PR #869, opened 2020-04-18
client plan mgmt

coryschwartz:
fix: https://github.com/ipfs/testground/issues/750


coryschwartz:
This is branched from feat/the-big-refactor which is not yet merged, It makes sense to merge this after that one.

go.mod
```diff
@@ -39,14 +41,16 @@ require (
 	github.com/otiai10/copy v1.0.2
 	github.com/pborman/uuid v1.2.0
 	github.com/pierrec/lz4 v2.5.1+incompatible // indirect
+	github.com/sergi/go-diff v1.1.0 // indirect
 	github.com/stretchr/testify v1.4.0
-	github.com/urfave/cli v1.22.1
+	github.com/urfave/cli/v2 v2.2.0
```
raulk, line 46: Awesome, thanks for doing the upgrade!

pkg/cmd/common.go
```diff
@@ -73,9 +72,11 @@ func createSingletonComposition(c *cli.Context) (*api.Composition, error) {
 	case 0:
 		return nil, errors.New("wrong format for test case name, should be: `<path to testplan>:testcase`, where `<path to testplan> is relative to $TESTGROUND_HOME")
 	case 2:
+		fmt.Println("it was 2", ss)
```
raulk, line 75: leftovers

pkg/cmd/common.go
```diff
@@ -21,8 +21,7 @@ func setupClient(c *cli.Context) (*client.Client, *config.EnvConfig, error) {
 	if err := cfg.Load(); err != nil {
 		return nil, nil, err
 	}
-
-	endpoint := c.GlobalString("endpoint")
+	endpoint := c.String("endpoint")
```
raulk, line 24: In v1 we had to access the global flag namespace because the `endpoint` flag was defined in the root command -- is that no longer the case? We should test this change.

pkg/cmd/plan.go
```diff
@@ -0,0 +1,212 @@
...
+			Usage: "create plan `PLAN_NAME`",
+			Flags: []cli.Flag{
+				&cli.StringFlag{
+					Name:     "module",
+					Usage:    "Create module named `MODULE_NAME`",
+					Required: false,
+					Value:    "github.com/your/module/name",
+				},
+			},
+			Action: createCommand,
```
raulk, line 37: This is cool. We'll have to refine the godocs a little to reflect how this works, but a command like this looks, on the surface, useful for the "quick start" user flow, as long as it doesn't introduce a lot of complexity.

pkg/cmd/plan.go
```diff
@@ -0,0 +1,212 @@
...
+		&cli.Command{
+			Name:  "create",
+			Usage: "create plan `PLAN_NAME`",
+			Flags: []cli.Flag{
+				&cli.StringFlag{
+					Name:     "module",
+					Usage:    "Create module named `MODULE_NAME`",
+					Required: false,
+					Value:    "github.com/your/module/name",
+				},
```
raulk, line 35: This command actually creates a **Golang** plan. It will fall short once we have SDKs for multiple languages (soon, with TypeScript next). Plans written in other languages will have to produce entirely different artifacts.

If we want to build this sustainably, we'll want to:

- introduce a flag: `--target=go`, `--target=ts-node`, `--target=ts-browser`.
- store all templates for a language under a directory with the target name.
- package those templates into the compiled `testground` binary via a tool like https://github.com/markbates/pkger.
- when this command is called, load the templates, invoke them, and produce the result.

Alternatively, how about storing these templates in a repo? `github.com/testground/templates`, with a directory per target env.

On creating a test plan, we could fetch the latest commit, pick the right subdirectory, and emit the templated artifacts?

pkg/cmd/plan.go
```diff
@@ -0,0 +1,212 @@
...
+			},
+			Action: createCommand,
+		},
+		&cli.Command{
+			Name:   "import",
+			Usage:  "`GIT_REPO` [local_repo]",
```
raulk, line 41:
```suggestion
			Usage:  "`GIT_REPO` [location]",
```

pkg/cmd/plan.go
```diff
@@ -0,0 +1,212 @@
...
+		}
+		tmpl.Execute(f, c.String("module"))
+		f.Close()
+	}
+	return nil
+}
+
+func importCommand(c *cli.Context) error {
+	if c.Args().Len() < 1 {
+		return errors.New("missing reuired argument GIT_REPO")
```
raulk, line 101:
```suggestion
		return errors.New("missing required argument GIT_REPO")
```

pkg/cmd/plan.go
raulk, line 41:
```suggestion
			Usage:  "`GIT_URL` [local_repo]",
```

pkg/cmd/plan.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	gitURL := c.Args().First()
+
+	var gitDir string
+	if c.Args().Len() > 1 {
+		gitDir = c.Args().Get(1)
+	} else {
+		sl := strings.Split(gitURL, "/")
+		gitDir = sl[len(sl)-1]
```
raulk, line 110: Why not let the user specify a location relative to `$TESTGROUND_HOME/plans`, and take care of creating any parents if they don't exist?

raulk: If we were to keep this logic, we should be using `filepath` for path manipulations, as it uses the OS separator.

pkg/cmd/plan.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	cloneOpts := git.CloneOptions{
+		URL: c.Args().First(),
+	}
+
+	_, err := git.PlainClone(filepath.Join(cfg.Dirs().Plans(), gitDir), false, &cloneOpts)
+	if err != nil {
+		return err
+	}
```
raulk, line 125: We probably want to check if the target already exists, and fail if so.

raulk: Unless the git clone already does that (which now that I think of it, it probably does), but we may not want to leave it up to Git. I'd rather make that check explicit.

pkg/cmd/plan.go
```diff
@@ -0,0 +1,212 @@
...
+	return nil
+}
+
+func rmCommand(c *cli.Context) error {
+	if c.Args().Len() != 1 {
+		return errors.New("Missing required argument PLAN_DIR")
```
raulk, line 131: By convention, error strings start with a lower-case.

raulk: Using the term `PLAN_DIR` requires the user to know what that variable is. Can we be more descriptive?

pkg/cmd/plan.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	cfg := &config.EnvConfig{}
+	if err := cfg.Load(); err != nil {
+		return err
+	}
+
+	return os.RemoveAll(filepath.Join(cfg.Dirs().Plans(), c.Args().First()))
```
raulk, line 139: Since this is a destructive operation, we should ask for confirmation first.

pkg/cmd/plan.go
```diff
@@ -0,0 +1,212 @@
...
+		&cli.Command{
+			Name:   "import",
+			Usage:  "`GIT_REPO` [local_repo]",
+			Action: importCommand,
+		},
+		&cli.Command{
+			Name:   "branch",
+			Usage:  "`LOCAL_REPO` `BRANCH_NAME`",
+			Action: importCommand,
+		},
```
raulk, line 48: We discussed omitting the branch command. There are too many edge cases, and it's likely that the user will want to manage branches directly. Also, it only applies to Git origins only (we want the `testground plan` command to be able to manage local symlinks).

pkg/cmd/common.go
coryschwartz: Thanks. Removed.

pkg/cmd/plan.go
coryschwartz: Git already does this, and it reports the message `repository already exists`

pkg/cmd/common.go
raulk: @coryschwartz ?

pkg/cmd/plan.go
raulk: @coryschwartz ?

pkg/cmd/plan.go
raulk: @coryschwartz ?

pkg/cmd/plan.go
raulk: @coryschwartz ?

pkg/cmd/plan.go
raulk: @coryschwartz ?

pkg/cmd/plan.go
raulk: 👍

raulk: I'm marking this as resolved, then.

pkg/cmd/plan.go
raulk: @coryschwartz ?

pkg/cmd/common.go
coryschwartz: Yeah this still works, i.e.
```
testground --endpoint 192.168.1.50:8081 healthcheck --runner local:docker
Apr 28 16:03:13.267328	INFO	using home directory: /home/cory/testground
Apr 28 16:03:13.267750	INFO	testground client initialized	{"addr": "192.168.1.50:8081"}

pkg/cmd/plan.go
coryschwartz: I created a --target flag which currently just a simple map. I somehow missed the idea of using pkger. We could store these on another repo or download them on the fly as well.

pkg/cmd/plan.go
coryschwartz: This was originally picking off the last section of the URL for the name of the git directory.

pkg/cmd/plan.go
coryschwartz: I've made changes to the error messages.

pkg/cmd/plan.go
coryschwartz: I removed the branch command -- yeah, it didn't work with local imports. Additionally, I enabled git with a flag `--git` so that it would work even with local bare repos!

coryschwartz:
Closing, merging https://github.com/testground/testground/pull/928 instead

Robmat05:
This ticket is combined with #928 to accomplish the original goal of #750. Most of the feature sets have been accomplished. with the exception of what is called out in #930 which will be fixed shortly

PR #872, opened 2020-04-20
change ipfs/testground to iptestground/sidecar

nonsense:
TODO:
- [x] wait for the initial `iptestground/sidecar:edge` build to happen on Docker Hub.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,9 +6,9 @@ define eachmod
 	@find . -type f -name go.mod -print0 | xargs -I '{}' -n1 -0 bash -c 'dir="$$(dirname {})" && echo "$${dir}" && cd "$${dir}" && $(1)'
 endef
 
-.PHONY: install tidy mod-download lint build-all docker install test
+.PHONY: install tidy mod-download lint build-all docker-sidecar install test
 
-install: goinstall docker
+install: goinstall docker-sidecar
 
 goinstall:
 	go install .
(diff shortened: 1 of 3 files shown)


pkg/runner/local_docker.go
```diff
@@ -118,7 +118,7 @@ func (r *LocalDockerRunner) Healthcheck(ctx context.Context, engine api.Engine,
 	sidecarContainerOpts := docker.EnsureContainerOpts{
 		ContainerName: "testground-sidecar",
 		ContainerConfig: &container.Config{
-			Image:      "ipfs/testground:latest",
```
nonsense, line 121: We should probably fix this. If someone builds a release, we want to be picking up the given release docker image (for example `iptestground/sidecar:v0.4.0`), rather than `edge` or `latest`.

So this should be a var that we overwrite during release.

Makefile
```diff
@@ -29,8 +29,8 @@ lint:
 build-all:
 	$(call eachmod,go build -tags balsam -o /dev/null ./...)
 
-docker:
-	docker build -t ipfs/testground .
+docker-sidecar:
+	docker build -t iptestground/sidecar:edge .
```
nonsense, line 33: The semantics I use are:
`edge` - latest master build
`latest` - latest stable release (currently v0.4.0)
`v0.4.0`, `v0.3.0`, etc.

nonsense: Also if someone picks the source for a given release (v0.4.0 or v0.3.0, etc.), this will always be building latest master. I know we are not pulling images from testground, but maybe we should re-consider that... other how am I going to run release v0.4.0 after we release v0.5.0 - what/who is going to pull the sidecar image with the correct version for me?

PR #877, opened 2020-04-21
storm testplan

nonsense:
Trying to establish a baseline on the number of TCP connections we can establish between containers, considering that we are running Weave on the data plane.

### DIFF
--- a/plans/benchmarks/main.go
+++ b/plans/benchmarks/main.go
@@ -25,13 +25,8 @@ func run(runenv *runtime.RunEnv) error {
 		return BarrierBench(runenv)
 	case "subtree":
 		return SubtreeBench(runenv)
-	case "all":
-		_ = StartTimeBench(runenv)
-		_ = NetworkInitBench(runenv)
-		_ = NetworkLinkShapeBench(runenv)
-		_ = BarrierBench(runenv)
-		_ = SubtreeBench(runenv)
-		return nil
+	case "storm":
+		return Storm(runenv)
 	default:
 		msg := fmt.Sprintf("Unknown Testcase %s", c)
 		return errors.New(msg)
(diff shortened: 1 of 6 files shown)


plans/storm/main.go
```diff
@@ -0,0 +1,251 @@
...
+	defer cancel()
+
+	ch := make(chan *ListenAddrs)
+	if _, _, err := client.PublishSubscribe(subCtx, PeerTopic, mynodeInfo, ch); err != nil {
+		return nil, errors.Wrap(err, "publish/subscribe failure")
+	}
+
+	res := []string{}
+
+	for i := 0; i < runenv.TestInstanceCount; i++ {
```
nonsense, line 239: @raulk we have to think if we can speed this up. When we have 10k instances, each of them has to read 10k records == 10^10 calls.

Ultimately this operation takes ~5-6min. and I wonder if we can make it faster.

Robmat05:
Per @nonsense, w need to iterate further on this. Specifically, create dashboards for it, and lock the run params and everything else in code. Moving to v0.6 as P1

raulk:
Next steps to get this merged:

- @nonsense will work on https://github.com/testground/testground/issues/994.
- @nonsense will migrate the storm test plan to use the new metrics API in the runenv.
- @nonsense will create dashboards and enhance the cluster:k8s SLO document on the docs site to explain graphically.

plans/benchmarks/main.go
```diff
@@ -25,13 +25,8 @@ func run(runenv *runtime.RunEnv) error {
 		return BarrierBench(runenv)
 	case "subtree":
 		return SubtreeBench(runenv)
-	case "all":
```
nonsense, line 28: I am removing `all`, because `WaitNetworksInitialised` cannot be called twice right now, making this test to fail. Once we fix the sidecar, we might want to re-enable it again, but for now, we have to individually call different cases.

PR #894, opened 2020-04-22
remove pkg package

nonsense:
We spoke about this a few days ago, I think it is a good idea.

No point having a `pkg` package, when everything lives in it.


nonsense:
> Actually, if we intend to keep some test plans in this repo, I think it's not superfluous to have that top-level separation.

Even with `plans` directory staying with a few tests (`network`, `placebo`, etc.) I think we are better off with `pkg`. Then again I have no strong feelings about it, so I am good either way - we can either merge or close.

coryschwartz:
This really makes little difference to me. I see the point about the plans being separate, which seems cleaner, but I have no opinion really one way or the other.

PR #901, opened 2020-04-23
save circleci go mod and remove travis yaml

coryschwartz:


coryschwartz:
I removed .travis.yml in https://github.com/testground/testground/pull/901.

The circleci migration is complete, at least as a replica of the travis configuration. Someone with the right github superpowers will need to upgrade the billing status, however,  so that docker image layer caching can be enabled.

Issue #572, opened 2020-02-18
move build/image push to the k8s cluster


nonsense:
When trying to use the cluster:k8s runner, the build/image push process is super slow if your upload speed is low.

We should come up with a way of building and pushing from a remote builder, rather than from the local developer machine, which has low upload bandwidth, because this increases the feedback loop with minutes.

Robmat05:
Will be fixed in #849

nonsense:
This is not done yet, there are at least 2 open PRs related to it.

PR #902, opened 2020-04-23
set build container network to `host` (needed for cluster:k8s runner)

nonsense:
Related with: https://github.com/testground/infra/pull/18

Addresses: https://github.com/testground/testground/issues/572

---

We need the `build container` to be on the `host` network, so that it can access a goproxy, which is also attached to the `host` network. Otherwise we have no way of discovering the goproxy service from our cluster (at least I couldn't figure out one).

---

TODO:
- [ ] decide on `default` setup for goproxy.

### DIFF
--- a/pkg/build/golang/docker.go
+++ b/pkg/build/golang/docker.go
@@ -55,7 +55,7 @@ type DockerGoBuilderConfig struct {
 	// Docker image to, if PushRegistry is true.
 	RegistryType string `toml:"registry_type" overridable:"yes"`
 
-	// GoProxyMode specifies one of "on", "off", "custom".
+	// GoProxyMode specifies one of "local", "direct", "remote".
 	//
 	//   * The "local" mode (default) will start a proxy container (if one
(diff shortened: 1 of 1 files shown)


nonsense:
@raulk currently `docker:go` default option for goproxy is to start a container (`go_proxy_mode=local`).

This is quite bad for the k8s setup - we really don't want testground to be starting containers by default alongside the daemon.

---

How shall we address this?

Options:
1. We could introduce another builder `k8s:go`, similar to `docker:go`, and not have the `go_proxy_mode=local` functionality at all.
2. We could change the default behaviour for now (maybe use `direct` by default), and have users use `local` when running testground locally, or using `localhost:8081` with `remote`, when using `k8s`.

nonsense:
> I am curious why the communication with the goproxy is not done through a service on k8s

Because the `build container` does not have access to the k8s services. The `build container` is not started by k8s, so DNS is not set up.

PR #903, opened 2020-04-23
medium sized windows

coryschwartz:
Set the size parameter on the windows orb

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,8 +93,8 @@ jobs:
 
   # https://circleci.com/docs/2.0/executor-intro/#windows
   test-windows:
-    executor: win/default
-    parameters:
+    executor:
+      name: win/default
       size: medium
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


coryschwartz:
I'm doing something wrong. It looks like the executor takes a "size" parameter, but I still have 16G even if it is set to "medium", (the smallest size)

Logged in over SSH to verify, sure enough 16G:
```
PS C:\Users\circleci> wmic MemoryChip get Capacity
wmic MemoryChip get Capacity
Capacity     
16106127360  
```

coryschwartz:
I understand now. "medium" means something different depending on the OS selection.

https://circleci.com/product/features/resource-classes/

So "medium", the smallest class we can select, is a 7G linux VM and a 15G Windows VM.

Closing without merging. 

By the way, I can't specify "small", because the schema is an `enum ("medium" "large" "xlarge" "2xlarge")`

Issue #907, opened 2020-04-23
Epic: Observations + aggregate metrics directions


raulk:
After some discussion within the testground Core team, we came to these conclusions. They haven't been reviewed by all team members in this form yet, so YMMV and it's subject to minor changes.

## A. Lifecycle events to be dispatched immediately

We will want to ingest/ship some vital signs immediately, so that the user can view progress of the test plan in real-time by watching Grafana.

These are a discrete set of meta-events, what we're calling "lifecycle events", and the volume is pretty low, so should not produce an observer effect:

- instance start, network initialized, instance failed, instance succeeded, instance crashed events.
- stage entry/exit events (when we implement the stages API).
by default, key sync service events (barriers we’re waiting on, time waiting, etc.)
- if a test plan is testing the sync service, it can disable those via `client.DisableMetrics()`.

## B. Future: Lifecycle events feed the supervisor/watchtower service

Eventually, we will want the watchtower/supervisor service to monitor the lifecycle events on InfluxDB, in order to:

- Tell the daemon (runner) what the progress / status is, so it can be communicated to the client.
- Declare global success/failure.
- Maybe purge the sync service for the run’s keys.

## C. [go-metrics](https://github.com/rcrowley/go-metrics) to be namespaced under `runenv.M()` for aggregate metrics

- Replace the Prometheus delegates with go-metrics.
- Registered metrics should be output every 1 second (or user configurable) as JSON events on metrics.out via zap.
- We will need to unify the schema of these events with the "metric" runenv event.

## D. User-recorded observations and metrics will be shipped on completion

- We will redirect raw observations (`runenv.RecordMetric`) and periodic aggregate metrics dump to a `metrics.out` output asset.
- When the test plan finishes, we traverse `metrics.out` and batch-insert all metrics, linked with their original timestamp, into InfluxDB.

## E. Infrastructure metrics

- Infrastructure health should be monitored via Prometheus IMO.
- No pushgateway involved, just standard Prometheus + node_exporter + redis exporter + kubernetes, etc.
- Combined with infrastructure monitoring via Prometheus (visualised on the same Grafana), to correlate any failures with infrastructure incidents.

## F. Future: efficient binary format for large volumes of observations/metrics

- JSON is a rather inefficient format for large volumes of data points.
- Implement a binary mode for storing metrics in an efficient binary format like Apache Parquet or Apache Avro.
- User would enable it via `runenv.EnableBinaryOutput()`.
[...]

PR #913, opened 2020-04-24
generalise redis route to services routes; prepare for InfluxDB addition

nonsense:
This will be needed for https://github.com/testground/testground/issues/907 and I can see this generalisation has been done for Docker already.

One change that I am introducing is that the sidecar doesn't panic if it can't resolve the IP of a service - allowing for both required and optional services.

---

This is useful for me, because I also run InfluxDB with almost every cluster, and want to have the route allowed, without having to recompile the sidecar.

### DIFF
--- a/pkg/sidecar/docker_reactor.go
+++ b/pkg/sidecar/docker_reactor.go
@@ -14,10 +14,10 @@ import (
 	"github.com/vishvananda/netlink"
 	"github.com/vishvananda/netns"
 
-	"github.com/testground/testground/pkg/docker"
-	"github.com/testground/testground/pkg/logging"
 	"github.com/testground/sdk-go/runtime"
 	"github.com/testground/sdk-go/sync"
+	"github.com/testground/testground/pkg/docker"
+	"github.com/testground/testground/pkg/logging"
 )
 
 // PublicAddr points to an IP address in the public range. It helps us discover
@@ -31,22 +31,24 @@ import (
 var PublicAddr = net.ParseIP("1.1.1.1")
 
 type DockerReactor struct {
-	client  *sync.Client
-	routes  []net.IP
-	manager *docker.Manager
+	client         *sync.Client
+	servicesRoutes []net.IP
+	manager        *docker.Manager
 }
 
 func NewDockerReactor() (Reactor, error) {
 	// TODO: Generalize this to a list of services.
 	wantedRoutes := []string{
 		os.Getenv(EnvRedisHost),
+		os.Getenv(EnvInfluxdbHost),
 	}
 
 	var resolvedRoutes []net.IP
 	for _, route := range wantedRoutes {
 		ip, err := net.ResolveIPAddr("ip4", route)
 		if err != nil {
-			return nil, fmt.Errorf("failed to resolve host %s: %w", route, err)
+			logging.S().Warnw("failed to resolve host", "host", route, "err", err.Error())
+			continue
 		}
 		resolvedRoutes = append(resolvedRoutes, ip.IP)
 	}
@@ -65,9 +67,9 @@ func NewDockerReactor() (Reactor, error) {
 	client.EnableBackgroundGC(nil)
 
 	return &DockerReactor{
-		client:  client,
-		routes:  resolvedRoutes,
-		manager: docker,
+		client:         client,
+		servicesRoutes: resolvedRoutes,
+		manager:        docker,
 	}, nil
 }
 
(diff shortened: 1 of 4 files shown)


pkg/sidecar/k8s_reactor.go
```diff
@@ -37,17 +37,26 @@ var (
...
+		os.Getenv(EnvInfluxdbHost),
+	}
 
-	redisIp, err := net.ResolveIPAddr("ip4", redisHost)
-	if err != nil {
-		return nil, fmt.Errorf("failed to resolve redis: %w", err)
+	var resolvedRoutes []net.IP
+	for _, route := range wantedRoutes {
+		ip, err := net.ResolveIPAddr("ip4", route)
+		if err != nil {
```
nonsense, line 55: Change in behaviour here - we no longer return an error if we can't resolve the service, allowing for `optional` services.

pkg/sidecar/docker_reactor.go
```diff
@@ -31,22 +31,24 @@ import (
...
 		os.Getenv(EnvRedisHost),
+		os.Getenv(EnvInfluxdbHost),
 	}
 
 	var resolvedRoutes []net.IP
 	for _, route := range wantedRoutes {
 		ip, err := net.ResolveIPAddr("ip4", route)
 		if err != nil {
-			return nil, fmt.Errorf("failed to resolve host %s: %w", route, err)
+			logging.S().Warnw("failed to resolve host", "host", route, "err", err.Error())
```
nonsense, line 50: Change in behaviour here - we no longer return an error if we can't resolve the service, allowing for `optional` services.

pkg/sidecar/k8s_reactor.go
```diff
@@ -175,20 +184,24 @@ func (d *K8sReactor) manageContainer(ctx context.Context, container *docker.Cont
 		return nil, fmt.Errorf("failed to get link by name %s: %w", controlNetworkIfname, err)
 	}
 
-	// Get the routes to redis. We need to keep these.
-	redisRoute, err := getRedisRoute(netlinkHandle, d.redis)
-	if err != nil {
-		return nil, fmt.Errorf("cant get route to redis: %s", err)
+	servicesIPs := []net.IP{}
```
raulk, line 187: nit: `var servicesIPs []net.IP`

pkg/sidecar/k8s_reactor.go
```diff
@@ -37,17 +37,26 @@ var (
...
+	manager        *docker.Manager
+	servicesRoutes []net.IP
 }
 
 func NewK8sReactor() (Reactor, error) {
-	redisHost := os.Getenv(EnvRedisHost)
+	// TODO: Generalize this to a list of services.
+	wantedRoutes := []string{
+		os.Getenv(EnvRedisHost),
+		os.Getenv(EnvInfluxdbHost),
```
raulk, line 49: Since this logic is getting more sophisticated, and we will likely want to debug sidecar sooner or later, it would be good to make this a `map[string]string` of `service name => IP/hostname`, so that we can print out the service name in log messages below.

pkg/sidecar/k8s_reactor.go
```diff
@@ -207,18 +220,20 @@ func (d *K8sReactor) manageContainer(ctx context.Context, container *docker.Cont
...
-					logging.S().Debugw("successfully added route", "route.Src", newroute.Src, "route.Dst", newroute.Dst.String(), "gw", newroute.Gw, "container", container.ID)
+			for _, serviceIP := range servicesIPs {
+				if route.Dst.Contains(serviceIP) {
+					newroute := route
+					newroute.Dst = &net.IPNet{
+						IP:   serviceIP,
+						Mask: net.CIDRMask(32, 32),
+					}
+
+					logging.S().Debugw("adding service route", "route.Src", newroute.Src, "route.Dst", newroute.Dst.String(), "gw", newroute.Gw, "container", container.ID)
```
raulk, line 231: It would be useful to log the name of the service we added (see above comment).

pkg/sidecar/k8s_reactor.go
coryschwartz: Does it makes sense to have a single env variable? I'm thinking like a comma or colon-delimeted string 

```
SIDECAR_ALLOW=service1:service2:service3
```

nonsense: We should totally add a mapping from service name to service that we want to allow in the sidecar. I will see about address this in a future PR.

PR #914, opened 2020-04-24
set pod memory and cpu based on group definition

nonsense:
1. Fixing a bug where we don't set the pod CPU and memory based on definition, but we are still using the defaults.

2. Reducing the CPU / memory limits for the `mkdir-outputs` container. It exits right at the start of execution, but no point requesting the defaults (which are 100m and 256Mi).

### DIFF
--- a/pkg/runner/cluster_k8s.go
+++ b/pkg/runner/cluster_k8s.go
@@ -20,12 +20,12 @@ import (
 
 	"golang.org/x/sync/errgroup"
 
+	"github.com/testground/sdk-go/runtime"
 	"github.com/testground/testground/pkg/api"
 	"github.com/testground/testground/pkg/conv"
 	hc "github.com/testground/testground/pkg/healthcheck"
 	"github.com/testground/testground/pkg/logging"
 	"github.com/testground/testground/pkg/rpc"
-	"github.com/testground/sdk-go/runtime"
 
 	v1 "k8s.io/api/core/v1"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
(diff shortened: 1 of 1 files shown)


pkg/runner/cluster_k8s.go
```diff
@@ -733,6 +751,12 @@ func (c *ClusterK8sRunner) createTestplanPod(ctx context.Context, podName string
 							MountPropagation: &mountPropagationMode,
 						},
 					},
+					Resources: v1.ResourceRequirements{
+						Limits: v1.ResourceList{
+							v1.ResourceMemory: resource.MustParse("10Mi"),
```
coryschwartz, line 756: I can't think of any reason to have more than 10Mi for a container that just creates a directory.Looks okay to me.

pkg/runner/cluster_k8s.go
```diff
@@ -219,6 +219,24 @@ func (c *ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow *rpc
 				}},
 		})
 
+		podCPU := defaultCPU
```
coryschwartz, line 222: just a style thing, but we have another way for deriving default configs in the docker runner. Not sure why we use mergo.Merge there and just if-statements in the k8s-runner.

pkg/runner/cluster_k8s.go
nonsense: Probably less is fine as well, but then again this is freed as soon as the container completes, so it is just for `scheduling` calculations.

pkg/runner/cluster_k8s.go
nonsense: I will check how we do that in the local runner and see if we can align in a future PR.

PR #915, opened 2020-04-24
improve CircleCI config; clean up go.mod/go.sum; .dockerignore more files

raulk:
## Changes

* The way we were installing Go in our CI build was going to going to cause problems down the line, particularly when attempting to upgrade the Go version.
  * We can now select the go version via a parameter.
  * It takes 10 sec to install and back, so no need to persist in cache or else.
  * In fact, see notes below. I started persisting in the workspace, but stopped because it was actually slower.

* Dropped the Windows build.

* Improved cache keys.

## Things I evaluated and I discarded

* I evaluated the golang orb. This seemed great at first, but our tests require a Docker daemon. So I enabled the remote docker daemon, but this failed because we bind mount local directories for outputs. So I reverted to using machine executors, which also award us more flexibility for the future.
* I evaluated using workspaces. Unfortunately they were too slow. Takes less time to install Go from the Internet, checkout the project, and restore the caches, than it takes to do that once and save it in the workspace, then load it.


.circleci/config.yml
```diff
@@ -1,106 +1,92 @@
...
-            tar -C ~/ -xzvf ~/go.tar.gz
-
+          name: "make tidy"
+          command: make tidy && git diff --exit-code
+      - run:
+          name: "make mod-download"
+          command: make mod-download
       - save_cache:
-          key: v1-pkg-cache
+          key: 'v2-pkg-cache-{{ checksum "go.sum" }}-{{ .Environment.GOVERSION }}'
```
coryschwartz, line 62: good idea.

.circleci/config.yml
```diff
@@ -1,106 +1,92 @@
...
           paths:
-            - ~/go
+            - << pipeline.parameters.workspace-dir >>/go/<< pipeline.parameters.go-version >>/pkg/mod
```
coryschwartz, line 64: I was hoping to avoid having to download go installation every time. What's faster and more reliable-- restoring cache or downloading from dl.google.com?

raulk: There are workspaces and caches.

Caches survive builds, and we _could_ cache go _as long as_ we embed the go version inside the cache key. We were not doing that before, so even if we upgraded the go version, we would still be left with the old one.

Caches are fast-ish. In practice, it takes 10 sec to download and unpack go. Adding the go distribution to the cache will bloat it up by ~100M, which will also add up to the cache download... All in all, we're looking at single digits seconds gains, not worth it.

OTOH, workspaces are storage area for a single workflow, shared across jobs. They are DAMN SLOW; stay away.

PR #918, opened 2020-04-27
sidecar/k8s: add name of service

nonsense:
This is fixing a bug on `master` where `influxdb` env var is not set and host is `""`.

### DIFF
--- a/pkg/sidecar/k8s_reactor.go
+++ b/pkg/sidecar/k8s_reactor.go
@@ -37,26 +37,37 @@ var (
 )
 
 type K8sReactor struct {
-	client         *sync.Client
-	manager        *docker.Manager
-	servicesRoutes []net.IP
+	client          *sync.Client
+	manager         *docker.Manager
+	allowedServices []AllowedService
 }
 
 func NewK8sReactor() (Reactor, error) {
-	// TODO: Generalize this to a list of services.
-	wantedRoutes := []string{
-		os.Getenv(EnvRedisHost),
-		os.Getenv(EnvInfluxdbHost),
+	wantedServices := []struct {
+		name string
+		host string
+	}{
+		{
+			"redis",
+			os.Getenv(EnvRedisHost),
+		},
+		{
+			"influxdb",
+			os.Getenv(EnvInfluxdbHost),
+		},
 	}
 
-	var resolvedRoutes []net.IP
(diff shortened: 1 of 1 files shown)


pkg/sidecar/k8s_reactor.go
```diff
@@ -37,26 +37,37 @@ var (
...
+	manager         *docker.Manager
+	allowedServices []AllowedService
 }
 
 func NewK8sReactor() (Reactor, error) {
-	// TODO: Generalize this to a list of services.
-	wantedRoutes := []string{
-		os.Getenv(EnvRedisHost),
-		os.Getenv(EnvInfluxdbHost),
+	wantedServices := []struct {
```
nonsense, line 46: As a next improvement, I suggest we resolve the `wanted services` prior to allowing the routes for a given container, with a cache of 1-2min. or so.

This would allow the sidecar to recover relatively fast in case `redis` or `influxdb` or any other service is moved. WDYT?

pkg/sidecar/k8s_reactor.go
```diff
@@ -37,26 +37,37 @@ var (
...
+			os.Getenv(EnvInfluxdbHost),
+		},
 	}
 
-	var resolvedRoutes []net.IP
-	for _, route := range wantedRoutes {
-		ip, err := net.ResolveIPAddr("ip4", route)
+	var resolvedServices []AllowedService
+	for _, s := range wantedServices {
+		if s.host == "" {
```
nonsense, line 62: This is a bug currently on `master` - we are trying to resolve an empty host (`influxdb`).

pkg/sidecar/k8s_reactor.go
raulk: SGTM. We could also resolve once per run_id -- that removes the timing dependency.

That way, containers belonging to a given run_id scheduled on a given node would only resolve the services once. Once all containers are deleted, we would clean up the cache for that run_id.

pkg/sidecar/k8s_reactor.go
raulk: 👍

pkg/sidecar/k8s_reactor.go
nonsense: Sounds good. I will see about updating this in a future PR.

PR #924, opened 2020-04-28
add timeouts on n.cninet.AddNetworkList

nonsense:


pkg/sidecar/k8s_network.go
```diff
@@ -107,12 +107,23 @@ func (n *K8sNetwork) ConfigureNetwork(ctx context.Context, cfg *sync.NetworkConf
...
-			_, err = n.cninet.AddNetworkList(ctx, netconf, rt)
-			return err
-		})
-		if err != nil {
-			return fmt.Errorf("failed to add network through cni plugin: %w", err)
+		errc := make(chan error)
+
+		go func() {
+			err = retry(3, 2*time.Second, func() error {
+				_, err = n.cninet.AddNetworkList(ctx, netconf, rt)
```
nonsense, line 114: I am pretty sure we block here when `weave` goes biserk. I've been seeing the `trying to add a link` log line and nothing after it for a very long time.

pkg/sidecar/docker_reactor.go
```diff
@@ -122,6 +123,12 @@ func (d *DockerReactor) handleContainer(ctx context.Context, container *docker.C
 		return nil, nil
 	}
 
+	if strings.Contains(info.Name, "mkdir-outputs") {
```
nonsense, line 126: Since `containers` in `pods` share the env variables that we set, currently we try to process both containers in the testplan pod - `mkdir-outputs` and the actual testplan container.

If the `mkdir-outputs` doesn't exit right away and sidecar detects it, sometimes we even go as far as trying to add the `weave` network to it, before the context is canceled and the operation is stopped.

nonsense: slightly offtopic, but - now that `testground daemon` is on k8s, we could very well mount EFS on it, and have the cluster_k8s.go runner init the output folders prior to a testplan run.

pkg/docker/manager.go
```diff
@@ -167,7 +168,7 @@ func (m *Manager) Watch(ctx context.Context, worker WorkerFn, labels ...string)
 			handle := m.NewContainerRef(containerID)
 			err := worker(cctx, handle)
 			if err != nil {
-				if errors.Is(err, context.Canceled) {
+				if errors.Is(err, context.Canceled) || strings.Contains(err.Error(), "context canceled") { // docker doesn't wrap errors
```
nonsense, line 171: Because of:

```
Apr 28 19:54:41.924868	[31mERROR[0m	sidecar worker failed: failed to initialise the container: inspect failed: error during connect: Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.40/containers/96b9615ab3ab68667249a32a61f3887123a6461f61874e9089943e4713452157/json": context canceled	{"host": "unix:///var/run/docker.sock", "container": "96b9615ab3ab68667249a32a61f3887123a6461f61874e9089943e4713452157"}
```

pkg/sidecar/k8s_network.go
```diff
@@ -107,12 +107,23 @@ func (n *K8sNetwork) ConfigureNetwork(ctx context.Context, cfg *sync.NetworkConf
...
+			errc <- err
+		}()
+
+		select {
+		case err := <-errc:
+			if err != nil {
+				return fmt.Errorf("failed to add network through cni plugin: %w", err)
+			}
+		case <-time.After(30 * time.Second):
+			return fmt.Errorf("timeout waiting on cninet.AddNetworkList")
```
nonsense, line 126: I've already seen this error when running a build of this version of the code...

pkg/sidecar/k8s_network.go
nonsense: @raulk it'd be interesting to monitor how long this takes in the `good` case, once we get go-metrics in the codebase.

Issue #910, opened 2020-04-23
Ship user-recorded observations and metrics on instance completion


raulk:
See point D of #907.

Issue #909, opened 2020-04-23
Namespace `go-metrics` under `runenv.M()` for aggregate metrics


raulk:
See point C of #907.

Issue #908, opened 2020-04-23
Generate lifecycle events and dispatch them immediately to InfluxDB


raulk:
See point "A. Lifecycle events to be dispatched immediately" of #907.

Issue #755, opened 2020-03-27
Evaluate using a push model with InfluxDB or Elasticsearch as our store.


Robmat05:
We need to look at this evaluation with an unbiased analysis on which application we should use as we decide to move to event based and metrics/stats reporting.

Analysis criteria should include (but not limited to):

- Reliability, at current volume and at scale
- Supplemental work required to on-board application
- Cost, at current volume and at scale
- Comparative performance (robustness)

Robmat05:
1st plan for testing InfluxDB:

1. Set up a InfluxDB Cloud: https://www.influxdata.com/products/influxdb-cloud/
2. Modify the benchmark test plan to stop using Prometheus, and instead record point-in-time observations via RecordMetric.
3. At the end of the test, process the stdout file in the outputs directory, collect all metrics and push them in batches to InfluxDB.
4. See how well this performs.
In the future, we can use Vector to do the log processing and pushing (https://vector.dev/), but right now this is the quickest way to test drive this stack.

coryschwartz:
The infra team has an influxdb cloud account; is there any benefit of having them in the same place, compare test results with real results, etc.

Working on getting the benchmark plan migrated

raulk:
@coryschwartz not really. Any comparison will be done at the Grafana level, which can pull from many data sources. Whatever is quickest.

coryschwartz:
Thanks, I think I'll have a PR to share tonight,  at least good enough to do some basic evaluation

coryschwartz:
@raulk check out the two draft PRs linked here. both are capable of pushing metrics into influxdb, and both are in quite terrible condition, but are functioning.

coryschwartz:
@raulk I can give you a demo of either one of these if you want.

It's easy to push more data than the free tier on influx cloud so I rate-limited the the collect method PR.

Robmat05:
This ticket has come OFF of the priority critical path for the upcoming release as other feature/bugs have leapfrogged its priority.

I am placing this back into  ‘Selected for Dev’ so that we may be able to pick it back up prior to release date once higher priority tasks are resolved.

raulk:
We ended up selecting InfluxDB as our main metrics store. Prometheus has been relegated to infrastructure monitoring only. We may introduce Elasticsearch later to centralise infrastructure logs.

Thanks @coryschwartz and @nonsense for your input.

See #907 for the general direction, and https://github.com/testground/sdk-go/pull/3/ for concrete API design.

PR #934, opened 2020-04-29
inject influxdb env variables; fix sidecar; fix image.

raulk:
Relates to:

https://github.com/testground/testground/issues/755
https://github.com/testground/testground/issues/907
https://github.com/testground/testground/issues/910
https://github.com/testground/testground/issues/908
https://github.com/testground/testground/issues/909

### DIFF
--- a/pkg/config/loader.go
+++ b/pkg/config/loader.go
@@ -72,7 +72,7 @@ func (e *EnvConfig) Load() error {
 		}
 		logging.S().Infof(".env.toml loaded from: %s", f)
 	} else {
-		logging.S().Warnf("no .env.toml found at %s; system may be misconfigured", f)
+		logging.S().Infof("no .env.toml found at %s; running with defaults", f)
 	}
 	return nil
 }
--- a/pkg/runner/local_common.go
+++ b/pkg/runner/local_common.go
@@ -80,7 +80,7 @@ func localCommonHealthcheck(ctx context.Context, hh *healthcheck.Helper, cli *cl
 		healthcheck.StartContainer(ctx, ow, cli, &docker.EnsureContainerOpts{
 			ContainerName: "testground-influxdb",
 			ContainerConfig: &container.Config{
-				Image: "library/influxdb",
+				Image: "library/influxdb:1.8",
 				Env:   []string{"INFLUXDB_HTTP_AUTH_ENABLED=false", "INFLUXDB_DB=testground"},
 			},
 			HostConfig: &container.HostConfig{
--- a/pkg/runner/local_common.go
+++ b/pkg/runner/local_common.go
@@ -80,7 +80,7 @@ func localCommonHealthcheck(ctx context.Context, hh *healthcheck.Helper, cli *cl
 		healthcheck.StartContainer(ctx, ow, cli, &docker.EnsureContainerOpts{
 			ContainerName: "testground-influxdb",
 			ContainerConfig: &container.Config{
-				Image: "bitnami/influxdb",
+				Image: "library/influxdb",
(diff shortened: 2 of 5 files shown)


pkg/config/loader.go
```diff
@@ -72,7 +72,7 @@ func (e *EnvConfig) Load() error {
 		}
 		logging.S().Infof(".env.toml loaded from: %s", f)
 	} else {
-		logging.S().Warnf("no .env.toml found at %s; system may be misconfigured", f)
```
nonsense, line 75: I think this is a warning. If there is no `.env.toml` we are not inferring the `region` for AWS for example, so using `cluster:k8s` would fail.

pkg/runner/local_common.go
```diff
@@ -80,7 +80,7 @@ func localCommonHealthcheck(ctx context.Context, hh *healthcheck.Helper, cli *cl
 		healthcheck.StartContainer(ctx, ow, cli, &docker.EnsureContainerOpts{
 			ContainerName: "testground-influxdb",
 			ContainerConfig: &container.Config{
-				Image: "bitnami/influxdb",
+				Image: "library/influxdb",
```
nonsense, line 83: Probably a good idea to specify the exact tag we are using now.

pkg/runner/local_docker.go
```diff
@@ -225,6 +215,7 @@ func (r *LocalDockerRunner) Run(ctx context.Context, input *api.RunInput, ow *rp
 
 		// Serialize the runenv into env variables to pass to docker.
 		env := conv.ToOptionsSlice(runenv.ToEnvVars())
+		env = append(env, "INFLUXDB_ADDR=http://testground-influxdb:8086")
```
nonsense, line 218: Not sure why we have both `INFLUXDB_HOST` and `INFLUXDB_ADDR`... ?

pkg/runner/local_common.go
raulk: Yeah, should be fine because 1.8+ is good for us, but can do.

pkg/config/loader.go
raulk: It really depends on the use case. For example, I am not running cluster:k8s, just small-scale tests with local:docker, and defaults are fine for me. Maybe we can display the warning on the cluster:k8s runner only. 

What I really wouldn't like to do is to greet every single new user of testground with a warning.

pkg/runner/local_docker.go
raulk: `INFLUXDB_HOST` is the hostname for the sidecar.
`INFLUXDB_ADDR` is the http endpoint that the API client needs.

pkg/config/loader.go
nonsense: > What I really wouldn't like to do is to greet every single new user of testground with a warning.

You have a point there.

pkg/runner/local_docker.go
nonsense: You can infer one from the other, and it'd be good if we reduce to the minimum service discovery parameters necessary to spin up the system, and keep just `INFLUXDB_HOST`. I believe we do the same for REDIS, but we don't send `http://redis:6379` as an env var.

raulk: That would require hardcoding the port number in the SDK. The `INFLUXDB_HOST` doesn't carry the port number. I don't really like doing that (even though we are doing it for Redis, but that's a vestige more than a desire).

nonsense: Yeah, that's what I meant - these dependencies are really the same (albeit one required and the other optional), so the service discovery can work in absolutely the same way. Redis is on 6379, Influx is on 8086 - I don't see anything bad with that.

PR #939, opened 2020-04-30
add unit tests to sidecar

coryschwartz:
This adds tests for the sidecar handler

The handler passes configurations to a mock network

Of course this doesn't test the functionality of either the Docker or K8s networks, just the handler function.


pkg/sidecar/sidecar_test.go
```diff
@@ -0,0 +1,72 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	client, err := sync.NewBoundClient(ctx, runenv)
+	if err != nil {
+		t.Error(err)
+	}
+
+	nw := NewMockNetwork()
+
+	time.Sleep(time.Duration(rand.Int()%10) * time.Second)
```
coryschwartz, line 41: What I am trying to do is avoid having multiple sidecars access the same redis keys. Will fix this.

pkg/sidecar/sidecar_test.go
```diff
@@ -0,0 +1,72 @@
...
+		Network: "mynetwork",
+		State:   "mynetworkdone",
+	}
+	topic := sync.NetworkTopic(inst.Hostname)
+	client.MustPublishAndWait(ctx, topic, netCfg, "mynetworkdone", 1)
+
+	// verify that the handler has configured the network
+	numConf := len(nw.Configured)
+	if numConf != 1 {
+		t.Error("Expected to be configured one time. Actual number of configures:", numConf)
```
coryschwartz, line 67: Right now this is always reporting multiple calls to configure the network, which is probably not right.

pkg/sidecar/sidecar_test.go
```diff
@@ -0,0 +1,139 @@
...
+	rand.Seed(time.Now().UnixNano())
+}
+
+const (
+	CONFIGUREDNETWORK   = "testtesttest_net"
+	CONFIGUREDBANDWIDTH = uint64(12345)
+	NETWORKDONE         = "configureddone"
+)
+
+func unique_runenv() (hostname string, runenv *runtime.RunEnv) {
```
nonsense, line 24:
```suggestion
func newUniqueRunenv() (hostname string, runenv *runtime.RunEnv) {
```

pkg/sidecar/sidecar_test.go
```diff
@@ -0,0 +1,139 @@
...
+		return
+	}
+
+	if doConfigure {
+		netCfg := sync.NetworkConfig{
+			Network: CONFIGUREDNETWORK,
+			State:   NETWORKDONE,
+			Default: sync.LinkShape{
+				Bandwidth: CONFIGUREDBANDWIDTH,
+			},
```
nonsense, line 72: `CallbackState` is required, no? Also this is one of the fields we should document better, and maybe test here as well. WDYT?

pkg/sidecar/sidecar_test.go
```diff
@@ -0,0 +1,139 @@
...
+		for {
+			select {
+			case <-planDone:
+				validator(t, nw)
+				return
+			case err := <-errsCh:
+				if err != nil {
+					t.Error(err)
+					t.FailNow()
+				}
```
nonsense, line 128: I suggest we add a timeout here. If this test fails, we don't want to block the CI job for a long time.

pkg/sidecar/sidecar_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+		errsCh := make(chan error)
+		planDone := make(chan int)
+
+		// Create Sidecar instance
+		client, err := sync.NewBoundClient(ctx, runenv)
+		if err != nil {
+			t.Error(err)
```
nonsense, line 99: `t.Fatal` is just one call, but this is fine as well :)

pkg/sidecar/sidecar_test.go
```diff
@@ -0,0 +1,139 @@
...
+	"strconv"
+	"testing"
+	"time"
+)
+
+func init() {
+	rand.Seed(time.Now().UnixNano())
+}
+
+const (
```
nonsense, line 18:
```suggestion
const (
	// dummy network configuration params that we request and later verify that sidecar has applied.
```

It is not immediately clear what these values are for, so maybe worth adding a comment.

pkg/sidecar/sidecar_test.go
```diff
@@ -0,0 +1,139 @@
...
+	assert.Equal(t, CONFIGUREDNETWORK, con.Network, "expected configured network name to be passed to network driver")
+	assert.Equal(t, CONFIGUREDBANDWIDTH, con.Default.Bandwidth, "expected requested bandwidth to be set")
+}
+
+// planRequestNetworkConfigure is a testground plan.
+// It requests a change to the network configuration via the sync service.
+func planRequestNetworkConfigure(ctx context.Context, runenv *runtime.RunEnv, hostname string, done chan int, doConfigure bool) (err error) {
+	client, err := sync.NewBoundClient(ctx, runenv)
+	if err != nil {
+		return
```
nonsense, line 58:
```suggestion
		return err
```

Nitpick: I see that `err` is named, but just `return` doesn't really convey that you return the error.

nonsense:
Overall LGTM, I added a few comments, but ultimately this is a good start for unit tests for the sidecar, and we verify that the actual handlers are called.

pkg/sidecar/mock.go
```diff
@@ -0,0 +1,60 @@
+package sidecar
+
+import (
+	"context"
+	"errors"
+	"github.com/testground/sdk-go/sync"
+	gosync "sync"
+)
```
raulk, line 8: reorder imports in groups:

```
<stdlib>

<local module>

<testground>

<third party deps>
```

pkg/sidecar/mock.go
```diff
@@ -0,0 +1,60 @@
...
+
+func NewMockNetwork() *MockNetwork {
+	active := make([]string, 0)
+	configured := make([]*sync.NetworkConfig, 0)
+	mux := gosync.Mutex{}
+	return &MockNetwork{
+		Active:     active,
+		Configured: configured,
+		Closed:     false,
+		L:          &mux,
```
raulk, line 30: 1. No need for the mutex pointer here.
2. The convention is to embed the mutex by value, if you don't care about the lock/unlock methods being available publicly. Otherwise the convention is to use a variable name `lk` or `mu`, so that the lock is only available internally.

pkg/sidecar/mock.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+}
+
+// Network
+type MockNetwork struct {
+	Active     []string
+	Configured []*sync.NetworkConfig
+	Closed     bool
+	L          gosync.Locker
```
raulk, line 40: Let's use the mutex hat idiomatic pattern: https://dmitri.shuralyov.com/idiomatic-go#mutex-hat

pkg/sidecar/mock.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func (m *MockNetwork) Close() error {
+	m.Closed = true
```
raulk, line 44: This is racy. We should be taking the lock here.

pkg/sidecar/mock.go
```diff
@@ -0,0 +1,60 @@
...
+	return nil
+}
+
+func (m *MockNetwork) ConfigureNetwork(ctx context.Context, cfg *sync.NetworkConfig) error {
+	if m.Closed {
+		return errors.New("mock network is closed.")
```
raulk, line 50:
```suggestion
		return errors.New("mock network is closed")
```

pkg/sidecar/mock.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	m.L.Lock()
+	m.Configured = append(m.Configured, cfg)
+	m.L.Unlock()
+	return nil
+}
+
+func (m *MockNetwork) ListActive() []string {
+	return m.Active
```
raulk, line 59: When is this modified?

pkg/sidecar/sidecar_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+import (
+	"context"
+	"github.com/stretchr/testify/assert"
+	"github.com/testground/sdk-go/runtime"
+	"github.com/testground/sdk-go/sync"
+	"math/rand"
+	"strconv"
+	"testing"
+	"time"
```
raulk, line 11: import groups, as above.

pkg/sidecar/sidecar_test.go
```diff
@@ -0,0 +1,139 @@
...
+)
+
+func init() {
+	rand.Seed(time.Now().UnixNano())
+}
+
+const (
+	CONFIGUREDNETWORK   = "testtesttest_net"
+	CONFIGUREDBANDWIDTH = uint64(12345)
+	NETWORKDONE         = "configureddone"
```
raulk, line 21: https://golang.org/doc/effective_go.html#mixed-caps

pkg/sidecar/sidecar_test.go
raulk, line 24: Can't we just use https://github.com/testground/sdk-go/blob/207954a995a3933d3c55a0466e483491ffa78e1d/runtime/test_utils.go#L14?

pkg/sidecar/sidecar_test.go
```diff
@@ -0,0 +1,139 @@
...
+			}
+		}
+	}
+}
+
+func TestSidecarConfiguresNetworking(t *testing.T) {
+	// Don't configure the network, accepting the defaults.
+	t.Run("DefaultNetworkConfiguration", subtestSidecarNetworking(verifyDefaultNetwork, false))
+	// Configure the network to change the bandwidth, etc.
+	t.Run("ConfiguredNetwork", subtestSidecarNetworking(verifyConfiguredNetwork, true))
```
raulk, line 138: Just two test cases where each test case uses entirely different parameters is a bit smelly. In my head, these would be two separate `Test...` methods, where `subtestSidecarNetworking` would do some kind of setup, and the asserts would be part of those methods.

There's way too much indirection here IMO.

raulk:
The way that we use the sync service in this test makes me think that we need to implement a stubbed sync service that works entirely locally (in memory).

For that we'll need to extract an interface off `sync.Client`, e.g. `sync.Interface` (following the [`sort.Interface`](https://golang.org/pkg/sort/#Interface) convention), and use in components where we need to stub the actual implementation out, like the sidecar.

In theory that could cause a tiny performance penalty due to dynamic method dispatch at runtime, but I'm pretty sure the compiler will optimize and, also, branch prediction will work in our favour.

Robmat05:
@raulk to add a mock sync.Client that should unblock one of the above change requests

coryschwartz:
> The way that we use the sync service in this test makes me think that we need to implement a stubbed sync service that works entirely locally (in memory).
> 
> For that we'll need to extract an interface off `sync.Client`, e.g. `sync.Interface` (following the [`sort.Interface`](https://golang.org/pkg/sort/#Interface) convention), and use in components where we need to stub the actual implementation out, like the sidecar.
> 
> In theory that could cause a tiny performance penalty due to dynamic method dispatch at runtime, but I'm pretty sure the compiler will optimize and, also, branch prediction will work in our favour.

I definitely agree. We had discussed earlier (and an earlier version of this test) was using https://github.com/alicebob/miniredis

coryschwartz:
I've made changes to https://github.com/testground/sdk-go/pull/16 so this works together. This is substantially different, and simpler now that it can use the InMem client

Of course, this only works for now if you add replace lines to `go.mod` 

To review, switch the `sdk-go` branch to `feat/inmem-client` and configure mod replacements.

pkg/sidecar/sidecar_test.go
```diff
@@ -0,0 +1,93 @@
+package sidecar
```
coryschwartz, line 1: This file is a little repetitive. Could switch to test by table, if it seems too verbose.

pkg/sidecar/sidecar_linux.go
```diff
@@ -19,6 +19,7 @@ const (
 var runners = map[string]func() (Reactor, error){
 	"docker": NewDockerReactor,
 	"k8s":    NewK8sReactor,
+	"mock":   NewMockReactor,
```
coryschwartz, line 22: This will enable us to use the sidecar in CI/kind (although not in any meaningful way)

pkg/sidecar/mock.go
```diff
@@ -0,0 +1,118 @@
+package sidecar
```
coryschwartz, line 1: I have this in the "sidecar" package rather than the "sidecar_test" package and I am a bit torn about this.

I need this to be in the sidecar package in order to use the unexported handler, and as a side benefit we may be able to use this in CI with the real nonInMem sync service if we want to. It would be a NoOp -- but still sync-service-using mode. 

On the other hand, maybe using this in CI isn't all that useful, and by moving it to sidecar_test we could avoid having this compiled in.

@nonsense what do you think?

pkg/sidecar/sidecar_test.go
```diff
@@ -0,0 +1,93 @@
...
+		t.Fatal(err)
+	}
+	ctx := context.Background()
+	r := reactor.(*MockReactor)
+
+	go func() {
+		if err := r.Handle(ctx, handler); err != nil {
+			t.Error(err)
+		}
+	}()
```
nonsense, line 31: This is indeed a bit verbose. We could at least extract into a function, as it is used in all test method, but could be done later as well.

nonsense:
We are waiting for me to issue a release on `testground/sdk-go` so that we update `go.mod`, wait for the tests to pass, and merge this, right? @coryschwartz

coryschwartz:
I've updated the go-mod to use testground/sdk@master  just so we can see the tests pass, but of course we will want to change this tomorrow when the new tag is created.

PR #942, opened 2020-04-30
migrate example & network plan to latest SDK.

raulk:
The new SDK includes improvements in metrics (https://github.com/testground/sdk-go/pulls/3) and in the network API (PR link WIP).


plans/network/main.go
```diff
@@ -210,41 +176,23 @@ func pingpong(runenv *runtime.RunEnv) error {
...
-		}
 		return nil
 	}
-	err = pingPong("200", 200*time.Millisecond, 210*time.Millisecond)
+	err = pingPong("200", 200*time.Millisecond, 215*time.Millisecond)
 	if err != nil {
 		return err
 	}
 
 	config.Default.Latency = 10 * time.Millisecond
```
nonsense, line 190: Offtopic, but this is a bit too granular to measure consistently in order environments, but locally.

raulk: Yup, we should probably set latencies that are orders of magnitude different, so that our leniency ranges can also be laxer.

Issue #778, opened 2020-04-01
testground daemon blocks after testplan run unexpectedly


nonsense:
```
➜  testground run single benchmarks/subtree \
    --builder=docker:go \
    --runner=cluster:k8s \
    --build-cfg bypass_cache=true \
    --build-cfg push_registry=true \
    --build-cfg registry_type=aws \
    --run-cfg keep_service=false \
    --instances=5000 --collect
```

```
Apr  1 15:23:43.725307  INFO    all testplan instances in `Succeeded` state     {"req_id": "4067a57e", "runner": "cluster:k8s", "run_id": "1b6df03133b6", "took": "7m38.730703885s"}
Apr  1 15:26:05.268001  INFO    run finished successfully       {"req_id": "4067a57e", "plan": "benchmarks", "case": "subtree", "runner": "cluster:k8s", "instances": 5000}
Apr  1 15:26:05.268302  DEBUG   request handled {"ruid": "4067a57e", "command": "run"}
```

For some reason the `testground daemon` blocked for 2min after `all testplan instances in Succeeded state`. AFAICT we are supposed to just return the runid after that, and we shouldn't wait for anything.

---

On the `testground client` side I saw: 

```
Apr  1 15:23:43.725209  DEBUG   testplan pods state     {"req_id": "4067a57e", "runner": "cluster:k8s", "run_id": "1b6df03133b6", "running_for": "7m38.730605118s", "succeeded": 5000, "running": 0, "pending": 0, "failed": 0, "unknown": 0}
Apr  1 15:23:43.725307  INFO    all testplan instances in `Succeeded` state     {"req_id": "4067a57e", "runner": "cluster:k8s", "run_id": "1b6df03133b6", "took": "7m38.730703885s"}

unexpected EOF
```

---

Current commit is: `0f974d2`

raulk:
Code has changed quite a bit, and there have been no more reports of this issue. Please capture a goroutine dump when this happens the next time -- otherwise it's hard to debug as it's not easily reproduceable. Reopen if it reappears.

nonsense:
Actually I found out, what this is about - we are waiting for the k8s API to clear up all the pods from `Completed` state, and this takes a while.

PR #943, opened 2020-04-30
remove cpu limits for testplan pods

nonsense:


pkg/runner/cluster_k8s.go
```diff
@@ -304,6 +304,9 @@ func (c *ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow *rpc
 		}
 	}
 
+	if !cfg.KeepService {
```
nonsense, line 307: Related to: https://github.com/testground/testground/issues/778 (+ http timeouts on daemon)

raulk:
I support this. I don’t see CPU limits being useful really generally — other than in specific test plans where you want to deliberately choke the thing being tested.

Memory limits are important to trigger garbage collection at sane points.

Also, from @nonsense:

> memory limits are important so that k8s kills our pods with OOM, and not the kernel
> if we leave it to the kernel, the kernel might kill anything
> and the host goes biserk

PR #950, opened 2020-05-01
fix healthcheck feedback to user

nonsense:
It looks like we are not accounting for the case where a specific `healthcheck` is failing, but we don't have a `fixer` for it - for example `redis` disappearing on our Kubernetes cluster.

In this case currently we respond to the user that `healthchecks: ok` which is misleading, so I suggest we at least give back a `warning` and continue with the run. (i.e. no change in behaviour)

After seeing the warning the user can call `testground healthcheck --runner cluster:k8s` and get a detailed report on what appears to be broken.

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -154,9 +154,11 @@ func (e *Engine) DoBuild(ctx context.Context, comp *api.Composition, basesrc str
 			return nil, fmt.Errorf("healthcheck and fix errored: %w", err)
 		} else if !rep.FixesSucceeded() {
 			return nil, fmt.Errorf("healthcheck fixes failed; aborting:\n%s", rep)
+		} else if !rep.ChecksSucceeded() {
+			ow.Warnf(aurora.Bold(aurora.Yellow("some healthchecks failed, but continuing")).String())
+		} else {
+			ow.Infof(aurora.Bold(aurora.Green("healthcheck: ok")).String())
 		}
-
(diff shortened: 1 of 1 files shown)


pkg/engine/engine.go
```diff
@@ -154,9 +154,11 @@ func (e *Engine) DoBuild(ctx context.Context, comp *api.Composition, basesrc str
 			return nil, fmt.Errorf("healthcheck and fix errored: %w", err)
 		} else if !rep.FixesSucceeded() {
 			return nil, fmt.Errorf("healthcheck fixes failed; aborting:\n%s", rep)
+		} else if !rep.ChecksSucceeded() {
+			ow.Warnf(aurora.Bold(aurora.Yellow("some healthchecks failed, but continuing")).String())
```
raulk, line 158: nitpick: isn't this printed out as a final result? "continuing" doesn't make sense when this healthcheck is invoked with `testground healthcheck`, i.e. it's a terminal operation in itself.

nonsense: I am not 100% sure this is printed when `testground healthcheck` is called, let me check.

`testground healthcheck` is showing a detailed report of all healthchecks.

nonsense: @raulk actually no need to check. It is obvious this is part of `DoBuild` and `DoRun`.

raulk: duh, I had overlooked the context

PR #957, opened 2020-05-04
amend example .env.toml

nonsense:
No need to confuse our users with defaults for `cluster:swarm` runner, that we no longer support.


env-example.toml
```diff
@@ -23,13 +23,6 @@ access_token = "docker hub access token"
 #   3. Test plan definition.
```
raulk, line 23:
```suggestion
#   3. Test plan manifest.
```

PR #958, opened 2020-05-04
remove aws credentials from sample .env.toml

nonsense:
I suggest we remove these from the sample `.env.toml`, because it is not normal for users to have to enter these at all.

Usually we pick up the default credentials from the host or the IAM policy.

I personally have never had to specify my credentials in this file, so I am not sure in what use-case one would want to do that.

### DIFF
--- a/env-example.toml
+++ b/env-example.toml
@@ -5,8 +5,6 @@
 # which may be used by several components.
 #
 ["aws"]
-access_key_id = "<aws access key id>"
-secret_access_key = "<aws secret access key>"
 region = "<aws region>"
 
 ["dockerhub"]


env-example.toml
```diff
@@ -5,8 +5,6 @@
 # which may be used by several components.
 #
 ["aws"]
-access_key_id = "<aws access key id>"
-secret_access_key = "<aws secret access key>"
 region = "<aws region>"
```
raulk, line 8: It's odd that this TOML section now only has the region. Don't we specify it as a env var as well?

nonsense: It is indeed odd. I am not sure why the current version of the code is not picking it up automatically. We have to investigate, and maybe get rid of this altogether.

PR #961, opened 2020-05-05
Put getting started above "how does it work"

momack2:
getting started is short! "how it works" is long. Suggest putting this first so it's easy for folks to scroll past, and then back to. This could also benefit from a few sentences added to the getting started to orient folks on WHAT they're trying to do and HOW to do it.


raulk:
I'll take care of the above!

PR #964, opened 2020-05-05
clean up the CLI.

raulk:
* use imperative rather than descriptive style in usage texts.
* use placeholders where possible.
* rename `testground plan import {--source => --from}`
* sort top-level commands alphabetically.
* general clean-up, unification, sanity.


pkg/cmd/healthcheck.go
```diff
@@ -12,12 +12,12 @@ import (
 
 var HealthcheckCommand = cli.Command{
 	Name:   "healthcheck",
-	Usage:  "checks, and optionally heals, the preconditions for the runner to be able to run properly",
+	Usage:  "validate/heal the preconditions for the runner to be able to operate properly",
```
nonsense, line 15:
```suggestion
	Usage:  "validate/fix the preconditions for the runner to be able to operate properly",
```

nonsense:
Overall looks good to me, but haven't tested it thoroughly.

pkg/cmd/healthcheck.go
coryschwartz: +1 for fix

pkg/cmd/plan.go
```diff
@@ -23,56 +23,57 @@ import (
...
 	Subcommands: cli.Commands{
 		&cli.Command{
 			Name:  "create",
-			Usage: "create a plan named `PLAN_NAME`",
+			Usage: "creates a new test plan, using a template from github.com/testground/plan-templates",
 			Flags: []cli.Flag{
 				&cli.StringFlag{
 					Name:     "remote",
-					Usage:    "origin for your repository i.e. git@github.com:your/repo",
+					Usage:    "`URL` of the repo where this plan will be hosted i.e. git@github.com:your/repo",
```
coryschwartz, line 34: I intentionally didn't use URL here since git origins are not necessarily URLs. They might be SSH, or paths to local bare repos, etc.

pkg/cmd/root.go
```diff
@@ -26,6 +34,6 @@ var RootFlags = []cli.Flag{
 	},
 	&cli.StringFlag{
 		Name:  "endpoint",
-		Usage: "set the daemon endpoint URI (overrides .env.toml)",
+		Usage: "set the daemon endpoint `URI` (overrides .env.toml)",
```
coryschwartz, line 37: The backticks have syntactic significance, by the way.  https://github.com/urfave/cli/blob/master/docs/v2/manual.md#placeholder-values

raulk: Yup, I'm precisely adding them for that reason!

pkg/cmd/plan.go
raulk: Funny thing is that git docs calls them URLs!

image

PR #965, opened 2020-05-05
require flags on single run

coryschwartz:
I've gone through and set all those compositions which make up the `Global` composition struct as required flags.

Failure to add these flags causes a message to be printed:

```
Required flags "builder, plan, instances, runner, testcase" not set
```

This is because some options, "builder", "runner", "instances"  require command-line flags even when they have default values in the manifest and setting them as required flags will reduce confusion for the user when the composition cannot be constructed.


nonsense:
Something changed along the way and now the behaviour is different. This PR LGTM, but if we are going to specify to mark these arguments as required, we might as well get rid of the defaults.

raulk:
@nonsense I don't think we support default builders and runners anyway. We require the user to be completely explicit about how they want to build/run the test plan; no more spooky action at a distance.

PR #971, opened 2020-05-06
add toml annotation to selectors

nonsense:


raulk:
@nonsense @coryschwartz this is cool, but some compositions (e.g. DHT and pubsub) will break -- also check the examples/docs just in case! Users had been living with the `Selectors` gotcha without complaining.

nonsense:
> @nonsense @coryschwartz this is cool, but some compositions (e.g. DHT and pubsub) will break -- also check the examples/docs just in case! Users had been living with the `Selectors` gotcha without complaining.

Good point, I didn't consider that this is actually not backward-compatible. Shall we revert, until we decide how to fix properly?

nonsense:
@raulk we appear to not be using `selectors` in our examples. Maybe we should just get the DHT and pubsub teams to update their compositions and have `selectors` instead of `Selectors` inside composition files?

raulk:
Yes, that works. Can you open an issue on `libp2p/test-plans` and the pubsub repo? Would be awesome if you could scan the `ipfs/test-plans` repo for usages of selectors in compositions? (I don't think the bitswap tests use selectors, but just in case cc @dirkmc).

Example in the DHT compositions: https://github.com/libp2p/test-plans/blob/master/dht/compositions/all-both-k8s.toml#L21 (although these will fail for other reasons: they use pre-built artefacts, and some reference commits that no longer exist)

nonsense:
Added a PR to `libp2p/test-plans`

`ipfs/test-plans` is not using any selectors.

PR #977, opened 2020-05-07
ipam.IPs string slice

nonsense:


laurentsenta:
@nonsense I lack context on this change, is this a draft we still want to bring to completion?

nonsense:
I never managed to merge this upstream in Weave, but this fork is necessary in order to have the following functionality, if Weave is used as a data network:

```
It allows for specifying an IP address through the CNI plugin, similar to weave attach and claim that IP.
```

Without this patch, it was not possible to attach a *specific* IP to a container, and this was one of the features of Testground back in the day.

PR #980, opened 2020-05-07
artifacts

coryschwartz:
This is for discussion on how we want to bring in artifacts (potentially for Filecoin)

This is not what we want this to look like in the final version.

This should never be merged. It is a placeholder for discussion and design purposes


coryschwartz:
Closed, as this was a quick and dirty experiment for adding "artifacts". 

With https://github.com/testground/testground/pull/988 now merged, users who want to have files embedded in their image can use the docker:generic builder.

We discussed the possibility of adding larger multi-GB files which would be unreasonable to include in an image -- I'm going to open an issue to explore what we need to do to support filecoin proofs and other large-asset use cases.

Issue #983, opened 2020-05-08
docker:go builder: add extension hooks to Dockerfile template


raulk:
Some projects need to customize the builds of the test plan images, e.g. Filecoin Lotus performs FFI calls over to Rust proofs code. They need to add the relevant libraries to the build container.

While we plan to implement a `docker:generic` builder (#692) shortly that enables test plan writers to supply their Dockerfile entirely, not use cases require such level of advanced configuration.

Some cases require _some_ level of tuning, and having to supply an entire Dockerfile is overkill, and most of the content would be copy-pasted.

## Proposal

- We already use a Go template to produce the Dockerfile. Add extension hooks to this template, along with builder parameters, so that test plan manifests can inject Dockerfile directives at different points.
- Allow users to select the base image of the runtime container, defaulting to `busybox:1.31.1-glibc` if not specified.

PR #984, opened 2020-05-08
docker:go builder: support Dockerfile extensions and CGO.

raulk:
Fixes https://github.com/testground/testground/issues/983.

---

* `docker:go` builder configuration:

  * Add a `DockerfileExtensions` field, with nested fields such that 1 field = 1 hook.
  * Add a `RuntimeImage` field so that the user can specify the runtime image, defaulting to `busybox:1.31.1-glibc`. 

* Created a `dockercustomize` with echos on all extension points with which you can manually verify that hooks are being injected properly.
* After fiddling with `ARG`, `ENV`, and `RUN echo`, I couldn't find a way to output the runtime image being used, for verification. But if you run the `dockercustomize` build, watch the Docker output, and perform a `docker images ls | grep <image_id>`, you'll see that indeed the override is taking effect and `debian` is picked. Changing it to `debiannnnn` will make the build fail.

```
Step 20/28 : FROM ${RUNTIME_IMAGE} AS binary
 ---> 58075fe9ecce
[...]
⟩ docker image list  | grep 58075fe9ecce
debian                      buster              58075fe9ecce        5 weeks ago         114MB
debian                      latest              58075fe9ecce        5 weeks ago         114MB
```

* removed the `overridable` struct tag -- it's leftover garbage from the past.


nonsense:
I find this rather complicated - not only you need to know what a Dockerfile is, but you also have to know exactly where those 8 hooks go into our Dockerfile in order to produce a build:
- you have to review and understand the existing Dockerfile
- you have to make sure you're hook play well with it

I think this is a few times more complicated that just writing your own Dockerfile.

I don't want to sound harsh, but I really don't see how this can be useful to anyone, who has spend more than a day working with Docker. It introduces more complexity that value in my opinion.

nonsense:
Apart from the indirection this introduces, technically it would work.

raulk:
**We need to document what the hooks are.** Just like notification/callback systems, it's important to document when those notifications are called.

> I really don't see how this can be useful to anyone, who has spend more than a day working with Docker. It introduces more complexity that value in my opinion.

Simple examples:

* adding an extra binary to the final image (remember when we were shipping IPFS in all images; well, with this, a test plan requiring IPFS can just add 1-2 lines to perform the wget and inject it into the resulting image, without having to maintain an entirely separate Dockerfile).
* adding static libraries that need to be present at runtime.
* adding debugging tools, e.g. networking tools.
* copying input assets (e.g. test vectors).
* fetching an extra dependency (the case of Filecoin).

For all those scenarios, forcing the user to copy a Dockerfile and having to maintain it over time as the boundary/contract between Testground and the test plan evolves, is overkill.

---

As I've said elsewhere, this doesn't solve all use cases, but it solves those where the amount of instructions you want to inject is small and pretty delimited. We're trying to offer options, not a one-shot silver bullet for everything.

Only fall down to the basics (managing your own Dockerfile) + the overhead that comes with it (keeping it in sync with new versions of Testground) when you truly need it.

raulk:
In a nutshell, I'm trying to avoid our answer to the question:

> How can I just add one or two extra steps to the build?

being: "you have to fork the Dockerfile entirely, maintain it over time, and switch to a different builder".

^^ That would be terrible UX, and a last resort if you ask me.

Issue #692, opened 2020-03-17
Bring your own Dockerfile builder


coryschwartz:
I know this was here at one time, and was done away with. If there was a builder for which the user could supply their own Dockerfile, plan writers could more easily modify the test environment. 

Bifrost runs ipfs with a web proxy in front of it, in some cases with url rewriting.

Lots of home users will run ipfs from behind firewalls.

Additionally, infrastructure applications could be added by test composition.
For example, browser IPFS may require additional proxies, preloaders, or other infrastructure which is not provided by the current test suite, and need not be provided most of the time. With a "bring your own Dockerfile" builder, this additional infrastructure could be added as a build composition.

coryschwartz:
This issue is similar to https://github.com/ipfs/testground/issues/49

I had a slightly different implementation in mind when I wrote this, than what they have in that ticket, but either method would accomplish the same goal.

Robmat05:
Should this be something that lives in our v0.4 sprint backlog? Or is this just a kind of reminder that this needs work in a future sprint. Just a bit more info on where this should live would be helpful to understanding our work load for the current sprint

Issue #644, opened 2020-03-09
Move image push logic from `docker:go` builder to `cluster:k8s` runners


raulk:
This has already been a footgun for a few people. If a runner needs its image to be shipped somewhere, it should be doing it automatically. The `docker:go` builder doesn't know if the runner needs it pushed or not, that's why it's a footgunny option right now.

PR #988, opened 2020-05-09
docker generic builder

coryschwartz:
fix: https://github.com/testground/testground/issues/692

Users can include their own Dockerfile along with the plan.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,11 +79,8 @@ jobs:
     steps:
       - setup
       - run:
-          name: "make build-all"
-          command: make build-all
-      - run:
-          name: "make docker"
-          command: make docker
+          name: "install"
+          command: make install
       - run:
           name: "make test"
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -87,6 +87,6 @@ jobs:
       - run:
           name: "make test"
           command: |
-            mkdir -p $HOME/testground/plans
-            ln -s $HOME/project/plans/placebo $HOME/testground/plans/placebo
+            testground plan import --from $HOME/project/plans/placebo
+            testground plan import --from $HOME/project/plans/example
             go test -v ./...
--- a/pkg/build/common/util.go
+++ b/pkg/build/common/util.go
@@ -1,0 +1,99 @@
+package common
+
+import (
+	"context"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+
+	"github.com/testground/testground/pkg/api"
+	"github.com/testground/testground/pkg/aws"
+	"github.com/testground/testground/pkg/docker"
+	"github.com/testground/testground/pkg/rpc"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/client"
+)
+
+func PushToAWSRegistry(ctx context.Context, ow *rpc.OutputWriter, client *client.Client, in *api.BuildInput, out *api.BuildOutput) error {
+	// Get a Docker registry authentication token from AWS ECR.
+	auth, err := aws.ECR.GetAuthToken(in.EnvConfig.AWS)
+	if err != nil {
+		return err
+	}
+
+	// AWS ECR repository name is testground-<region>-<plan_name>.
+	repo := fmt.Sprintf("testground-%s-%s", in.EnvConfig.AWS.Region, in.TestPlan)
+
+	// Ensure the repo exists, or create it. Get the full URI to the repo, so we
+	// can tag images.
+	uri, err := aws.ECR.EnsureRepository(in.EnvConfig.AWS, repo)
+	if err != nil {
+		return err
+	}
+
+	// Tag the image under the AWS ECR repository.
+	tag := uri + ":" + in.BuildID
+	ow.Infow("tagging image", "tag", tag)
+	if err = client.ImageTag(ctx, out.ArtifactPath, tag); err != nil {
+		return err
+	}
+
+	// TODO for some reason, this push is way slower than the equivalent via the
+	// docker CLI. Needs investigation.
+	ow.Infow("pushing image", "tag", tag)
+	rc, err := client.ImagePush(ctx, tag, types.ImagePushOptions{
+		RegistryAuth: aws.ECR.EncodeAuthToken(auth),
+	})
+	if err != nil {
+		return err
+	}
+
+	// Pipe the docker output to stdout.
+	if err := docker.PipeOutput(rc, ow.StdoutWriter()); err != nil {
+		return err
+	}
+
+	// replace the artifact path by the pushed image.
+	out.ArtifactPath = tag
+	return nil
+}
+
+func PushToDockerHubRegistry(ctx context.Context, ow *rpc.OutputWriter, client *client.Client, in *api.BuildInput, out *api.BuildOutput) error {
+	uri := in.EnvConfig.DockerHub.Repo + "/testground"
+
+	tag := uri + ":" + in.BuildID
+	ow.Infow("tagging image", "source", out.ArtifactPath, "repo", uri, "tag", tag)
+
+	if err := client.ImageTag(ctx, out.ArtifactPath, tag); err != nil {
+		return err
+	}
+
+	auth := types.AuthConfig{
+		Username: in.EnvConfig.DockerHub.Username,
+		Password: in.EnvConfig.DockerHub.AccessToken,
+	}
+	authBytes, err := json.Marshal(auth)
+	if err != nil {
+		return err
+	}
+	authBase64 := base64.URLEncoding.EncodeToString(authBytes)
+
+	rc, err := client.ImagePush(ctx, uri, types.ImagePushOptions{
+		RegistryAuth: authBase64,
+	})
+	if err != nil {
+		return err
+	}
+
+	ow.Infow("pushed image", "source", out.ArtifactPath, "tag", tag, "repo", uri)
+
+	// Pipe the docker output to stdout.
+	if err := docker.PipeOutput(rc, ow.StdoutWriter()); err != nil {
+		return err
+	}
+
+	// replace the artifact path by the pushed image.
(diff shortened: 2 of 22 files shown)


pkg/build/generic/docker.go
```diff
@@ -0,0 +1,100 @@
...
+	defer cancel()
+
+	opts := types.ImageBuildOptions{
+		Tags:        []string{id, in.BuildID},
+		BuildArgs:   cfg.BuildArgs,
+		NetworkMode: "host",
+	}
+
+	imageOpts := docker.BuildImageOpts{
+		BuildCtx:  plansrc,
```
coryschwartz, line 62: I'm slightly torn about this -- the base for docker:go is one level higher than this so the plan can sit next to the SDK. Maybe I should do this here as well...

The reason I decided to do it this way is that the user can run "docker build" from the plan directory and the the exact result that the daemon will build and that seems intuitive and unsurprising.

hacdias:
This looks rad, I tested it and it works 😃

pkg/build/common/util.go
```diff
@@ -0,0 +1,99 @@
+package common
```
raulk, line 1: Packages `common` and files `util` are frowned upon in Go. Let's get rid of the `pkg/build/golang` nesting, so that it's at parity with `pkg/runner` -- wanted to do this for v0.5 but didn't get around to do the busywork. Then all builders will be in a flat package, and we can name this file `registries.go`, which is what these functions are about. I'll file a PR for the shifting around.

pkg/build/generic/docker.go
```diff
@@ -0,0 +1,100 @@
+package generic
```
raulk, line 1: With the new package structure I proposed above, this file can be in the `pkg/build` flat package, and be called `docker_generic.go`.

pkg/build/generic/docker.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/testground/testground/pkg/build/common"
+	"github.com/testground/testground/pkg/docker"
+	"github.com/testground/testground/pkg/rpc"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/client"
+)
+
+const (
+	buildNetworkName = "testground-build"
```
raulk, line 19: Can we make this a package constant? It's used elsewhere.

pkg/build/generic/docker.go
raulk: * I think for homogeneity with other builders, we should absolutely place the plan's source under `/plan`.
* I don't think the `docker:generic` builder can support SDK linkage, because the way you link the SDK will depend entirely on the language the plan it's built with. With go, it's with `replace` directives; with JS/node/npm/Rust/Java, it will be different.

pkg/build/generic/docker.go
```diff
@@ -0,0 +1,100 @@
...
+		cli, err = client.NewClientWithOpts(cliopts...)
+	)
+
+	ow = ow.With("build_id", id)
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
+	defer cancel()
+
+	opts := types.ImageBuildOptions{
+		Tags:        []string{id, in.BuildID},
+		BuildArgs:   cfg.BuildArgs,
```
raulk, line 57: A comment for our future selves:

```suggestion
		BuildArgs:   cfg.BuildArgs,    // ok if nil
```

pkg/build/generic/docker.go
raulk: This const is not used in this file.

pkg/build/generic/docker.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+type DockerGenericBuilder struct {
+	Enabled bool
+}
+
+type DockerGenericBuilderConfig struct {
+	BuildArgs    map[string]*string `toml:"build_args"`
+	PushRegistry bool               `toml:"push_registry"`
+	RegistryType string             `toml:"registry_type"`
```
raulk, line 33: It's OK to get this in, but it becomes more apparent that this shouldn't be in the builders. How would you feel about working on https://github.com/testground/testground/issues/644 straight after we've merged this? We can then remove the push from here and from `docker:go`.

plans/example/manifest.toml
```diff
@@ -4,6 +4,13 @@ name = "example"
 builder = "exec:go"
 runner = "local:exec"
 
+[builders."docker:generic"]
+  enabled = true
+  [builders."docker:generic".build_args]
+  build_image = "golang:alpine"
+  run_image = "alpine"
```
raulk, line 11: These `build_args` aren't actually used in the Dockerfile.

Can you add a few `RUN echo $ARG1` instructions in the Dockerfile and set them here?

raulk:
@coryschwartz there are lint failures. Mind taking a look?

raulk:
PR to flatten the package `pkg/build` package: https://github.com/testground/testground/pull/989.

pkg/build/generic/docker.go
coryschwartz: That's for sure. Re-implementing the image-push logic in two builders builders is just rediculous! Yeah I can take a stab at https://github.com/testground/testground/issues/644 after this.

pkg/build/generic/docker.go
coryschwartz: Right,  removed from this file  in the new flattened version.

coryschwartz:
I added args to the example Dockerfile so it is visible what that args are being applied.

I'm using the args to control the images -- which is what I imagine a plan writer might do to switch the code version in the build. This is how it looks when the image is built from nothing:

```
testground run single --builder docker:generic --runner local:docker --plan example --testcase output --instances 1

Step 1/10 : ARG build_image
Step 2/10 : ARG run_image
Step 3/10 : FROM ${build_image} AS builder
 ---> dda4232b2bd5
Step 4/10 : WORKDIR /plan
 ---> Using cache
 ---> 16122fe1e1b8
Step 5/10 : COPY . .
 ---> df784c4ba500
Step 6/10 : RUN go build -o /testplan
 ---> Running in 779740552e1d
go: downloading github.com/testground/sdk-go v0.2.0
go: downloading github.com/go-redis/redis/v7 v7.2.0
go: downloading go.uber.org/zap v1.14.1
go: downloading github.com/influxdata/influxdb1-client v0.0.0-20191209144304-8bf82d3c094d
go: downloading github.com/dustin/go-humanize v1.0.0
go: downloading github.com/hashicorp/go-multierror v1.1.0
go: downloading github.com/prometheus/client_golang v1.5.1
go: downloading github.com/avast/retry-go v2.6.0+incompatible
go: downloading github.com/rcrowley/go-metrics v0.0.0-20200313005456-10cdbea86bc0
go: downloading github.com/hashicorp/errwrap v1.0.0
go: downloading github.com/prometheus/common v0.9.1
go: downloading github.com/prometheus/client_model v0.2.0
go: downloading github.com/cespare/xxhash/v2 v2.1.1
go: downloading github.com/golang/protobuf v1.3.2
go: downloading github.com/prometheus/procfs v0.0.8
go: downloading go.uber.org/atomic v1.6.0
go: downloading go.uber.org/multierr v1.5.0
go: downloading github.com/beorn7/perks v1.0.1
go: downloading github.com/matttproud/golang_protobuf_extensions v1.0.1
 ---> 81eacad159bd
 ---> 81eacad159bd
Step 7/10 : FROM ${run_image}
 ---> f70734b6a266
Step 8/10 : COPY --from=builder /testplan /testplan
 ---> 2cf2822ec146
Step 9/10 : EXPOSE 6060
 ---> Running in 9639cb8613ce
 ---> 696ce63185b3
Step 10/10 : ENTRYPOINT [ "/testplan"]
 ---> Running in 511a0598f0f4
a ---> f3de369144e4
 ---> f3de369144e4
Successfully built f3de369144e4
Successfully tagged 15d876ab3419:latest
```

plans/example/manifest.toml
coryschwartz: I forgot to add these in the actual Dockerfile that got included in this PR but added now so you can see they are used

pkg/build/generic/docker.go
coryschwartz: Per this comment, I've moved the BuildCtx to basesrc rather than plansrc.

As a result, the Dockerfile has to be written slightly differently -- everything has to reference `plan/___`  rather than just plane `./` which to me seems a little unusual. Maybe this would be different if the Dockerfile were uploaded inside the plan directory. Let me know if this seems more or less confusing!

plans/example/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+ARG run_image
+
+FROM ${build_image} AS builder
+WORKDIR /build
+COPY . .
+RUN cd plan && go build -o /testplan 
+
+FROM ${run_image}
+COPY --from=builder /testplan /testplan
+COPY plan/artifact.txt /
```
coryschwartz, line 11: Including additional files in the image. Notice the path is "plan/artifact.txt", because the base context is one directory higher.

.circleci/config.yml
```diff
@@ -79,14 +79,11 @@ jobs:
...
-          command: make docker
+          name: "install"
+          command: make install
       - run:
           name: "make test"
           command: |
-            mkdir -p $HOME/testground/plans
-            ln -s $HOME/project/plans/placebo $HOME/testground/plans/placebo
+            testground plan import --from $HOME/project/plans/placebo
+            testground plan import --from $HOME/project/plans/example
```
raulk, line 88: nice

plans/example/Dockerfile
nonsense: Why is the base context one directory higher? I am not sure I understand this.

.circleci/config.yml
Robmat05: I’ve create #1010 for Cory to update relevant docs

coryschwartz: Doc-ed!

PR #989, opened 2020-05-12
flatten pkg/build package.

raulk:
Shuffle around a few things and flatten the `pkg/build` package to make it symmetric to `pkg/runner`.


raulk:
This is pretty uncontroversial and does not affect functionality, so I'm merging. Shout after the fact if something doesn't add up.

PR #990, opened 2020-05-12
extract `default` network name

nonsense:
`default` is a special name of the only data network we currently set up and allow traffic shaping on.

Therefore it is best if we extract it and amend the docs at https://docs.testground.ai/traffic-shaping#configure-traffic-shaping to use `network.DefaultDataNetwork` and explain a bit better what this value is, together with `CallbackState`.

### DIFF
--- a/pkg/sidecar/k8s_network.go
+++ b/pkg/sidecar/k8s_network.go
@@ -17,6 +17,10 @@ import (
 	"github.com/vishvananda/netlink"
 )
 
+const (
+	defaultDataNetwork = "default"
+)
+
 type k8sLink struct {
 	*NetlinkLink
 	IPv4, IPv6 *net.IPNet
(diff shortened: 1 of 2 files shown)


pkg/sidecar/k8s_network.go
```diff
@@ -17,6 +17,10 @@ import (
 	"github.com/vishvananda/netlink"
 )
 
+const (
+	defaultDataNetwork = "default"
```
raulk, line 21: I think we can afford to make this public and add godoc.

nonsense: No need for this to be public - we need it public on the SDK side only, no package is currently using this particular one.

nonsense:
Now that both PRs are merged, I will update the docs to be a bit more user-friendly in terms of which fields mean what, and how they can do traffic shaping on Testground.

PR #991, opened 2020-05-12
relax latency expectations on network/ping-pong test

nonsense:
@hacdias reported that he got failures on `local:docker`, such as:

```
expected an RTT between 20ms and 35ms, got 48.096866ms	{"req_id": "3601308d"}
```

Therefore I am relaxing a bit the RTT expectation to be within 100ms - this should work consistently on CI environments and locally I guess.

### DIFF
--- a/plans/network/main.go
+++ b/plans/network/main.go
@@ -182,17 +182,17 @@ func pingpong(runenv *runtime.RunEnv) error {
 
 		return nil
 	}
-	err = pingPong("200", 200*time.Millisecond, 215*time.Millisecond)
+	err = pingPong("200", 200*time.Millisecond, 300*time.Millisecond)
 	if err != nil {
 		return err
 	}
 
-	config.Default.Latency = 10 * time.Millisecond
+	config.Default.Latency = 50 * time.Millisecond
 	config.CallbackState = "latency-reduced"
 	netclient.MustConfigureNetwork(ctx, config)
 
 	runenv.RecordMessage("ping pong")
-	err = pingPong("10", 20*time.Millisecond, 35*time.Millisecond)
+	err = pingPong("100", 100*time.Millisecond, 200*time.Millisecond)
 	if err != nil {
 		return err
 	}


raulk:
Ultimately Testground can do no more than call out to kernel traffic shaping utilities. But making a 200ms latency allowable with a 100ms setting is not acceptable for a platform like Testground. Maybe these failures are uncovering something we do wrong at the Netlink level.

nonsense:
> But making a 200ms latency allowable with a 100ms setting is not acceptable for a platform like Testground.

@raulk We have 200ms latency, because you have a RTT of 100ms, so 2*200ms.

nonsense:
@raulk @hacdias reported the following results on his local machine:

```
expected an RTT between 20ms and 35ms, got 48.096866ms	{"req_id": "3601308d"}
```

How much do you suggest we set these values, so that we get consistent tests on local machines and on CircleCI? We don't want this test to fail on CircleCI once we start running it.

I see you are not happy with 100ms, but it looks like 28ms is also not good.

nonsense:
Note that I am setting these values based on experience with Travis. I don't have good intuition on how this will work on CircleCI, but I imagine it is similar.

raulk:
> @raulk We have 200ms latency, because you have a RTT of 100ms, so 2*100ms.

The patch does the following -- setting a latency of 50ms, where the perfect RTT would be 100ms, but it allows for twice as much:

```go
	config.Default.Latency = 50 * time.Millisecond
	err = pingPong("100", 100*time.Millisecond, 200*time.Millisecond)
```

nonsense:
> > @raulk We have 200ms latency, because you have a RTT of 100ms, so 2*100ms.
> 
> The patch does the following -- setting a latency of 50ms, where the perfect RTT would be 100ms, but it allows for twice as much:
> 
> ```go
> 	config.Default.Latency = 50 * time.Millisecond
> 	err = pingPong("100", 100*time.Millisecond, 200*time.Millisecond)
> ```

You're right - I was referring to the previous change for the RTT of 200ms. We have two RTTs in this test.

In any case, my question still stands - what is a good value between 28ms and 100ms, having in mind that this is a local machine, and is supposed to work on a local environment - we don't really have any network or continent data transfer.

plans/network/main.go
```diff
@@ -182,17 +182,17 @@ func pingpong(runenv *runtime.RunEnv) error {
 
 		return nil
 	}
-	err = pingPong("200", 200*time.Millisecond, 215*time.Millisecond)
```
nonsense, line 185: RTT 1 test.

plans/network/main.go
```diff
@@ -182,17 +182,17 @@ func pingpong(runenv *runtime.RunEnv) error {
...
 	}
 
-	config.Default.Latency = 10 * time.Millisecond
+	config.Default.Latency = 50 * time.Millisecond
 	config.CallbackState = "latency-reduced"
 	netclient.MustConfigureNetwork(ctx, config)
 
 	runenv.RecordMessage("ping pong")
-	err = pingPong("10", 20*time.Millisecond, 35*time.Millisecond)
+	err = pingPong("100", 100*time.Millisecond, 200*time.Millisecond)
```
nonsense, line 195: RTT 2 test.

raulk:
@nonsense the first thing I would try to do is remove the log statements inside the `pingpong` method. I suspect those are hitting the disk, and for some reason, @hacdias's disk could've been really slow.

nonsense:
@raulk I'd rather not remove the log statements, because this is a debug test, and these statements were added in order to find problems with the environment:

`before netclient.MustWaitNetworkInitialized` - we would block after this statement, if we have a problem with sync service or creating the default data network.

`before netclient.MustConfigureNetwork` - we would block after this statement, if we can't change the network configuration, etc.

---

1. We shouldn't be removing log statements hitting the disk or printing out to stdout, just because we want this test to be "fast".
2. Timeouts in the test should be the `max` values we observe in order to avoid a flaky test plan, which this one currently is. `15ms` is very aggressive considering that we want to run this test plan as an integration test.

hacdias:
For the record, I just run. it about 10 times now and none of them failed. Maybe, like @raulk said, it was related to my disk's speed at the moment.

nonsense:
@hacdias thanks for the report! I guess we have to run this on CircleCI as part of the integration testing we are doing this sprint and see how it performs there.

raulk:
Let's close this given the discussion above, and revisit if it reappears.

PR #993, opened 2020-05-12
k8s runner integration test

coryschwartz:
testing using kind for integration tests in CI


.circleci/config.yml
```diff
@@ -23,6 +23,30 @@ commands:
...
+            fi
+      - run:
+          name: "install kind"
+          command: |
+            if [ ! -f ~/bin/kind ]; then
+            curl -L https://github.com/kubernetes-sigs/kind/releases/download/v0.8.1/kind-linux-amd64 -o ~/bin/kind
+            chmod +x ~/bin/kind
+            fi
+      - save_cache:
+          key: 'bin-cache-{{ .Branch }}'
```
coryschwartz, line 47: This would get the latest kubectl, downloading it once per branch.. Maybe just get rid of the `{{ .Branch }}` and just keep everything static.

But the general idea is to add kubectl and kind binaries.

.circleci/config.yml
```diff
@@ -90,3 +117,44 @@ jobs:
...
+    steps:
+      - setup-install-bins
+      - run:
+          name: "build testground"
+          command: make install
+      - run:
+          name: "configure cluster"
+          command: |
+            kind create cluster --wait 90s
+            kubectl apply -f .circleci/pv.yaml
```
coryschwartz, line 133: Do we want these somewhere else?
These are used to create the fake EFS volume in our fake kubernetes cluster.

env-example.toml
```diff
@@ -24,6 +24,9 @@ access_token = "docker hub access token"
 [runners."cluster:k8s"]
 pod_resource_cpu      = "100m"
 pod_resource_memory   = "100Mi"
+sysctls = [
```
coryschwartz, line 27: Merged from https://github.com/testground/testground/pull/1008 which @nonsense made for this purpose.

This change makes the default sysctls empty unless other specified in env.toml.

pkg/runner/cluster_k8s.go
```diff
@@ -141,8 +142,15 @@ func (c *ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow *rpc
 
 	cfg := *input.RunnerConfig.(*ClusterK8sRunnerConfig)
 
-	defaultCPU := resource.MustParse(cfg.PodResourceCPU)
-	defaultMemory := resource.MustParse(cfg.PodResourceMemory)
+	var defaultCPU, defaultMemory resource.Quantity
```
coryschwartz, line 145: There are no defautls for defaultCPU and defaultMemory and without having them set in .env.toml causes an error, so I am doing this to create defaults.

the local:docker runner handles this differently. It has a default config struct and uses `mergo` to merge the defaults with supplied config. maybe this should be made the same.

.circleci/config.yml
```diff
@@ -90,3 +117,44 @@ jobs:
...
+          command: |
+            testground daemon &
+            testground plan import --from plans/placebo
+            testground build single --builder docker:go --plan placebo | tee build.out
+            export ARTIFACT=$(awk -F\" '/generated build artifact/ {print $8}' build.out)
+            docker tag $ARTIFACT testplan:placebo
+            kind load docker-image testplan:placebo
+            testground run single --runner cluster:k8s --builder docker:go --use-build testplan:placebo --instances 1 --plan placebo --testcase ok | tee run.out
+      - run:
+          name: "verify output"
```
coryschwartz, line 151: I had initially thought I would do --collect and verify the output captured by --collect is the same as what we see here. This does not work because I am only building a single-node k8s cluster and I have to label it as either `testground.nodetype=infra` or `testground.nodetype=plan`. If we want to test. So the collect-outputs pod waits forever a node.

I would point out that kind *does* support "multi-node" containers.
If we *do* want to test the "collection" pod behavior, I think I would just do something like this rather than a multi-node cluster.
1. set the label to `testground.nodetype=plan`
2. execute the plan, just as we already are
3. change the label to `testground.nodetype=infra`
4. test the collection path with `testground collect`

But for now I'm just checking the output by logging into the node and looking at the files.

.circleci/config.yml
nonsense: For now they are fine here I think. Ideally we want the test setup to be at least surprising as possible with very little indirection.

.circleci/config.yml
nonsense: Instead of doing `nodetype`, can't we have labels for multiple roles?

Something like:
`testground.role.infra = true`
`testground.role.plan = true`

This way, we can tag the same node with multiple roles, and avoid having to run multiple nodes for various roles.

pkg/runner/cluster_k8s.go
nonsense: Yes, this is an oversight on my part. We should have defaults, and ideally the same way as other runners.

nonsense: The only reason I left this like that, is because in general when you run `cluster:k8s`, so far I assumed it'd be AWS (as we don't have another setup until now), and you also had to specify `aws region` - therefore the `.env.toml` is really required when using `cluster:k8s`.

.circleci/config.yml
```diff
@@ -90,3 +117,44 @@ jobs:
...
+      # The plan is renamed as `testplan:placebo` because kind will check dockerhub if the tag is `latest`.
+      - run:
+          name: "run a plan"
+          command: |
+            testground daemon &
+            testground plan import --from plans/placebo
+            testground build single --builder docker:go --plan placebo | tee build.out
+            export ARTIFACT=$(awk -F\" '/generated build artifact/ {print $8}' build.out)
+            docker tag $ARTIFACT testplan:placebo
+            kind load docker-image testplan:placebo
```
nonsense, line 148: We should probably extract this into a `sh`, so that we can reuse for other test plans.

.circleci/config.yml
nonsense, line 151: I suggest we join the `run a plan` and `verify output` into one CircleCI `run`, and extract in `sh` that we can modify and add more tests to, rather than have to edit the circleci.yml to add more testplan runs.

I imagine we want to run testplans serially on the `testground daemon` for now.

WDYT?

PR #1008, opened 2020-05-13
extract sysctls into .env.toml ; extract collect-outputs requests in .env.toml

nonsense:
This PR is:
1. extracting `sysctls` into an `.env.toml` file - we want that, so that we can run integration tests on Kubernetes, in environments that don't allow us to set `sysctls`.
2. extracting `collect-outputs` pod requests into `.env.toml` - in the real-world we want to run this with a lot of RAM and CPU, but for test purposes we need fewer resources.

### DIFF
--- a/env-example.toml
+++ b/env-example.toml
@@ -22,8 +22,13 @@ access_token = "docker hub access token"
 #   4. Runner defaults (applied by the runner).
 #
 [runners."cluster:k8s"]
-pod_resource_cpu      = "100m"
-pod_resource_memory   = "100Mi"
+testplan_pod_cpu            = "100m"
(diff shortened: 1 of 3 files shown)


coryschwartz:
This can be closed whenever https://github.com/testground/testground/pull/993 is merged since I included these changes there.

nonsense:
Closing here, will be merged as part of https://github.com/testground/testground/pull/993

nonsense:
@coryschwartz looks like we didn't extract everything from this PR into the intergration tests PR - currently the `.env.toml` is wrong, `collect outputs pod` is still using hard-coded values on `master`, and the names of the params are not updated.

Issue #1005, opened 2020-05-12
 Unit tests for pkg/docker package. 


Robmat05:
Please see Epic description for details on specific tests

coryschwartz:
This package has the interfaces testground uses for interacting with docker. Should look at upstream docker tests and make sure our "ensure" APIs work as expected.

PR #1009, opened 2020-05-14
`docker` package tests

coryschwartz:
Implements a few (but not all) tests for pkg/docker

More will need to be added later, particularly for the docker manager.

partial fix: https://github.com/testground/testground/issues/1005


nonsense:
I see this is a draft PR, but I am not convinced we need these tests at all. I think the carry cost is probably higher than the value they bring. I'd rather we have a few integration tests with proper test plans that hit these APIs, than these unit tests.

coryschwartz:
@nonsense I get your point, for sure. All the integration tests will hit the docker code. I mostly want to check for honesty -- false negatives and false positives. My plan was to write these which interface directly with the docker daemon and the Manager since this is used by the sidecar,

pkg/rpctest/rpctest.go
```diff
@@ -0,0 +1,17 @@
+package rpctest
```
coryschwartz, line 1: I added this so I could create an OutputWriter, which is an argument in many functions. My intention is not that any binary other than tests will ever import this package.

I'm open to ideas, for doing this better.

coryschwartz: I've now removed this, and rpctest was added in https://github.com/testground/testground/pull/1015 instead.

pkg/docker/docker_test.go
```diff
@@ -0,0 +1,212 @@
...
+
+	"github.com/testground/testground/pkg/docker"
+	"github.com/testground/testground/pkg/rpc/rpctest"
+)
+
+var (
+	cli *client.Client
+)
+
+func init() {
```
nonsense, line 28: This is a candidate for `TestMain` if we ever need some kind of teardown - https://golang.org/pkg/testing/#hdr-Main

pkg/docker/docker_test.go
```diff
@@ -0,0 +1,212 @@
...
+	cli, err = client.NewClientWithOpts(client.FromEnv)
+	if err != nil {
+		panic(err)
+	}
+
+	cli.NegotiateAPIVersion(context.Background())
+	rand.Seed(time.Now().UnixNano())
+}
+
+func errfail(t *testing.T, err error) {
```
nonsense, line 39: Nitpick, but I don't think we need this.

Just having:
```
if err != nil {
  t.Fatal(err)
}
```

is fine.

---

Also without a call to `t.Helper()` I don't think this will be reporting correct line numbers on where the error happened, so it'd be hard to detect what error happened if a test calls it >1 times.

pkg/docker/docker_test.go
```diff
@@ -0,0 +1,212 @@
...
+		return "", err
+	}
+	return body.ID, err
+}
+
+// pull an image from docker hub.
+// create a container with a randomized name
+// Configure the container to be deleted when the test completes.
+// return the container ID.
+func pull_create_delete(ctx context.Context, t *testing.T, imageName string) (containerID string, containerName string) {
```
nonsense, line 107:
```suggestion
func pullCreateDelete(t *testing.T, ctx context.Context, imageName string) (containerID string, containerName string) {
```

pkg/docker/docker_test.go
```diff
@@ -0,0 +1,212 @@
...
+
+	imageName := "hello-world"
+	err := pullImage(ctx, imageName)
+	errfail(t, err)
+
+	_, found, err := docker.FindImage(ctx, ow, cli, imageName)
+	errfail(t, err)
+
+	if !found {
+		t.Fail()
```
nonsense, line 134: It'd be nice to fail with an error message, saying what was expected and what happened.
```suggestion
		t.Fatalf("expected to find a docker image named %d, but didnt", imageName)
```

pkg/docker/docker_test.go
```diff
@@ -0,0 +1,212 @@
...
+	id, name := pull_create_delete(ctx, t, image)
+	cont, err := docker.CheckContainer(ctx, ow, cli, name)
+	errfail(t, err)
+	if cont == nil {
+		t.Log("container not found. nil.")
+		t.Fail()
+	}
+	if cont.ID != id {
+		t.Log("incorrect container found. id does not match that created.")
+		t.Fail()
```
nonsense, line 194:
```suggestion
		t.Fatalf("incorrect container found. id does not match that created. expected %s, got %s", id, cont.ID)
```

pkg/docker/docker_test.go
raulk: Just use `require.NoError`. Look at `TestAllEvents` in the SDK, or at `selector_test.go`.

pkg/docker/docker_test.go
raulk: Yup. I don't think I ever investigated when `init()` vs. `t.Main()` run.

PR #1011, opened 2020-05-14
disable default CNI on kind, so that we support multiple networks

nonsense:


nonsense:
Closing in favour of https://github.com/testground/testground/pull/1013

PR #1013, opened 2020-05-15
sidecar/k8s-containerd runner

nonsense:
An attempt to add support for `containerd`.

TODO:
- [ ] fix DNS when pod is started on host network (i.e. `sidecar`)
- [ ] normalise the CIDR used - `podCIDR` and `servicesCIDR`
- [ ] add scripts to setup `kind` environment locally - then translate to CircleCI. this mostly includes:
- [ ] lower resources for all services
- [ ] no persistence for influxdb
- [ ] ...


nonsense:
Putting this on-hold for now. We might want to investigate further in the future because getting rid of Docker would reduce pod startup latency and resource consumption on nodes in cluster, according to benchmarks I've seen - https://kubernetes.io/blog/2018/05/24/kubernetes-containerd-integration-goes-ga/ and others.

laurentsenta:
@nonsense I added a not in https://github.com/testground/testground/issues/1523 but will close the PR for now since we're not going to complete this.

Issue #1004, opened 2020-05-12
 Unit tests for pkg/rpc package (chunker and teeing OutputWriter). 


Robmat05:
Please see Epic description for details on specific tests

coryschwartz:
I have part of this done in https://github.com/testground/testground/pull/1009 because I wanted to add the RecordedOutputWriter. Since the docker tests in that PR are in question, maybe I'll break out that part of it to a separate PR and take this as well.

coryschwartz:
The PR I just added has the rpctest broken out. See https://github.com/testground/testground/pull/1015  was originally written in https://github.com/testground/testground/pull/1009 but I stripped out the RPC component.

PR #1015, opened 2020-05-19
`rpc` package tests

coryschwartz:
Separated out the rpctest component from https://github.com/testground/testground/pull/1009 since there was question about the usefulness of the docker test and this part is unrelated.

Fix: https://github.com/testground/testground/issues/1004

### DIFF
--- a/pkg/rpc/rpc_test.go
+++ b/pkg/rpc/rpc_test.go
@@ -4,16 +4,51 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	"reflect"
+	"strings"
 	"testing"
 
 	"github.com/testground/testground/pkg/rpc"
 	"github.com/testground/testground/pkg/rpc/rpctest"
 )
 
-type tc struct {
+type testcase struct {
 	f        string
-	ct       rpc.ChunkType
-	expected string
+	ct       []rpc.ChunkType
+	param    interface{}
+	expected interface{}
+}
+
+func allowedType(ch rpc.Chunk, cts []rpc.ChunkType) bool {
+	for _, tp := range cts {
+		if ch.Type == tp {
+			return true
+		}
+	}
+	return false
+}
+
+func verifyExpected(t *testing.T, ch rpc.Chunk, expected interface{}) {
+	switch ch.Type {
+	// base64-encoded strings that look like this when decoded:
+	// May 19 09:51:43.022537	INFO	test	{"req_id": "TestLogging-Infof"}
+	// Check the data after the timestamp matches expected.
+	case rpc.ChunkTypeProgress:
+		t.Log(ch.Type, ch.Payload)
+		if !strings.Contains(ch.Payload.(string), expected.(string)[30:]) {
+			t.Log("ChunkTypeProgress data did not match what was expected.", ch.Payload)
+			t.Fail()
+		}
+
+	// string or []uint8
+	case rpc.ChunkTypeResult, rpc.ChunkTypeBinary:
+		if ch.Payload != expected {
+			t.Log("data did not match what was expected.", ch.Payload)
+			t.Fail()
+		}
+
+	case rpc.ChunkTypeError:
+		t.Log("", ch.Payload)
+	}
 }
 
 func TestLogging(t *testing.T) {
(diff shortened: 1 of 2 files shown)


pkg/rpc/rpc_test.go
```diff
@@ -0,0 +1,129 @@
...
+			t.Error(err)
+			t.FailNow()
+		}
+		var ch rpc.Chunk
+		err = json.Unmarshal(body, &ch)
+		if err != nil {
+			t.Error(err)
+			t.FailNow()
+		}
+		t.Run("expected chunktype: "+test.f, func(t *testing.T) {
```
raulk, line 60: This usage of `t.Run` is weird. `t.Run` is used to run subtests, here you're using it to group asserts on test logic that has already run. Try grouping the entire for block inside `t.Run`?

pkg/rpc/rpc_test.go
```diff
@@ -0,0 +1,129 @@
...
+		t.Run("expected content: "+test.f, func(t *testing.T) {
+			p := ch.Payload.(string)
+			teststr := p
+			expected := test.expected
+			if test.ct == rpc.ChunkTypeProgress {
+				// ChunkTypeProgress contains unpredictable timestamps.
+				// Rather than deserializing, just check the last part of the data.
+				part := len(p) - 50
+				teststr = p[part:]
+				expected = test.expected[part:]
```
raulk, line 76: Can't we just use https://golang.org/pkg/bytes/#Contains?

raulk: or https://golang.org/pkg/strings/#Contains
or https://golang.org/pkg/strings/#HasSuffix

pkg/rpc/rpctest/rpctest.go
```diff
@@ -0,0 +1,17 @@
+package rpctest
```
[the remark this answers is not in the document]
coryschwartz: The reason for the separate package is that OutputWriter is used everywhere and I wanted to be able to import something in other packages. I originally wrote this so I could test docker packages in https://github.com/testground/testground/pull/1009 and just needed some kind of OutputWriter without using the http server to create it. But yeah, if there's a better way, lets do that instead.

pkg/rpc/rpc_test.go
coryschwartz: Sure. I was just doing this for the output formatting.

pkg/rpc/rpc_test.go
```diff
@@ -0,0 +1,129 @@
...
+	// Fatal causes a crash. Just check the ones that send to the client.
+
+	tcs := []tc{
+		{"Info", rpc.ChunkTypeProgress, "TWF5IDE5IDA5OjUxOjQzLjAyMjI3OQkbWzM0bUlORk8bWzBtCXRlc3QJeyJyZXFfaWQiOiAiVGVzdExvZ2dpbmctSW5mbyJ9Cg=="},
+		{"Infof", rpc.ChunkTypeProgress, "TWF5IDE5IDA5OjUxOjQzLjAyMjUzNwkbWzM0bUlORk8bWzBtCXRlc3QJeyJyZXFfaWQiOiAiVGVzdExvZ2dpbmctSW5mb2YifQo="},
+		{"Infow", rpc.ChunkTypeProgress, "TWF5IDE5IDA5OjUxOjQzLjAyMjY1MAkbWzM0bUlORk8bWzBtCXRlc3QJeyJyZXFfaWQiOiAiVGVzdExvZ2dpbmctSW5mb3cifQo="},
+		{"Error", rpc.ChunkTypeProgress, "TWF5IDE5IDA5OjUxOjQzLjAyMjc1MQkbWzMxbUVSUk9SG1swbQl0ZXN0CXsicmVxX2lkIjogIlRlc3RMb2dnaW5nLUVycm9yIn0K"},
+		{"Errorf", rpc.ChunkTypeProgress, "TWF5IDE5IDA5OjUxOjQzLjAyMjg1MwkbWzMxbUVSUk9SG1swbQl0ZXN0CXsicmVxX2lkIjogIlRlc3RMb2dnaW5nLUVycm9yZiJ9Cg=="},
+		{"Errorw", rpc.ChunkTypeProgress, "TWF5IDE5IDA5OjUxOjQzLjAyMjk1NQkbWzMxbUVSUk9SG1swbQl0ZXN0CXsicmVxX2lkIjogIlRlc3RMb2dnaW5nLUVycm9ydyJ9Cg=="},
+		{"WriteResult", rpc.ChunkTypeResult, "test"},
```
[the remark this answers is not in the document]
coryschwartz: I'll take care of this today.

WriteError is a Json stream not just a single document
WriteBinary writes a []uint8

pkg/rpc/rpc_test.go
coryschwartz: Yeah, I might re-think this though to not use just stiring comparisons,let me hack on this today.

pkg/rpc/rpctest/rpctest.go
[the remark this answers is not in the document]
raulk: Yes, but the tests are already in a separate package (`rpc_test`). This PR introduces:

* `pkg/rpc/rpc_test.go` under package `rpc_test`.
* `pkg/rpc/rpctest/rpctest.go` under package `rpctest`.

I think the latter is redundant.

pkg/rpc/rpctest/rpctest.go
```diff
@@ -0,0 +1,17 @@
...
+	"github.com/testground/testground/pkg/rpc"
+)
+
+// NewRecordedOutputWriter returns an OutputWriter where the response is recorded.
+func NewRecordedOutputWriter(req_id string) (rec *httptest.ResponseRecorder, ow *rpc.OutputWriter) {
+	req := httptest.NewRequest("GET", "/", strings.NewReader(""))
+	req.Header.Add("X-Request-ID", req_id)
+	rec = httptest.NewRecorder()
+	ow = rpc.NewOutputWriter(rec, req)
+	return rec, ow
```
raulk, line 16: @coryschwartz ^^ still needs fixing. (golang.org/doc/effective_go.html#mixed-caps)

```suggestion
func NewRecordedOutputWriter(reqID string) (rec *httptest.ResponseRecorder, ow *rpc.OutputWriter) {
	req := httptest.NewRequest("GET", "/", strings.NewReader(""))
	req.Header.Add("X-Request-ID", reqID)
	rec = httptest.NewRecorder()
	ow = rpc.NewOutputWriter(rec, req)
	return rec, ow
```

PR #1017, opened 2020-05-19
fix make test

nonsense:
`make test` doesn't work locally unless you have `placebo` and `example` in the `TESTGROUND_HOME` directory.

Therefore I am adding an `testground plan import` prior to the call to `go test` on every module.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,5 +37,11 @@ docker-sidecar:
 docker-testground:
 	docker build -t iptestground/testground:edge -f Dockerfile.testground .
 
-test:
+clean:
+	rm -rfi $${TESTGROUND_HOME}/plans/placebo
+	rm -rfi $${TESTGROUND_HOME}/plans/example
+
+test: install
+	testground plan import --from ./plans/placebo || true
+	testground plan import --from ./plans/example || true
 	$(call eachmod,go test -p 1 -v $(GOTFLAGS) ./...)
--- a/Makefile
+++ b/Makefile
@@ -37,10 +37,6 @@ docker-sidecar:
 docker-testground:
 	docker build -t iptestground/testground:edge -f Dockerfile.testground .
 
-clean:
-	rm -rfi $${TESTGROUND_HOME}/plans/placebo
-	rm -rfi $${TESTGROUND_HOME}/plans/example
-
 test: install
 	testground plan import --from ./plans/placebo || true
 	testground plan import --from ./plans/example || true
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -37,5 +37,11 @@ docker-sidecar:
 docker-testground:
 	docker build -t iptestground/testground:edge -f Dockerfile.testground .
 
+clean:
```
nonsense, line 40: `make test` might also fail with outdated `plans` in TESTGROUND_HOME... if this happens a user could just do `make clean` and then `make test`, which should always work.

Makefile
```diff
@@ -37,5 +37,11 @@ docker-sidecar:
 docker-testground:
 	docker build -t iptestground/testground:edge -f Dockerfile.testground .
 
-test:
+clean:
+	rm -rfi $${TESTGROUND_HOME}/plans/placebo
+	rm -rfi $${TESTGROUND_HOME}/plans/example
```
raulk, line 42: I really dislike deleting anything outside the working copy. Let's not do this. The `|| true` below should be enough.

Makefile
```diff
@@ -37,5 +37,11 @@ docker-sidecar:
...
+
+test: install
+	testground plan import --from ./plans/placebo || true
+	testground plan import --from ./plans/example || true
```
raulk, line 46: This looks a lot like the CircleCI tasks. Let's make CircleCI call the makefile: https://github.com/testground/testground/blob/master/.circleci/config.yml#L113-L116

Makefile
nonsense: -i is for confirming. `placebo` and `example` are really Testground plans.

nonsense: `|| true` is not enough if you have an outdated plan in your TESTGROUND_HOME.

Makefile
nonsense: Yeah, I am planning on doing something about that, just not sure how exactly right now.

Makefile
nonsense: You're right that, `-i` is probably not good enough of an excuse to acll `rm` outside of the working dir... maybe TESTGROUND_HOME is not set and a user tries to delete /plans/placebo...

I don't like that our tests depend on TESTGROUND_HOME, maybe it is worth actually fixing this.

nonsense:
Rerequesting review, after removing the interactive rm part. I think this is still better than `master` because new users would be able to run `make test` and see our tests passing. The TESTGROUND_HOME directory plans could still diverge from the codebase, but we can solve that later.

PR #1023, opened 2020-05-20
remove BuildID from docker based builders

nonsense:
TODO:
- [ ] investigate further if there is a docker API that returns the BuildID - I couldn't find one immediately.
- [x] tag image with `testplan-name:<build id>` prior to pushing (when working with local:docker)
- [ ] decide whether to keep `BuildID` because of `exec:go` builder.

### DIFF
--- a/pkg/build/docker_generic.go
+++ b/pkg/build/docker_generic.go
@@ -38,7 +38,6 @@ func (b *DockerGenericBuilder) Build(ctx context.Context, in *api.BuildInput, ow
 	cliopts := []client.Opt{client.FromEnv, client.WithAPIVersionNegotiation()}
 
 	var (
-		id       = in.BuildID
 		basesrc  = in.BaseSrcPath
 		cli, err = client.NewClientWithOpts(cliopts...)
 	)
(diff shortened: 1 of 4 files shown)


pkg/build/docker_generic.go
```diff
@@ -38,20 +38,17 @@ func (b *DockerGenericBuilder) Build(ctx context.Context, in *api.BuildInput, ow
...
 	if err != nil {
 		return nil, err
 	}
 
-	ow = ow.With("build_id", id)
 	ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
 	defer cancel()
 
 	opts := types.ImageBuildOptions{
-		Tags:        []string{id, in.BuildID},
```
nonsense, line 54: Previously this was creating the default tag, but it creates `latest` tag, and we don't really want that.

pkg/build/docker_generic.go
```diff
@@ -64,15 +61,23 @@ func (b *DockerGenericBuilder) Build(ctx context.Context, in *api.BuildInput, ow
...
 
 	ow.Infow("build completed", "took", time.Since(buildStart).Truncate(time.Second))
 
 	out := &api.BuildOutput{
-		ArtifactPath: in.BuildID,
+		ArtifactPath: imageID,
+	}
+
+	// Default tag for the testplan image
+	defaultImageTag := fmt.Sprintf("%s:%s", in.TestPlan, imageID)
```
nonsense, line 76: If I build the `benchmarks` testplan, this would tag the image built in the previous step, such as:

```
REPOSITORY                                                                           TAG                                      IMAGE ID            CREATED              SIZE
<none>                                                                               <none>                                   aeaad624797f        About a minute ago   21.2MB
```

to

```
REPOSITORY                                                                           TAG                                      IMAGE ID            CREATED              SIZE
benchmarks                                                                           aeaad624797f                             aeaad624797f        About a minute ago   21.2MB
```

nonsense:
Closing in favour of https://github.com/testground/testground/pull/1024

PR #1024, opened 2020-05-20
deterministic docker image name and tag

nonsense:
Currently we are generating a unique build ID, and naming our test plan images with it, which results in unnecessary queries to the Docker Registry when Docker need to create containers from a given image.

This PR is updating the name scheme for our testplan images, so that we reference them with:
`<testplan-name>:<docker image id>`

---

We still need to keep the unique `BuildID` in order to determine the `docker image id` for a given Docker build (i.e. we call `docker image list` and filter based on the unique build id, in order to fetch the `docker image id`.

### DIFF
--- a/pkg/docker/image.go
+++ b/pkg/docker/image.go
@@ -1,15 +1,12 @@
 package docker
 
 import (
-	"bufio"
-	"bytes"
 	"context"
-	"errors"
 	"fmt"
-	"io"
 	"strings"
 
 	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/client"
 	"github.com/docker/docker/pkg/archive"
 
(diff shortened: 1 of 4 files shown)


nonsense:
@coryschwartz @raulk see closed PR https://github.com/testground/testground/pull/1023 for an alternative way to get the docker image id. I'd rather we keep the unique build ID and filter by it, rather than reading the output of `docker build` and grep-ing for `Successfully built` line. WDYT?

pkg/docker/image.go
```diff
@@ -101,3 +103,23 @@ func FindImage(ctx context.Context, ow *rpc.OutputWriter, client *client.Client,
 	}
 	return nil, false, nil
 }
+
+func GetImageID(ctx context.Context, cli *client.Client, defaultTag string) (string, error) {
```
raulk, line 107: We should probably unify this with `FindImage` above.

pkg/docker/image.go
```diff
@@ -101,3 +103,23 @@ func FindImage(ctx context.Context, ow *rpc.OutputWriter, client *client.Client,
...
+	if err != nil {
+		return "", fmt.Errorf("docker image list failed: %w", err)
+	}
+
+	if len(images) != 1 {
+		return "", fmt.Errorf("unexpected number of images returned by docker image list, expected 1, got: %d", len(images))
+	}
+
+	// get 3cde7451eb28 from sha256:3cde7451eb28a3199f2c7d4e8e02a98f2e96b9a34dd4a9bc7eeaa5a192a1536f
+	return images[0].ID[7 : 7+12], nil
```
raulk, line 124: There's an implicit precondition here that we're not testing against. We should do a `HasPrefix` assertion and panic if false.

raulk: I'll push this change and merge.

pkg/docker/image.go
raulk: Should probably also add a test.

Issue #975, opened 2020-05-07
Consider how to introduce network partitioning functionality in Testground


nonsense:
Filecoin will need soon network partitioning functionality. We should consider how to implement this. It could be done on the:
- application-side (libp2p swarm)
- network-side (network rules, as envisioned in the sidecar)
- through weave npc maybe

The main problem is: how do we write a testplan that introduces a network partition at a given point, so that nodes on either side of the partition can still talk to each other. We should consider how to do that when connections are already established, and the partition is temporary, and the problem goes away after a certain period of time.

raulk:
The data model for this is already in place in the sdk `[network` package](https://pkg.go.dev/github.com/testground/sdk-go@v0.2.1/network?tab=doc#LinkShape). Test instances can set specific filter actions (DENY, ACCEPT) for specific subnets.

However, the sidecar is currently not doing anything with those settings. I guess we need to:

- [ ] Provider helper methods to calculate IP addresses we want assigned to ourselves, within a particular subnet of the test run's subnet.
  * Instances in the A side of the partition calculate IP addresses in a subnet `/24` out of the test run's `/16`.
  * Instances in the B side get a different `/24` subnet.
  * We use the sync service to decide which gets which.
- [ ] Instances request network configuration changes that DENIES the other subnet.
  * Sidecar needs to support this, I guess via `iptables`.

PR #1027, opened 2020-05-21
sidecar use Linkshape FilterAction

coryschwartz:
Partial fix for  https://github.com/testground/testground/issues/975

Implement LinkRule FilterAction in the sidecar.


coryschwartz:
@nonsense this works by just null-routing certain routes or hosts.

This PR works, but we should work work toward adding additional qdisc instead and filter by the network. I think we should do this for the upcoming release and create a ticket to improve it and resolve all the TODOs that are strewn throughout this code.

plans/example/drop.go
```diff
@@ -0,0 +1,60 @@
...
+		},
+		CallbackState: "network-configured",
+	}
+
+	runenv.RecordMessage("before netclient.MustConfigureNetwork")
+	netclient.MustConfigureNetwork(ctx, config)
+	runenv.RecordMessage("after netclient.MustConfigureNetwork")
+
+	// This will give you time to log in and see what's going on.
+	time.Sleep(10 * time.Minute)
```
raulk, line 57: Can we build a test that:

1. starts 4 instances.
2. each requests an IP in a different subnet in the data plane.
3. each starts a TCP listener on that IP.
4. they do some blocking/accepting actions and verify that traffic did/did not get across.
5. records successes or failures based on what happened.

pkg/sidecar/link.go
```diff
@@ -181,6 +182,29 @@ func (l *NetlinkLink) Shape(shape sync.LinkShape) error {
...
+// TODO(cory) actually process the shape per network.
+// For now, this simply adds a route based on the Filter
+func (l *NetlinkLink) AddRules(rules []network.LinkRule) error {
+	for _, rule := range rules {
+		r := netlink.Route{
+			Dst: &rule.Subnet,
+		}
+		switch rule.Filter {
+		case network.Accept:
+			continue
```
raulk, line 194: Will this work for the case where we're going _from_ blocking _to_ accepting a subnet?

coryschwartz: that's a good point, you should be able to accept later once you've blocked it.

plans/splitbrain/manifest.toml
```diff
@@ -0,0 +1,33 @@
+name = "network"
+
+[defaults]
+builder = "exec:go"
+runner = "local:exec"
```
raulk, line 5:
```suggestion
```

This section has been deprecated.

plans/splitbrain/manifest.toml
```diff
@@ -0,0 +1,33 @@
...
+enabled = true
+
+[builders."exec:go"]
+enabled = true
+
+[runners."local:docker"]
+enabled = true
+
+[runners."local:exec"]
+enabled = true
```
raulk, line 17: Should probably disable the `local:exec` runner since this test requires the sidecar.

plans/splitbrain/main.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	runtime.InvokeMap(testcases)
+}
+
+func setup(ctx context.Context, runenv *runtime.RunEnv) (client *sync.Client, nwclient *network.Client) {
+	client = sync.MustBoundClient(ctx, runenv)
+	defer client.Close()
+
+	if !runenv.TestSidecar {
+		return nil, nil
```
raulk, line 46: Return an error?

plans/splitbrain/main.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	nwclient = network.NewClient(client, runenv)
```
raulk, line 49: Convention is to call it `netclient`.

```suggestion
	netclient = network.NewClient(client, runenv)
```

plans/splitbrain/main.go
```diff
@@ -0,0 +1,152 @@
...
+			return err
+		}
+
+		ip, nw, err := net.ParseCIDR(addrs[0].String())
+		if err != nil {
+			return err
+		}
+
+		// Start an HTTP server
+		runenv.RecordMessage("I have address %s", ip)
```
raulk, line 78: Can we use [`MustGetDataNetworkIP()`](https://pkg.go.dev/github.com/testground/sdk-go@v0.2.1/network?tab=doc#Client.MustGetDataNetworkIP) instead?

plans/splitbrain/main.go
```diff
@@ -0,0 +1,152 @@
...
+		http.HandleFunc("/", func(w http.ResponseWriter, req *http.Request) { fmt.Fprintln(w, "hello.") })
+		go http.ListenAndServe(":8765", nil)
+
+		seq := client.MustSignalEntry(ctx, "pickaside")
+
+		// Generate a player and publish the address to the sync service.
+		me := node{int(seq) % 3, nw}
+		nodeTopic := sync.NewTopic("nodes", node{})
+		nodeCh := make(chan *node, 0)
+		client.MustPublishSubscribe(ctx, nodeTopic, &me, nodeCh)
```
raulk, line 89: Need to capture the subscription here and close it when we have finished consuming on line 97.

plans/splitbrain/main.go
```diff
@@ -0,0 +1,152 @@
...
+		for _, p := range nodes {
+			resp, err := http.Get(p.IPNet.IP.String() + ":8765")
+			if err != nil {
+				errs++
+			}
+			if resp.StatusCode == 200 {
+				status200++
+			}
+			total++
+		}
```
raulk, line 144: Can we succeed or fail the test depending on:
1. the test case (drop/reject/accept).
2. what side of the partition we're in

Using `RecordSuccess` or `RecordFailure`?

plans/example/drop.go
```diff
@@ -0,0 +1,60 @@
...
+func ExampleDropNetwork(runenv *runtime.RunEnv) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 300*time.Second)
+	defer cancel()
+
+	runenv.RecordMessage("before sync.MustBoundClient")
+	client := sync.MustBoundClient(ctx, runenv)
+	defer client.Close()
+
+	if !runenv.TestSidecar {
+		return nil
```
nonsense, line 22: This should return an error, the plan doesn't make sense without sidecar.

plans/splitbrain/main.go
```diff
@@ -0,0 +1,172 @@
...
+
+		// Each node starts an HTTP server to test for connectivity
+		runenv.RecordMessage("Starting http server")
+		http.HandleFunc("/", func(w http.ResponseWriter, req *http.Request) {
+			runenv.RecordMessage("received http request from %s", req.RemoteAddr)
+			fmt.Fprintln(w, "hello.")
+		})
+		go func() {
+			err := http.ListenAndServe(":8765", nil)
+			runenv.RecordFailure(err)
```
nonsense, line 75: We will always reach this point, so I am not sure why we always want to `RecordFailure` ?

plans/example/drop.go
```diff
@@ -0,0 +1,60 @@
+package main
```
nonsense, line 1: We should either come up with a simple plan (`splitbrain` seems to be it), or just include this snippet in the docs website. I don't see a reason to have it in `plans/example` as it is not really a plan and doesn't do anything, but rather just showcases functionality that you could use.

plans/splitbrain/main.go
```diff
@@ -0,0 +1,172 @@
...
+		var errs int
+		var status200 int
+		var total int
+
+		// Try to reach out to each node and see what happens.
+		httpclient := http.Client{
+			Timeout: time.Minute,
+		}
+
+		// When running the "accept" testcase, there should be no failures.
```
nonsense, line 140: I think this is a candidate for a comment at the top of the file or for the `routeFilter` function. It is not immediately obvious what this testplan is about, what it is testing and what the rules for region A, B and C are.

nonsense:
Overall LGTM, added a few comments that I think it'd be nice to address, and then we can iterate further on this if necessary.

PR #1028, opened 2020-05-22
move ECR push to cluster:k8s runner.

raulk:
Fixes https://github.com/testground/testground/issues/644.

_Untested_


pkg/build/docker_generic.go
```diff
@@ -23,9 +23,7 @@ type DockerGenericBuilder struct {
 }
 
 type DockerGenericBuilderConfig struct {
-	BuildArgs    map[string]*string `toml:"build_args"` // ok if nil
-	PushRegistry bool               `toml:"push_registry"`
-	RegistryType string             `toml:"registry_type"`
+	BuildArgs map[string]*string `toml:"build_args"` // ok if nil
```
coryschwartz, line 26: We might one day want to support other registries. I never much liked the "registry_type" model. Since they are all docker registries, you just need credentials and the regular docker client.

You can log into ecs with the regular docker client, for example. We could potentially support any docker endpoint by just using the regular docker client and providing a model to pass credentials

pkg/runner/registry.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/docker/docker/client"
+)
+
+// pushToAWSRegistry takes a RunInput and pushes all images to the AWS ECR registry.
+// It then replaces the local artifact paths by the remote artifact paths.
+func pushToAWSRegistry(ctx context.Context, ow *rpc.OutputWriter, client *client.Client, in *api.RunInput) error {
+	ow.Infow("acquiring ECR authentication token")
+
+	// Get a Docker registry authentication token from AWS ECR.
+	auth, err := aws.ECR.GetAuthToken(in.EnvConfig.AWS)
```
coryschwartz, line 22: roughly equivilent to the aws commandline `aws ecr get-login` this gives us a temporary (12h) token to use with the docker client. We *could* split this up so one method of getting credentials is by aws token and another way is by commandline, etc. 

Although if we don't want to bother with other registries at this time, this is fine.

raulk: I'd rather not fork a command-line process and deal with all that complexity, when we have an API that does the same as the CLI itself.

What registry is `kind` configured to use? Can we just add a registry type for `kind`? Is it using the local Docker daemon reigstry?

raulk:
@coryschwartz I need some more information on where `kind` expects to find the image. Does it look in the local daemon image registry?

pkg/runner/cluster_k8s.go
```diff
@@ -142,8 +146,22 @@ func (c *ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow *rpc
 
 	cfg := *input.RunnerConfig.(*ClusterK8sRunnerConfig)
 
+	// Right now we only support provider = "aws"; no provider defaults to "aws".
```
nonsense, line 149: We also support DockerHub, and we should get this to work on `kind` as well (i.e. cluster_k8s working without having to push an image, for example if you already specify an `artifact`)

pkg/build/docker_generic.go
nonsense: We actually support DockerHub apart from AWS ECR.

nonsense:
Closing in favour of https://github.com/testground/testground/pull/1032

PR #1030, opened 2020-05-25
Implement nullop builder

wldhx:
`exec:nullop`, a builder that returns a pre-built binary.

I'm writing a Haskell SDK for testground, and it's convenient for me to just drop a binary to `$TESTGROUND_HOME/data/work/exec-nullop--$TESTPLAN` rather than either spinning docker up or writing a whole Haskell builder.

### DIFF
--- a/pkg/build/nullop.go
+++ b/pkg/build/nullop.go
@@ -1,0 +1,35 @@
+package build
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"reflect"
+
+	"github.com/testground/testground/pkg/api"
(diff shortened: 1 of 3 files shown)


nonsense:
@wldhx sounds like you don't need a builder at all at this stage. You can specify the artifact for your run with `--use-build` flag, or with the `artifact` field in your `composition` file.

There is a bug currently though that `--builder` is required, so you would still have to specify a `compatible` builder, but you could just use `exec:go` for now.

---

Let me know if that makes sense, and if it works for you.

wldhx:
Works for me indeed. Cheers!

PR #1031, opened 2020-05-25
integration tests for sidecar on local:docker

nonsense:
This PR is adding a run for the `network/ping-pong` testplan, which exercises the `sidecar` and confirms we get two outputs with `outcome: ok` from the run.


nonsense:
@coryschwartz merging this, because base is `integ-tests-k8s-collect-outputs` and it is fairly ortogonal (i.e. separate file with another test.)

PR #1032, opened 2020-05-25
move docker images push to cluster:k8s runner

nonsense:
This PR is taking over https://github.com/testground/testground/pull/1028 since Raul is not around this week.

I am removing the default expectation that cluster:k8s is working with AWS, and assuming that `provider` is configured properly by the Testground operator:
- provider == "" => no provider, hence no image pushing (useful for testing with `kind` or when we have another system responsible for distribution of images)
- provder == "aws" => push image to AWS ECR
- provider == "dockerhub" => push image to DockerHub

TODO:
- [x] test that `aws` still works (locally and on kops cluster)
- [ ] test that `dockerhub` still works (locally and on kops cluster)
- [x] test that test plans can override the `provider` (i.e. if a `composition` file has specified an artifact, we wouldn't want `cluster:k8s` runner to try to push any images)

---

Note that the `artifact` field now has 2 functions - it contains the `ImageID` once a builder produces a build, but then gets replaced with the `uri + ImageID` once we push an image somewhere because of the use of `provider`.

This means that folks could experiencee bad UX if they use `provider==""` and `-w` to produce a build, but I guess that is fine and can be considered misconfiguration of Testground (i.e. you are rewriting your composition files with produced artifacts, but you don't want to push the artifacts anywhere).


Robmat05:
like #1028 this PR fixes #644

PR #1035, opened 2020-05-26
LRU cache for pushed docker images

nonsense:
This PR is adding in-memory LRU cache for pushed Docker images to remote repositories. I think it is worth having this, because it removes >10sec. per image from test plan runs that are trying to submit images that have already been pushed to a given remote.


nonsense:
@coryschwartz somehow these changes got reverted, probably due to the fact we merged 5-6 PRs around the same time.

Issue #1025, opened 2020-05-21
sidecar errors for a test run should be shown in test output


willscott:
#### Version information:
b1773a47ede14f027c74b253a1a01caec0a6256d
x86_64 / debian

#### Description:
when debugging / setting up a new test, a logic error meant multiple containers were requesting the same configured IP address from the sidecar. The sidecar in it's docker logs reports an error "address already in use" in failing to perform this action, but the observed behavior when running the test is that the context for the `configureNetwork` command times out, and no error is provided besides that the command failed.

It would be great for the underlying sidecar error to be exposed in the diagnostic errors for the test to help with investigation / debugging

nonsense:
I've added a PR that addresses this for `local:docker` runner. We can't do the same for `cluster:k8s`, because it is not a good idea to tail logs from 100 of containers - we need proper log aggregation tooling across the whole stack for Kubernetes - ELK, or something like that.

PR #1036, opened 2020-05-26
append sidecar logs to testground output

nonsense:
Partially fixes: https://github.com/testground/testground/issues/1025


nonsense:
> This does make the output divergent between local and cluster runners, but I think it's worthwhile to see this easier anyway. As discussed elsewhere, we'll use some log aggregation on cluster runners.

Output is already divergent on `cluster:k8s` and `local:docker` - for larger test run we don't show any output when running in Kubernetes.

I think it is an added value to see the logs from the sidecar when you are running your test, and this is also what Will found out.

I will merge this for now, let's see if Raul has different opinions when he is back.

Issue #1006, opened 2020-05-13
influxdb network is unreachable (local:docker runner)


nonsense:
If `sidecar` container is running and `influxdb` container is not, the healthcheck would start influxdb, but the routes would not work, because currently sidecar resolves dependencies on startup and not on every run.

```
May 13 11:46:35.070731  INFO    32.6513s    MESSAGE << c5728c35817f >> failed to send batch to InfluxDB; attempt 4; err: Post http://testground-influxdb:8086/write?consistency=&db=testground&precision=ns&rp=: dial tcp 192.18.0.5:8086: connect: network is unreachable      {"req_id": "d67a7590"}
```

raulk:
We need to strike a balance between (1) resolving each service every time we manage a container (in a 10k run that means 20k or 30k DNS requests in a very short amount of time, something that k8s and docker nameservers don't appreciate), (2) and resolving only once at the start of the sidecar.

Potential fix could be to cache resolved IPs, and for every container we instrument, ping the IPs to validate them, re-resolving on failure.

Could probably use https://github.com/rs/dnscache and/or the singleflight package: https://pkg.go.dev/golang.org/x/sync/singleflight?tab=doc to help deduplicate a failure storm.

nonsense:
> (1) resolving each service every time we manage a container (in a 10k run that means 20k or 30k DNS requests in a very short amount of time, something that k8s and docker nameservers don't appreciate), (2) and resolving only once at the start of the sidecar.

Actually I think (1) is not going to work, because if a required/optional service `moves` in the middle of a run we have a problem that is hard to address - we have already resolved the previous IP, and we have blacklisted all other IPs, but the service IP.

Hence if a service (let's say InfluxDB) moves in the middle of us allowing/blocking network paths for a given test run, we have to start from scratch for all containers in the test run, which makes the whole thing rather complicated.

---

I think for the short/mid-term an OK solution would be to just resolve all services once at the beginning of any test run - this will work fine even if we have concurrent test runs, which we are not planning right now.

Robmat05:
closed by #1037

PR #1037, opened 2020-05-26
resolve services on each test run

nonsense:
Addresses: https://github.com/testground/testground/issues/1006

---

This PR is adding an LRU cache to both K8s and Docker reactors to keep track of the `run-ids` that the `sidecar` has seen. The first time the `sidecar` sees a new `run-id`, it tries to resolve the necessary auxiliary services (influxdb, redis, etc.) so that we know which routes to allow when managing the networks for the test run containers.

---

TODO:
- [x] confirm that #1006 works - if influxdb is started after the sidecar, we get measurements in influxdb

FOLLOW UP:
- [x] confirm that the Redis client continue to work if Redis moves to another IP. Redis client ideally should handle re-connects.
- [ ] add a retry mechanism in case our DNS queries fail when seeing a new test run id.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,6 +22,7 @@ require (
 	github.com/google/uuid v1.1.1
 	github.com/gorilla/mux v1.7.3
 	github.com/hashicorp/go-multierror v1.1.0
+	github.com/hashicorp/golang-lru v0.5.1
 	github.com/imdario/mergo v0.3.8
 	github.com/kubernetes/client-go v11.0.0+incompatible
 	github.com/logrusorgru/aurora v0.0.0-20191017060258-dc85c304c434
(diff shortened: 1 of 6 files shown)


coryschwartz:
I don't see the reason for caching the resolved names in application. Name resolution is typically time-based with TTLs and cached at the DNS resolver or another service nscd, etc -- To have (potentially) a different name resolved for each runid seems like a break from expectations.

Besides, once a resolved name is added to the cache, the only time it would be read again is for network changes, right? If the address of influxdb has changed between test start and a network change, isn't it desirable that the routes would be changed?

nonsense:
> To have (potentially) a different name resolved for each runid seems like a break from expectations.

If a services is stopped and started between test runs, then the DNS will return the new IP on Docker or Kubernetes - I don't think Docker or Kubernetes return stale DNS records.

> the only time it would be read again is for network changes, right?
Yes - we are block routes based on IPs, so we need to resolve the IPs of the services we want to allow.

> If the address of influxdb has changed between test start and a network change, isn't it desirable that the routes would be changed?
I'd say no. Resizing clusters, and moving/upgrading/services should happen in between test runs.

We don't have zero-downtime, and we assume services are available during the test runs.

> Why cache in the application and not just resolve directly?

Not sure what you mean with that.

nonsense:
@coryschwartz note that the `follow-up` for `confirm that the Redis client continue to work if Redis moves to another IP. Redis client ideally should handle re-connects.` depends on https://github.com/testground/sdk-go/pull/18

nonsense:
@coryschwartz following our conversation from yesterday about whether we need to do any kind of caching of resolved IPs - I'd rather not address this change in this PR. We already have caching on `master` for the necessary services (once when we start the `sidecar`), and this PR is just improving on that.

If we want to avoid caching altogether, we have to install node local DNS caches on Kubernetes, and investigate if we want to do something on `local:docker` and `local:exec` runners - this is a bit more elaborate, so let's first fix the bugs, and maybe improve further on it in follow up PRs. This change is rather simple, so we can easily revert when we find a better solution.

Issue #1041, opened 2020-05-27
add `version` command


nonsense:
We need to have a version command, which shows us the commit and/or the tag of the build.

Build time argument is probably fine for that. (https://blog.alexellis.io/inject-build-time-vars-golang/)

PR #1045, opened 2020-05-28
add version command

nonsense:
Addresses: https://github.com/testground/testground/issues/1041

---

Having a git commit is already a nice improvement and would make our lives easier when debugging other clusters (gossipsub, drand, etc.)

We should consider how we want to add a `semver` in the future.

### DIFF
--- a/main.go
+++ b/main.go
@@ -21,7 +21,6 @@ func main() {
 	app.Flags = cmd.RootFlags
 	// Disable the built-in -v flag (version), to avoid collisions with the
 	// verbosity flags.
-	// TODO implement a `testground version` command instead.
 	app.HideVersion = true
 	app.Before = func(c *cli.Context) error {
 		configureLogging(c)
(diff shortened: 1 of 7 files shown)


main.go
```diff
@@ -15,7 +15,7 @@ func main() {
 	app := cli.NewApp()
 	app.Name = "testground"
 	app.Usage = "a platform for testing, benchmarking, and simulating distributed and p2p systems at scale"
-	app.Description = "testground is a platform for testing, benchmarking," +
+	app.Description = "testground is a platform for testing, benchmarking, " +
```
coryschwartz, line 18: On Line 22, we should remove the comment to implement a testground version command, since this accomplishes that. 

Also I would be okay with justing the urfave version

nonsense: For now we can use the built in version functionality in `urfave` because of the verbosity flags I think. Removed the TODO.

Issue #1044, opened 2020-05-27
introduce a supervisor component


nonsense:
### Requirements

-   As a user I should be able to queue a run or build job on Testground
-   As a user I should be able to check the status of my job
-   As a user if my job is complete, I should be able to view all relevant information for that job, which is a JSON marshaled output:
	-   Build tasks:
		-   If successful: resulting artifact
		-   If failed: compilation error or any other error
	-   Run tasks:
		-   If completed: all outputs related to the task, which are runner dependent, such as:    
			-   cluster:k8s
				-   Total instances per group (succeeded, failed, unknown, pending)
				-   Relevant events: OOM failures, ImagePull errors, not enough resources to schedule pod events, etc.
				-   Stages history + timestamps
			-   local:docker
		-   If in-progress:
			-   cluster:k8s
				-   Total instances per group (succeeded, failed, unknown, pending)
				-   Relevant events: OOM failures, ImagePull errors, not enough resources to schedule pod events, etc.
				-   Stages history + timestamp
-   As a user I should be able to use Testground synchronously, as we currently use it, i.e. `testground build` and `testground run` with `--wait` flag running synchronously and interactively.
    

### Responsibilities

-   Keeps state of `Scheduled` and `InProgress` tasks.    
-   Moves tasks to `Archive` store when they get to a `Completed` state.
-   When a test run is terminal (100% of pods are not `Running`, and do not stay `Pending` for >timeout, or we hit global timeout, or all pods/containers enter terminal state) task is moved to `Archive`
-   Runs `n` tasks concurrently (we start with N=1)
-   Tracks and supervises test runs and all test instances
-   Should be able to survive restart, ideally even if there are `InProgress` tasks
-   Should be upgradable in between test runs (draining on OS signal)
-   Doesn’t determine whether a test run is `Successful` or `Failed`, just `Completed`, `Scheduled`, or `InProgress`
    

### Integrate `supervisor` in `testground daemon`

-   As a developer, when I submit a job (run or build) to Testground, I should:
-   Get back a unique <JOB_ID> from Testground so that I can query job status while it is running.
-   Get back an error if there are too many jobs with status Scheduled and InProgress, so that I can try again later.

This task involves:
-   Remove the synchronous `engine.DoRun` and `engine.DoBuild` calls as part of the handlers, and change to calls to `supervisor`.
-   Add `--wait` flag to `testground run` and `testground build`, which will now:
	-   Submit a job and get back a unique <JOB_ID>
[...]

Issue #1043, opened 2020-05-27
introduce concept of jobs to Testground


nonsense:
Testground should formally models the concept of “jobs”:
- Build and run are jobs.
- Not all jobs execute synchronously any longer.
- They are queued in the job scheduling queue (R03).

See R2 under https://docs.google.com/document/d/19dA5SsiVDt43rnaSiaxNtXYYBN_MVaikvi0JFvpYlos/edit?ts=5ecaad42#

coryschwartz:
If we decide to go with https://github.com/testground/testground/pull/1051ffor R03  then then we probably need to save the multipart request from the client somewhere. EFS? In the database also? I'm not sure where we want to put that.

I included in the model for the task a "state" to communicate to the client where the job is in the process.

PR #1051, opened 2020-05-30
persistent priority task queue

coryschwartz:
This is a priority task queue which persists to level-db.

A new task comes in
`TaskStorage.Push` to add the item to the queue. The task will be rejected if the queue length is too long.

Retrieve a task to run:
`TaskStorage.Pop` to get the highest-priority task. This pops the item off of the in-memory queue, but it does *NOT* delete it from the storage. The task will remain in storage until deleted.

Lifecycle of the task:
`TaskStorage.SetState` will change the *state* of the task. This is just a marking to indicate when the task is building, running, or complete. My intention is that the runner/builder will keep the state up to date with what it's doing.

Get information about a tasl:
`TaskStorage.Get` will get a task from storage by its ID. This is to permit a client to get an update about a previously-submitted job, the HTTP server can return the current state of the job.

Delete the task:
`TaskStorage.Delete` will remove the task from storage.

This does not currently manage the content of plans, which could be large and might be on EFS. The data model might need to be changed. 

This relates to R3 of https://github.com/testground/testground/issues/1044
and also related to https://github.com/testground/testground/issues/1043


pkg/daemon/task.go
```diff
@@ -0,0 +1,61 @@
+package daemon
+
+import "time"
+
+const (
+	TaskStateRequested TaskState = iota // when client requests a task
```
nonsense, line 6: I think as a start having only `Requested`, `In progress` and `Completed` would be sufficient for both `build` and `run` jobs. At the moment Testground doesn't have `Requested`, all jobs go directly to `In progress`, and the way we detect `Completed` is by counting the containers on a given environment.

pkg/daemon/persist.go
```diff
@@ -0,0 +1,155 @@
...
+	heap.Push(s.tq, tsk)
+	return nil
+}
+
+// get the next item from the priority queue
+func (s *TaskStorage) Pop() (*Task, error) {
+	s.mux.Lock()
+	defer s.mux.Unlock()
+	if s.Len() == 0 {
+		return nil, nil
```
nonsense, line 89: I think we want to return an error if we try to `Pop` and empty queue. `nil` is a bit dangerous I think.

pkg/daemon/task.go
coryschwartz: We could definitely get rid of the `requested` state. What I was thinking is that clients submit a task in the `requested` state to the HTTP server, which pushes the task onto the queue and then responds with the task `scheduled` state.

Later, the client can get an update about the state of the task, which might be in any of the states (other than requested) -- For the server to render an update to the client, it's a simple matter of reading it from the database.

We could remove separate building and running states too;  that probably is more detail than necessary.

I wonder if we need a `failed-to-run` state of some kind to indicate build failures, etc.

pkg/daemon/persist.go
coryschwartz: You may be right! I'll return an error here.

pkg/task/storage.go
```diff
@@ -0,0 +1,155 @@
...
+		if tsk.State != TaskStateComplete {
+			heap.Push(s.tq, tsk)
+		}
+	}
+	iter.Release()
+	return iter.Error()
+}
+
+// Add an item to the priority queue
+func (s *TaskStorage) Push(tsk *Task) error {
```
nonsense, line 64: I find this name a bit confusing - `TaskStorage` sounds like the thing that takes care of storage for `Tasks`, but actually this type has the semantics for managing the state as well as storing - we modify `tsk.State` on Push, check for `s.Max`, etc.

pkg/task/storage.go
```diff
@@ -0,0 +1,155 @@
...
+	tsk.State = TaskStateScheduled
+	key := []byte(tsk.ID)
+	val, err := json.Marshal(tsk)
+	if err != nil {
+		return err
+	}
+	err = s.db.Put(key, val, nil)
+	if err != nil {
+		return err
+	}
```
nonsense, line 79: I'd expect `TaskStorage` to take care only of this section from this method, but it is also maintain a local queue (`s.tq`) and modifying state.

pkg/task/storage.go
```diff
@@ -0,0 +1,155 @@
...
+	heap.Push(s.tq, tsk)
+	return nil
+}
+
+// get the next item from the priority queue
+func (s *TaskStorage) Pop() (*Task, error) {
+	s.mux.Lock()
+	defer s.mux.Unlock()
+	if s.Len() == 0 {
+		return nil, fmt.Errorf("attempted pop on empty queue, returning nil pointer!")
```
nonsense, line 89:
```suggestion
		return nil, errors.New("attempted pop on empty queue, returning nil pointer!")
```

pkg/task/storage.go
```diff
@@ -0,0 +1,155 @@
...
+		if t.ID == tsk.ID {
+			heap.Remove(s.tq, i)
+			break
+		}
+	}
+	return nil
+}
+
+// Change the state of a task
+func (s *TaskStorage) SetState(tsk *Task, state TaskState) error {
```
nonsense, line 118: This is a weird signature. Normally in Go, unless you deserialise a struct, you don't modify function arguments.

It'd be nice to refactor this to not have such side effects, something like.
```
tsk.State = state // modify tsk in memory=
s.Put(tsk) // update in place `tsk` in the database
```

pkg/task/storage.go
```diff
@@ -0,0 +1,155 @@
...
+	err = s.db.Put(key, newbuf, s.WriteOpts)
+	if err != nil {
+		return err
+	}
+	tsk.State = state
+	return nil
+}
+
+// Get a task by key
+func (s *TaskStorage) Get(key string) (*Task, error) {
```
nonsense, line 140: This is a place where I'd expect to pass `task` pointer so that we can deserialise in it. Doing that could avoid escaping to the heap AFAIK.
```suggestion
func (s *TaskStorage) Get(key string, *Task) error {
```

pkg/task/storage.go
nonsense: Maybe just `Queue` is a better name for this type, then the current `TaskQueue` doesn't need to be exported, it could be a private type AFAICT. Developers would have to interact only with this `Queue` I think, and with the `Task` type.

pkg/task/storage.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// Method for closing the database. Always run Close when finished.
+func (s *TaskStorage) Close() error {
+	s.tq = nil
+	return s.db.Close()
+}
+
+// Read everything from the database into memory. Typically, you will not need to run this.
+func (s *TaskStorage) Reload() error {
```
nonsense, line 44: I think we should be doing this once on startup - maybe just call it `load` and not export it, as I think we only need to call it when we initialise this type.

pkg/task/task.go
```diff
@@ -0,0 +1,61 @@
+package task
+
+import "time"
+
+const (
+	TaskStateRequested TaskState = iota // when client requests a task
+	TaskStateScheduled                  // daemon accepts the task into the queue
+	TaskStateBuilding                   // daemon is building the task
+	TaskStateRunning                    // daemon is running the task
+	TaskStateComplete                   // daemon is not running the task. ready to collect.
```
nonsense, line 10: These are under the `task` package, you can drop the `Task` or even `TaskState` prefix.

`TaskState` -> `State`.

nonsense: As commented on Slack, I think it'd be cleaner if we have the following states:
1. `requested` (added to the queue, but not processed)
2. `in progress / processing` (popped from the queue, and building or running or both, depending on the request)
3. `completed` (no longer running (could be successful or failed or any other terminal state) ; outputs are ready to be processed; note that at the moment we don't have history on `failed` tasks (for example is a pod goes OOM and we don't have `keep_alive`, we won't find out - we need to think about that).

---

I think these 3 are enough, not sure why we need 5 states.

pkg/daemon/task.go
nonsense: I suggest we start simple and don't have a `failed-to-run` - that would be `in progress` going to `completed` after the global timeout (for example if you try to schedule too many testplan instances, and the cluster doesn't have autoscaler setup).

1. requested / scheduled (added to the queue, but not processed) - i don't think we need a separate states here

2. in progress / processing (popped from the queue, and building or running or both, depending on the request)

3. completed (no longer running (could be successful or failed or any other terminal state) ; outputs are ready to be processed; note that at the moment we don't have history on failed tasks (for example is a pod goes OOM and we don't have keep_alive, we won't find out - we need to think about that).

raulk: I agree with the state machine being modelled here. `Completed` should represent the terminal state, and we should probably model a `result` field with various outcomes:

* success
* failed

For the failed state, we probably want to have an `error` field in the data model, so that we can save the error for display.

pkg/daemon/persist.go
raulk: The idiomatic Go way is to return an `ok bool`, rather than an error. Would that work here?

pkg/task/storage.go
```diff
@@ -0,0 +1,155 @@
...
+	"sync"
+
+	"github.com/syndtr/goleveldb/leveldb"
+	"github.com/syndtr/goleveldb/leveldb/opt"
+)
+
+// TaskStorage is a leveldb-backed priority queue.
+// storage is persisted before the in-memory queue to prevent inconsistency between restarts; what
+// you see in the queue is the same as what is written the database
+type TaskStorage struct {
```
raulk, line 16: The task data model will evolve over time. It's important to incorporate a "schema version" field at the database level since the beginning.

- New testground versions may update the task schema.
- On `Load`, we should perform any [schema migration](https://en.wikipedia.org/wiki/Schema_migration) by applying incremental migration logic.

pkg/task/storage.go
```diff
@@ -0,0 +1,155 @@
...
+	s.mux.Lock()
+	defer s.mux.Unlock()
+	if s.Len() == 0 {
+		return nil, fmt.Errorf("attempted pop on empty queue, returning nil pointer!")
+	}
+	tsk := heap.Pop(s.tq).(*Task)
+	key := []byte(tsk.ID)
+	err := s.db.Delete(key, nil)
+	if err != nil {
+		return nil, err
```
raulk, line 95: I think this error should either be logged only, or should make the entire system panic. But returning it like that is not good, because we did actually pop a task from our memory representation, yet failed to deliver it to the caller.

pkg/task/storage.go
raulk: This should not be an error... The idiomatic go way is to return an `ok bool` return value.

pkg/task/storage.go
```diff
@@ -0,0 +1,155 @@
...
+
+	"github.com/syndtr/goleveldb/leveldb"
+	"github.com/syndtr/goleveldb/leveldb/opt"
+)
+
+// TaskStorage is a leveldb-backed priority queue.
+// storage is persisted before the in-memory queue to prevent inconsistency between restarts; what
+// you see in the queue is the same as what is written the database
+type TaskStorage struct {
+	Max       int
```
raulk, line 17: What would happen if the value of `Max` changes across restarts?

pkg/task/storage.go
```diff
@@ -0,0 +1,155 @@
...
+	Path      string
```
raulk, line 18: Users probably shouldn't be setting these values on the struct directly. We probably want a constructor:

```go
path := filepath.Join(envconfig.Dirs().Work(), "tasks.db")
tqueue, err := taskqueue.Open(path)
```

pkg/task/storage.go
```diff
@@ -0,0 +1,155 @@
...
+	DBOpts    *opt.Options
+	WriteOpts *opt.WriteOptions
+	ReadOpts  *opt.ReadOptions
```
raulk, line 21: I wouldn't make these settable from the outside. This data structure has a concrete purpose (to serve as a persistent task queue), and we should be configuring any required write and read options internally, as fit for that purpose.

pkg/task/storage.go
```diff
@@ -0,0 +1,155 @@
...
+	key := []byte(tsk.ID)
+	err := s.db.Delete(key, nil)
+	if err != nil {
+		return nil, err
+	}
+	return tsk, nil
+}
+
+// delete a task from the queue
+func (s *TaskStorage) Delete(tsk *Task) error {
```
raulk, line 101: I think this method should only require the ID. After all, if we introduce a `testground cancel <task_id>` command, we won't be populating the whole struct.

pkg/task/task.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+type TaskState int
+
+func (t TaskState) String() string {
+	return [...]string{
+		"requested",
+		"scheduled",
+		"building",
+		"running",
```
raulk, line 20: For the purposes of a task queue utility, all it should be knowing is that it's running. It should stay oblivious to the actual meaning of the job.

pkg/daemon/persist.go
nonsense: @raulk the way this `Queue` also encapsulates the actual storing, that would mean we have 3 arguments - return value, error, bool

I think it is just fine to return `ErrEmpty` and detect it, rather than add a 3rd argument.

pkg/task/storage.go
nonsense: I think changing to interface to include a `bool` return value, just so that we don't return `ErrEmpty` is not making it much cleaner to be honest.

pkg/task/storage.go
```diff
@@ -0,0 +1,155 @@
...
+// you see in the queue is the same as what is written the database
+type TaskStorage struct {
+	Max       int
+	Path      string
+	DBOpts    *opt.Options
+	WriteOpts *opt.WriteOptions
+	ReadOpts  *opt.ReadOptions
+	tq        *TaskQueue
+	db        *leveldb.DB
+	mux       sync.Mutex
```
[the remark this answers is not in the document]
nonsense: Or just embed if it protects the full struct.

pkg/daemon/task.go
coryschwartz: @raulk @nonsesne updated the model. Let me know what you think!

pkg/task/storage.go
nonsense: I agree we need a `schema` version field. I'd rather we don't dive into how migrations work for now, this shouldn't be very hard as soon as we need the first migration.

pkg/task/storage.go
```diff
@@ -0,0 +1,155 @@
...
+func (s *TaskStorage) Close() error {
+	s.tq = nil
+	return s.db.Close()
+}
+
+// Read everything from the database into memory. Typically, you will not need to run this.
+func (s *TaskStorage) Reload() error {
+	s.mux.Lock()
+	defer s.mux.Unlock()
+	s.tq = new(TaskQueue)
```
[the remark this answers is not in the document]
coryschwartz: True, will change to New constructor.

pkg/task/storage.go
coryschwartz: Added a version.

pkg/task/storage.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+	s.db = db
+	return s.Reload()
+}
+
+// Method for closing the database. Always run Close when finished.
+func (s *TaskStorage) Close() error {
+	s.tq = nil
+	return s.db.Close()
+}
```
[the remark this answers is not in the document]
coryschwartz: I removed close ( and added the Sync option to write operations)

pkg/task/storage.go
coryschwartz: Changed to a New-* style constructor.

pkg/daemon/persist.go
coryschwartz: Used a named error -- the error would just be used to detect if the queue is popped while empty, so I could be convinced to switch it for a boolean.

pkg/task/storage.go
coryschwartz: Good point. Unexported TaskQueue and changed the name of TaskStorage to Queue.

pkg/task/storage.go
coryschwartz: Changed to named errors.

pkg/task/storage.go
coryschwartz: I changed this -- Now "Get" and "SetState' interact directly with the k-v database only, and their signatures take a string ID as the key.

pkg/task/storage.go
coryschwartz: Good point. Changing.

pkg/task/task.go
coryschwartz: Done!

pkg/task/storage.go
coryschwartz: If max increases, then more items will be allowed on. If max decreases, pushes would be denied until queue length is less than max.

pkg/task/storage.go
coryschwartz: True, and fixed!

pkg/task/storage.go
coryschwartz: removed.

pkg/task/task.go
```diff
@@ -0,0 +1,68 @@
...
+// TaskResultAbort: testground encountered an error and the task has not been scheduled.
+// testground will respond to client with a TaskResultAbort status if it is unable to enqueue
+// the reqeusted task.
+type TaskResultStatus int
+
+const (
+	ResultNone TaskResultStatus = iota
+	ResultSuccess
+	ResultFail
+	ResultAbort
```
nonsense, line 39: I am not sure why we need `ResultAbort` - at least at the moment I don't think we have such state on the Kubernetes runner.

pkg/task/task.go
```diff
@@ -0,0 +1,68 @@
...
+		"ResultFail",
+		"ResultAbort",
+	}[t]
+}
+
+// TaskResult (kind: struct)  contains a status code. If the status is not TaskResultNone or
+// TaskResultSuccess, relevant errors will be included in this struct.
+type TaskResult struct {
+	Status TaskResultStatus `json:"status"`
+	Errors []error          `json:"errors"`
```
nonsense, line 55: This is something we have to discuss on how to implement, and whether we need at the moment - for example if a test run fails, because out of 100 pods, 1 went out-of-memory, currently we don't detect that, and what would happen is that the test run will reach the global timeout and be killed (i.e. the way I understand the results above, we would just get a `ResultFail`, and upon inspecting the `outputs` we would notice that 99 have a context deadline exceeded, and 1 is killed).

Ideally we need to improve on this.

pkg/task/task.go
```diff
@@ -0,0 +1,68 @@
...
+
+// Task (kind: struct) contains metadata about a testground task. This schema is used to store
+// metadata in our task storage database as well as the wire format returned when clients get the
+// state of a running or secheduled task.
+type Task struct {
+	Version  int        `json:"version"`  // Schema version
+	Priority int        `json:"priority"` // scheduling priority
+	Created  time.Time  `json:"created"`  // datetime created
+	ID       string     `json:"id"`       // unique identifier for this task
+	State    TaskState  `json:"state"`    // State of the task
```
nonsense, line 66: @coryschwartz @raulk it is fairly simple to change State into a pair of <timestamp, State>, and assume that the current state is the last entry from the slice. This way we would have history of the state transitions and easily calculate how long tasks stayed in scheduled state, how long they took to process, etc.

Together with some metadata, that might be useful information to run some stats on once we go live.

pkg/task/task.go
```diff
@@ -0,0 +1,68 @@
...
+	Result   TaskResult `json:"result"`   // result of the task, when terminal.
```
nonsense, line 67: We currently don't have a good definition of what the result of a task is:
1. Is it the result from the pods/containers execution vs `outcome` in test plan.
2. Is one failed test plan instance considered a task failure?

I think it makes sense dropping `Result` altogether from the `Task` struct, or modelling it to give information per test instance based on the pod/container status (9999 completed, 1 failed), or something like that.

pkg/task/task.go
nonsense: On a second thought - we definitely want to propagate `events` or `statuses` from Kubernetes to Testground, because in TaaS developers won't have access to the cluster, all their interactions would go though Testground.

Therefore if out of 1000 pods:
997: Complete successfully
1: Panic because of testplan code
1: Fails due to pulling image
1: Fails due to OOM

I imagine we want to propagate all this information back to the user, so we should model it under the schema that will track historical state of jobs.

pkg/task/task.go
coryschwartz: I was trying to indicate a difference between failures of the testground service and the status of the plan run itself. I was imaging this as "Fail" -- the plan failed, as indicated as the final result of the plan run, and "Abort" to mean testground tried to run this task and couldn't do it for some reason.

pkg/task/task.go
coryschwartz: That's an interesting idea. I like that.

pkg/task/task.go
coryschwartz: We definitely need to figure out what we want to include in the result struct.

pkg/task/storage.go
```diff
@@ -0,0 +1,262 @@
...
+	if q.tq.Len() == 0 {
+		return nil, ErrQueueEmpty
+	}
+	tsk := heap.Pop(q.tq).(*Task)
+	return tsk, nil
+}
+
+// Change the state of a task in the K-V store
+// This method
+func (q *Queue) AppendTaskState(id string, state TaskState) error {
```
coryschwartz, line 202: @nonsense @raulk I am about *this close* to changing this method to being a method of *Task instead. The only reason I haven't done it is it would add a pointer to leveldb into the task struct which would sometimes be nil.

Currently, "Push" and "Pop" operate on *Task. Obviously, a Task being pushed onto the queue does not have a reference to leveldb, and I want to avoid having any half-functional Tasks anywhere.

Right now, `Task` is just a data model without methods, and I think this is a feature. The `Task` is the same object on client and server, in the database and on the wire. We could change this model out for apache thrift, a protobuf, avro or any other serialization technique without having to change anything else.

Should I try to make Queue stick to being a typical work Queue and move this method somewhere else or maybe change Queue to be called `TaskMaster` and embrase the part-queue-part-key-value nature of this thing.

pkg/task/task.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+func (t TaskResultStatus) String() string {
+	return [...]string{
+		"ResultNone",
+		"ResultSuccess",
+		"ResultPlanFail",
+		"ResultTestgroundFail",
+	}[t]
+}
```
nonsense, line 47: I think we don't need this for the time being. We discussed that for now we will just have an `output` for every job - for `build` jobs that would be the resulting artifact. For `run` jobs that would be a value that includes the state of all test plans, something similar to:

```
{"succeeded": 4642, "running": 158, "pending": 0, "failed": 0, "unknown": 0}
```

pkg/task/task.go
```diff
@@ -0,0 +1,72 @@
...
+
+// DatedTaskState (kind: struct) is a TaskState with a timestamp.
+type DatedTaskState struct {
+	TaskState TaskState `json:"state"`
+	Entered   time.Time `json:"entered"`
```
nonsense, line 52: Nitpick, but I think `Created` is more idiomatic than `Entered`.
```suggestion
	Created   time.Time `json:"created"`
```

pkg/task/task.go
```diff
@@ -0,0 +1,74 @@
...
+
+// Task (kind: struct) contains metadata about a testground task. This schema is used to store
+// metadata in our task storage database as well as the wire format returned when clients get the
+// state of a running or secheduled task.
+type Task struct {
+	Version  int              `json:"version"`  // Schema version
+	Priority int              `json:"priority"` // scheduling priority
+	ID       string           `json:"id"`       // unique identifier for this task, specifically, a UUID
+	States   []DatedTaskState `json:"states"`   // State of the task
+	Result   TaskResult       `json:"result"`   // result of the task, when terminal.
```
nonsense, line 65: I think we can postpone defining `TaskResult` (and not include `RunResult` and `BuildResult` - they are runner/build dependent, for example the result from cluster:k8s would be different to the result from other orchestrators), and just leave it as a `string` or `interface{}` here for now. WDYT?

pkg/task/task.go
```diff
@@ -0,0 +1,74 @@
...
+	Unknown   int `json:"unknown"`
+}
+
+// TaskResult (kind: struct)  contains a status code. If the status is not TaskResultNone or
+// TaskResultSuccess, relevant errors will be included in this struct.
+type TaskResult struct {
+	BuildResult BuildResult `json:"build"`
+	RunResult   RunResult   `json:"run"`
+	Errors      []error     `json:"errors"`
+}
```
nonsense, line 55: As the comment below - I don't think the queue should know about these structs - they should be generic and dependent on the builder/runner, as far as the queue and the supervisor are concerned, we just have some `result`.

pkg/task/task.go
coryschwartz: That sounds reasonable.

pkg/task/task.go
coryschwartz: For that matter, the task.States log and the Result could all be stored separately since they are not relevant to the queue

coryschwartz:
Merging -- currently this is all un-reachable code, but this gives a starting-point for other TAAS work.

pkg/engine/engine.go
```diff
@@ -224,8 +224,13 @@ func (e *Engine) DoBuild(ctx context.Context, comp *api.Composition, basesrc str
 
 			ow.Infow("performing build for groups", "plan", plan, "groups", grpids, "builder", builder)
 
+			id, err := uuid.NewUUID()
```
nonsense, line 227: @coryschwartz this is actually affecting the names we generate for containers/images.

PR #1053, opened 2020-06-01
build job for macOS

nonsense:
This PR is adding a CircleCI build job for macOS.

At the moment we can't really run `go test` on CircleCI for macOS, because it relies on Docker.

TODO FOLLOWUP:
- [ ] cache Go installation

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,7 @@ commands:
     description: "install go, checkout and restore cache"
     steps:
       - run:
-          name: "install go on macos"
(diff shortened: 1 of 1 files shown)


nonsense:
Note that tests are failing now, as we can't compile on MacOS. Will push another commit after we merge the fix in https://github.com/testground/testground/pull/1052 to confirm that CircleCI setup is correct.

nonsense:
Merging this as it is non-contentious and we better find out early in case we have issues with compilation on macOS.

Issue #1056, opened 2020-06-01
add dependency (initContainer) to test plan pods on the sidecar


nonsense:
We need to make our Kubernetes dependencies more robust and add a dependency to the test plan pods on the local sidecar on the host. - similarly to method 2 at https://www.alibabacloud.com/blog/kubernetes-demystified-solving-service-dependencies_594110

When we have this it will be simple to just schedule 10k pods, wait for the Kubernetes Autoscaler to kick in, and for the test plan pods to start on the newly created hosts.

Issue #1055, opened 2020-06-01
add healthz endpoint to sidecar


nonsense:
We need to know when the sidecar is up and running, so that we can add a dependency on testplan pods to wait for it when autoscaling a cluster (i.e. we don't want testplan instances to be started before the stack is ready - flannel -> weave -> sidecar -> test plan pod).

We know that every host is running a sidecar, and sidecar is running on the host network, so we could just ping an `healthz` endpoints, for example `http://localhost:8883/healthz` to determine if  it is ready (see Method 2 at https://www.alibabacloud.com/blog/kubernetes-demystified-solving-service-dependencies_594110)

---

This will be needed prior to setting up the Kubernetes Autoscaler.

nonsense:
We are already starting an HTTP server with the default mux for `pprof`, so removing the `pprof` flag, and always starting the HTTP server is also an option.

PR #1057, opened 2020-06-01
sidecar: start http server by default

nonsense:
Addresses: https://github.com/testground/testground/issues/1055 and https://github.com/testground/testground/issues/1056

Also adds autoscaler support behind a feature flag.

Also adds configurable `collect` pod resource requests that were supposed to be addressed in #1008.

---

Ideally we don't want testplan pods to start on a given host before the dependencies have been started, and this is what this PR is about.

---

Related: Autoscaler PR on `testground/infra`: https://github.com/testground/infra/pull/43

---

TODO:
- [x] remove `pprof` flag from `testground/infra` repo when merging this.

### DIFF
--- a/pkg/cmd/sidecar.go
+++ b/pkg/cmd/sidecar.go
@@ -32,10 +32,6 @@ var SidecarCommand = cli.Command{
 			Usage:    "runner that will be scheduling tasks that should be managed by this sidecar; supported: 'local:docker', 'cluster:k8s'",
 			Required: true,
 		},
-		&cli.BoolFlag{
-			Name:  "pprof",
-			Usage: "enable pprof service on port 6060",
-		},
 	},
 }
 
@@ -44,12 +40,17 @@ func sidecarCommand(c *cli.Context) error {
 		return ErrNotLinux
 	}
 
-	if c.Bool("pprof") {
-		logging.S().Info("starting pprof")
-		go func() {
-			_ = http.ListenAndServe(":6060", nil)
-		}()
-	}
+	startHTTPServer()
 
 	return sidecar.Run(c.String("runner"))
+}
+
+func startHTTPServer() {
+	logging.S().Info("starting http server")
+	go func() {
+		http.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
+			fmt.Fprintf(w, "All good!")
+		})
+		_ = http.ListenAndServe(":6060", nil)
+	}()
 }
--- a/pkg/cmd/sidecar.go
+++ b/pkg/cmd/sidecar.go
@@ -48,9 +48,6 @@ func sidecarCommand(c *cli.Context) error {
 func startHTTPServer() {
 	logging.S().Info("starting http server")
 	go func() {
-		http.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
-			fmt.Fprintf(w, "All good!")
-		})
 		_ = http.ListenAndServe(":6060", nil)
 	}()
 }
--- a/pkg/runner/cluster_k8s.go
+++ b/pkg/runner/cluster_k8s.go
@@ -198,7 +198,7 @@ func (c *ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow *rpc
 	}
 
 	if !enoughResources {
-		return nil, errors.New("too many test instances requested, resize cluster if you need more capacity")
+		ow.Warnw("too many test instances requested, will have to wait for cluster autoscaler to kick in")
 	}
 
 	jobName := fmt.Sprintf("tg-%s", input.TestPlan)
--- a/pkg/runner/cluster_k8s.go
+++ b/pkg/runner/cluster_k8s.go
@@ -244,6 +244,13 @@ func (c *ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow *rpc
 					FieldPath: "status.podIP",
 				}},
 		})
+		env = append(env, v1.EnvVar{
+			Name: "HOST_IP",
+			ValueFrom: &v1.EnvVarSource{
+				FieldRef: &v1.ObjectFieldSelector{
+					FieldPath: "status.hostIP",
+				}},
+		})
 
 		podCPU := defaultCPU
 		if g.Resources.CPU != "" {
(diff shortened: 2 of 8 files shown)


pkg/runner/cluster_k8s.go
```diff
@@ -692,6 +699,27 @@ func (c *ClusterK8sRunner) createTestplanPod(ctx context.Context, podName string
...
+							MountPropagation: &mountPropagationMode,
+						},
+					},
+					Resources: v1.ResourceRequirements{
+						Limits: v1.ResourceList{
+							v1.ResourceMemory: resource.MustParse("10Mi"),
+							v1.ResourceCPU:    resource.MustParse("10m"),
+						},
+					},
+				},
```
nonsense, line 722: Ideally this should be behind a flag if the `testplan` needs `sidecar` or not.

nonsense: If we don't need the sidecar, maybe it will be necessary to add a `wait-for-flannel` initContainer.

pkg/cmd/sidecar.go
```diff
@@ -44,12 +40,17 @@ func sidecarCommand(c *cli.Context) error {
...
-	}
+	startHTTPServer()
 
 	return sidecar.Run(c.String("runner"))
 }
+
+func startHTTPServer() {
+	logging.S().Info("starting http server")
+	go func() {
+		http.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
```
coryschwartz, line 51: healthz?

pkg/runner/cluster_k8s.go
```diff
@@ -692,6 +699,27 @@ func (c *ClusterK8sRunner) createTestplanPod(ctx context.Context, podName string
 			},
 			RestartPolicy: v1.RestartPolicyNever,
 			InitContainers: []v1.Container{
+				{
+					Name:            "wait-for-sidecar",
+					Image:           "busybox",
+					ImagePullPolicy: v1.PullIfNotPresent,
+					Args:            []string{"-c", "until nc -vz $HOST_IP 6060; do echo \"Waiting for local sidecar to listen to $HOST_IP:6060\"; sleep 2; done;"},
```
coryschwartz, line 706: There might be a state in which the TCP socket is open, but still unhealthy -- i.e. missing permissions -- maybe wait until healthy?

pkg/cmd/sidecar.go
nonsense: At first I thought to check if HTTP on `/healthz` is responding, but then actually just decided to check if TCP server is running on 6060. I will remove it.

pkg/runner/cluster_k8s.go
nonsense: Missing permissions? Then everything would be failing, and that would be a misconfiguration on the environment.

Ideally we want to have an endpoint that says `sidecar is healthy` - but this is rather difficult - we'd have to run healthchecks on the CNI plugins as well. I think for now making sure that the process is running is good enough.

pkg/runner/cluster_k8s.go
nonsense: On a second thought, this is just fine - every host on Kubernetes must have a `sidecar`, so it is fine to wait for it.

pkg/runner/cluster_k8s.go
```diff
@@ -692,6 +699,27 @@ func (c *ClusterK8sRunner) createTestplanPod(ctx context.Context, podName string
 			},
 			RestartPolicy: v1.RestartPolicyNever,
 			InitContainers: []v1.Container{
+				{
```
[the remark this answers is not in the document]
nonsense: The `sidecar` has a dependency on `Redis` currently.

nonsense:
Tried this with https://github.com/testground/infra/pull/43 and it is working as expected - pods are in `Pending` mode, until new nodes are setup and `sidecars` become available on them.

pkg/runner/cluster_k8s.go
```diff
@@ -72,12 +72,6 @@ const (
 	NetworkInitialisationFailed     = "network initialisation failed"
 )
 
-var (
```
nonsense, line 75: This was supposed to be addressed in https://github.com/testground/testground/pull/1008

pkg/runner/cluster_k8s.go
```diff
@@ -110,9 +104,13 @@ type ClusterK8sRunnerConfig struct {
 	// Provider is the infrastructure provider to use
 	Provider string `toml:"provider"`
 
-	// Resources requested for each pod from the Kubernetes cluster
-	PodResourceMemory string `toml:"pod_resource_memory"`
-	PodResourceCPU    string `toml:"pod_resource_cpu"`
+	// Resources requested for each testplan pod from the Kubernetes cluster
```
nonsense, line 107: This was supposed to be addressed in https://github.com/testground/testground/pull/1008

pkg/runner/cluster_k8s.go
```diff
@@ -160,15 +158,8 @@ func (c *ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow *rpc
...
-	var defaultCPU, defaultMemory resource.Quantity
-	defaultCPU, err := resource.ParseQuantity(cfg.PodResourceCPU)
-	if err != nil {
-		defaultCPU = resource.MustParse("100m")
-	}
-	defaultMemory, err = resource.ParseQuantity(cfg.PodResourceMemory)
-	if err != nil {
-		defaultMemory = resource.MustParse("100Mi")
-	}
+	defaultCPU := resource.MustParse(cfg.TestplanPodCPU)
```
nonsense, line 161: @coryschwartz I think we want to panic here if `testplan_pod_cpu` is not set. Hardcoding default values in the codebase is confusing, because if folks misconfigure the `.env.toml` and do not apply these, would find out hours later that their pods have different resource requirements as expected.

This already happened once.

pkg/runner/cluster_k8s.go
```diff
@@ -900,10 +912,15 @@ func (c *ClusterK8sRunner) pushImagesToDockerRegistry(ctx context.Context, ow *r
 	return c.pushToDockerRegistry(ctx, ow, cli, in, ipo, uri)
 }
 
-func (c *ClusterK8sRunner) createCollectOutputsPod(ctx context.Context) error {
+func (c *ClusterK8sRunner) createCollectOutputsPod(ctx context.Context, input *api.CollectionInput) error {
```
nonsense, line 915: This was supposed to be addressed in #1008

pkg/runner/local_docker.go
```diff
@@ -121,7 +121,7 @@ func (r *LocalDockerRunner) Healthcheck(ctx context.Context, engine api.Engine,
 		ContainerConfig: &container.Config{
 			Image:      "iptestground/sidecar:edge",
 			Entrypoint: []string{"testground"},
-			Cmd:        []string{"sidecar", "--runner", "docker", "--pprof"},
+			Cmd:        []string{"sidecar", "--runner", "docker"},
```
nonsense, line 124: Removing `--pprof` flag, as now we always run an HTTP server on the sidecar.

PR #1059, opened 2020-06-02
introduce task abstraction to http server

coryschwartz:
Building on the storage model in https://github.com/testground/testground/pull/1051 this adds a task model.

Right now this is just the "build" handler with some modifications, but this shows the gist.

The handler takes requests and saves them to disk, then adds the task to the queue to be executed.

Fix: https://github.com/testground/testground/issues/1043 (when complete)
Fix https://github.com/testground/testground/issues/1044 (when complete)


coryschwartz:
Currently this works like this:

1. Start testground in "service" mode
2. Submit a build job with the client
`testground build single --builder docker:go  --plan example`
3. Use `curl` to see what was saved:
```
curl localhost:8080/task/a1ee7a18
{"t":112,"p":"SnVuICAyIDAzOjMxOjAzLjM4ODcyMwkbWzM0bUlORk8bWzBtCXsicHJpb3JpdHkiOjAsImNyZWF0ZWQiOiIyMDIwLTA2LTAxVDIwOjI0OjI5LjQ4MTY4NzUzMi0wNzowMCIsImlkIjoiYTFlZTdhMTgiLCJzdGF0ZSI6MX0JeyJyZXFfaWQiOiAiYjZkNmQ0YTEifQo="}

base64 -d <<<SnVuICAyIDAzOjMxOjAzLjM4ODcyMwkbWzM0bUlORk8bWzBtCXsicHJpb3JpdHkiOjAsImNyZWF0ZWQiOiIyMDIwLTA2LTAxVDIwOjI0OjI5LjQ4MTY4NzUzMi0wNzowMCIsImlkIjoiYTFlZTdhMTgiLCJzdGF0ZSI6MX0JeyJyZXFfaWQiOiAiYjZkNmQ0YTEifQo=
Jun  2 03:31:03.388723	INFO	{"priority":0,"created":"2020-06-01T20:24:29.481687532-07:00","id":"a1ee7a18","state":1}	{"req_id": "b6d6d4a1"}
```

This is clearly incomplete.
TODO:
* [ ] Add client support
* [ ] Cleaup all the code duplication OR add this as an option to the existing daemon command

pkg/engine/engine.go
```diff
@@ -49,6 +50,7 @@ type Engine struct {
 	runners map[string]api.Runner
 	envcfg  *config.EnvConfig
 	ctx     context.Context
+	stor    *task.TaskStorage
```
coryschwartz, line 53: I put this here, but maybe this doesn't need to be part of the engine. Maybe this is better situated as part of the under the Service struct.

pkg/service/service.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/testground/testground/pkg/config"
+	"github.com/testground/testground/pkg/engine"
+	"github.com/testground/testground/pkg/logging"
+
+	"github.com/gorilla/mux"
+	"github.com/pborman/uuid"
+)
+
+type Service struct {
+	server *http.Server
```
coryschwartz, line 19: Should the storage layer be here instead?

pkg/service/submit.go
```diff
@@ -0,0 +1,148 @@
...
+
+	"github.com/mholt/archiver"
+
+	"github.com/testground/testground/pkg/api"
+	"github.com/testground/testground/pkg/logging"
+	"github.com/testground/testground/pkg/rpc"
+	"github.com/testground/testground/pkg/task"
+)
+
+func (d *Service) submitHandler(engine api.Engine) func(w http.ResponseWriter, r *http.Request) {
```
coryschwartz, line 23: This is just the buildHandler, minus the part where it actually builds.

pkg/service/submit.go
```diff
@@ -0,0 +1,148 @@
...
+			State:    task.TaskStateScheduled,
+		}
+
+		engine.TaskStorage().Push(&tsk)
+
+		tgw.Infof("submitted %s.", ruid)
+	}
+}
+
+func consumeBuildRequest(r *http.Request, dir string) (*api.BuildRequest, string, string, error) {
```
coryschwartz, line 59: again, this is from the buildHandler

pkg/service/taskinfo.go
```diff
@@ -0,0 +1,34 @@
...
+	"encoding/json"
+	"net/http"
+
+	"github.com/gorilla/mux"
+
+	"github.com/testground/testground/pkg/api"
+	"github.com/testground/testground/pkg/rpc"
+)
+
+func (d *Service) taskinfoHandler(engine api.Engine) func(w http.ResponseWriter, r *http.Request) {
```
coryschwartz, line 13: simple ETL

pkg/config/env.go
```diff
@@ -11,18 +11,24 @@ type ConfigMap map[string]interface{}
...
+	Client      ClientConfig         `toml:"client"`
+	TaskStorage TaskStorageConfig    `toml:"taskstorage"`
 }
 
 func (e EnvConfig) Dirs() Directories {
 	return e.dirs
 }
 
+type TaskStorageConfig struct {
+	Path string `toml:"path"`
```
nonsense, line 28: I don't think we need that - we can just hardcode under the work directory for now, no?

pkg/service/loop.go
```diff
@@ -0,0 +1,46 @@
...
+		if err != nil {
+			return err
+		}
+
+		logging.S().Info("processing ", tsk.ID)
+
+		// TODO Actually do the build and run
+		// The data is already saved in $TESTGROUND_HOME.
+		// Maybe we want this to be configurable?
+		stor.SetState(tsk, task.TaskStateBuilding)
```
nonsense, line 36: As commented in the other PR, probably best to get rid of the side effect on `tsk`, and just mutate it, prior to serialising and updating.

pkg/service/loop.go
```diff
@@ -0,0 +1,46 @@
...
+		logging.S().Info("building ", tsk.ID)
```
nonsense, line 37: I think we should abstract `building` and `running`, and just `process` a task.

pkg/config/env.go
coryschwartz: True. If we want to change the storage path (i.e. to a persistent volume), we can just  move $TESTGROUND_HOME.

PR #1060, opened 2020-06-02
Feat/task http

coryschwartz:
Building on the storage model in https://github.com/testground/testground/pull/1051 this adds a task model.

Right now this is just the "build" handler with some modifications, but this shows the gist.

The handler takes requests and saves them to disk, then adds the task to the queue to be executed.

Fix: https://github.com/testground/testground/issues/1043 (when complete)


Robmat05:
This is an experimental PR, s/b draft. It is the same as #1059 and will close as ‘Discarded’

PR #1066, opened 2020-06-12
consistent content type for `collect outputs` by all runners

nonsense:


nonsense:
To be merged as part of https://github.com/testground/testground/pull/1067

PR #1076, opened 2020-06-17
migrate test plans and sidecar to new SDK.

raulk:
This PR migrates all test plans to the new SDK, which supports test plan initialization through a second parameter in the test function: `*run.InitContext`.

It also applies the `sync.{Interface=>Client}` and `sync.{=>Default}Client` renames. This is a breaking change, and users will need to migrate their code (should be super easy to do) when upgrading their SDK version.

---

Note: this PR is using replace directives that work with my local source layout; YMMV. But my goal was to integrate these changes and take them for a spin. I'm satisfied with the resulting API, and looking for confirmation from others. If it LGTY, I will release the SDK (minor release, to signal API changes) and I'll drop the replaces here.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -54,3 +54,5 @@ require (
 	k8s.io/apimachinery v0.17.0
 	k8s.io/client-go v0.0.0-20190706005506-4ed54556a14a
 )
+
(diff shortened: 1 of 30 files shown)


nonsense:
We have done that in another PR.

PR #1077, opened 2020-06-17
docker:go builder: move PreModDownload hook

nonsense:


pkg/build/docker_go.go
```diff
@@ -371,17 +371,17 @@ ARG BUILD_TAGS
 ENV TESTPLAN_EXEC_PKG ${TESTPLAN_EXEC_PKG}
 
 # PLAN_DIR is the location containing the plan source inside the container.
-ENV PLAN_DIR /plan/
+ENV PLAN_DIR /plan
+
+{{.DockerfileExtensions.PreModDownload}}
```
nonsense, line 376: @raulk It'd be nice if we can install packages as part of the testplan ala https://github.com/filecoin-project/lotus-testground/blob/master/manifest.toml#L12 that don't get invalidated every time we touch the `go.mod` from the testplan. WDYT?

pkg/build/docker_go.go
```diff
@@ -371,17 +371,17 @@ ARG BUILD_TAGS
 ENV TESTPLAN_EXEC_PKG ${TESTPLAN_EXEC_PKG}
 
 # PLAN_DIR is the location containing the plan source inside the container.
-ENV PLAN_DIR /plan/
+ENV PLAN_DIR /plan
```
nonsense, line 374: No trailing slash for the default `PLAN_DIR`, which makes it more intuitive to use in the following hooks.

pkg/build/docker_go.go
raulk: I don't expect us to modify the `go.mod` of a test plan very frequently, but I also don't see a reason why this shouldn't be merged.

raulk:
Needs rebasing on top of #1078 once that is merged.

pkg/build/docker_go.go
nonsense: Adding/removing debug packages and iterating on a testplan does hit `go.mod` often, resulting in invalidated cache.

PR #1078, opened 2020-06-17
docker:go: add option to skip runtime image.

raulk:
For builds that do more than just compiling go code (e.g. generating things, linking to dynamic libs, etc.), we might want to just use the build container as our runtime container. This is suboptimal, but it's a good way to bootstrap a test plan for a complex target, like Lotus/Filecoin.

### DIFF
--- a/pkg/build/docker_go.go
+++ b/pkg/build/docker_go.go
@@ -407,7 +407,7 @@ COPY . /
 
 RUN cd ${PLAN_DIR} \
     && go env -w GOPROXY="${GO_PROXY}" \
-    && GOOS=linux GOARCH=amd64 go build -o testplan ${BUILD_TAGS} ${TESTPLAN_EXEC_PKG}
+    && GOOS=linux GOARCH=amd64 go build -o /testplan ${BUILD_TAGS} ${TESTPLAN_EXEC_PKG}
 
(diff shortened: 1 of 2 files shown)


pkg/build/docker_go.go
```diff
@@ -424,4 +431,6 @@ COPY --from=builder /plan/testplan /
 
 EXPOSE 6060
 ENTRYPOINT [ "/testplan"]
+
+{{ end }}
```
nonsense, line 435: I think this should be above `EXPOSE 6060`.

nonsense: You also need:
```
{{ if .SkipRuntimeImage }}
RUN cp /testground_dep_list /
RUN /plan/testplan /
{{ end }}
```

raulk: Oh yeah, absolutely. And we need to put the test plan binary at the root.

raulk: Achieved the symmetry but differently: 

[https://github.com/testground/testground/pull/1078/commits/548da321a79cc807d1689fc69f1beec079873371 ](https://github.com/testground/testground/pull/1078/commits/548da321a79cc807d1689fc69f1beec079873371)

PR #1079, opened 2020-06-19
docker:go: rolling go build caches; make builders `Terminatable`; add basic termination logic to `docker:go`

raulk:
This commit introduces rolling/carry-over go build caches as a default behaviour in docker:go. Build caches images are maintained per test plan + go version tuple. Every time a build is started, we check if we have an image from a previous build, and if so, reuse it (clearing /plan, /sdk, /testground_dep_list first).

If a prior build cache image doesn't exist, we start with a blank go container.

The goal is to carry over the entire /go directory, which contains the pkg cache. We also set GOCACHE to /go/cache, in order to carry over the build object cache.

This behaviour can be disabled by setting the `disable_go_build_cache=true` option in the builder configuration.

The tradeoffs are as follows of using the rolling cache are as follows:

 * good: the rolling cache leads to super-fast go builds, as dependency sources and their respective build objects are cached across builds of the test plan. For example, if you import the same version of IPFS/Lotus/libp2p every time, you will no longer compile from their sources, but the go build will use the cached intermediate object files. This is akin to running a go build locally.

 * bad: the go toolchain records last access timestamps on every build in the filesystem (https://github.com/golang/go/blob/master/src/cmd/go/internal/cache/cache.go) which effectively invalidates Docker layer caching across builds. So even if you built exactly the same plan source, no Docker layer caches would be hit, because of the timestamps sliding. However, in practice, you will be iterating on the test plan source, and under that assumption, we have found that the end-to-end build is faster.

### DIFF
--- a/pkg/build/docker_go.go
+++ b/pkg/build/docker_go.go
@@ -1,6 +1,7 @@
 package build
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"os"
@@ -23,6 +24,8 @@ import (
 )
 
 const (
+	DefaultGoVersion = "1.14.4"
+
 	buildNetworkName = "testground-build"
 )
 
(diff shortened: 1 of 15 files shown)


pkg/build/docker_go.go
```diff
@@ -407,28 +549,36 @@ COPY . /
...
+#::: (OPTIONAL) RUNTIME CONTAINER
 #:::
 
-FROM ${RUNTIME_IMAGE} AS binary
+{{ if not .SkipRuntimeImage }}
+
+FROM ${RUNTIME_IMAGE} AS runtime
+
+# PLAN_DIR is the location containing the plan source inside the build container.
+ENV PLAN_DIR /plan
```
nonsense, line 570: I am missing a section that loads `testplan` in the root `/` dir when `.SkipRuntimeImage` is `false`. I think that would make the image incomptible.

pkg/build/docker_go.go
```diff
@@ -407,28 +549,36 @@ COPY . /
...
 COPY --from=builder /testground_dep_list /
-COPY --from=builder /testplan /
+COPY --from=builder ${PLAN_DIR}/testplan.bin /testplan
 
 {{.DockerfileExtensions.PostRuntimeCopy}}
 
+{{ else }}
+
+FROM builder AS runtime
+
```
nonsense, line 581: I am missing a section that loads `testplan` in the root `/` dir when `.SkipRuntimeImage` is `true`. I think that would make the image incompatible.

pkg/build/docker_go.go
```diff
@@ -407,28 +549,36 @@ COPY . /
...
 
 COPY --from=builder /testground_dep_list /
-COPY --from=builder /testplan /
+COPY --from=builder ${PLAN_DIR}/testplan.bin /testplan
 
 {{.DockerfileExtensions.PostRuntimeCopy}}
 
+{{ else }}
+
+FROM builder AS runtime
```
nonsense, line 580: Isn't this redundant? If the previous stage is the `builder`, why do we need this line?

raulk: To find the marker to extract the image ID of the last step of the builder image. Docker makes us parse stdout... (line 470).

nonsense: Actually I see it is used as a pointer, if that's the case, we probably want a comment on it.

pkg/build/docker_go.go
raulk: Fixed in 73e54d7.

raulk:
Fixes https://github.com/filecoin-project/oni/issues/32.

PR #1081, opened 2020-06-19
configurable build base image for docker:go

nonsense:


pkg/build/docker_go.go
```diff
@@ -56,7 +56,6 @@ type DockerfileExtensions struct {
 
 type DockerGoBuilderConfig struct {
 	Enabled    bool
-	GoVersion  string `toml:"go_version"`
```
nonsense, line 59: Note that we no longer use `go_version`, but we have to be explicit with the default base build image, such as `golang:1.14.4-buster` when upgrading Go.

PR #1083, opened 2020-06-22
oni -> master

nonsense:
This is a draft PR placeholder, which will contain all in-progress features, that we want to include in Testground properly once we are done with our in-house testplans.

Merged features:
1. do not block default route by sidecar on local:docker (https://github.com/testground/testground/pull/1082)
2. custom exposed ports by testplans


nonsense:
Rebased `oni` so that it includes `compositions: trickle default configs to all groups. `.

pkg/runner/cluster_k8s.go
```diff
@@ -753,6 +753,10 @@ func (c *ClusterK8sRunner) createTestplanPod(ctx context.Context, podName string
 							Name:          "pprof",
 							ContainerPort: 6060,
 						},
+						{
+							Name:          "lotus-rpc",
+							ContainerPort: 1234,
+						},
```
nonsense, line 759: Maybe this should be part of the `manifest.toml` where test plan can define what ports they want exposed on the test plan containers / pods. @raulk WDYT?

nonsense: I added `exposed_ports` to the `DockerfileExtensions`, which get picked up automatically by `local:docker`, as we are also using `PublishAllPorts`. 

I could iterate on them here as well, and map `pod ports` to the container in k8s.

PR #1085, opened 2020-06-23
add the ability to ship extra local sources in builds.

raulk:
This allows us to create isomorphism in complex builds, such as when test plans depend on external artifacts.

It also allows us to ship assets external to the test plan source.

Additionally, it unlocks the ability to override dependencies (e.g. via replace directives) replacing them with local copies. With the go builders, users no longer need to push changes in their upstream repos to get them picked up in their test plans.

## How is this done?

1. Manifests can now specify an `extra_sources` map, consisting of `builder => []paths`, which specify which paths to include as extras for which builders (conditional activation).
2. The paths are resolved relative to the plan source directory (where the manifest resides), evaluating symlinks first. This makes use that if the plan under `$TESTGROUND_HOME/plans/planA` is a symlink, we pick up the extras from the symlinked location.
3. Extras are packed into a zip file, and are sent in the multipart request.
4. Builders are free to use these extras.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"mime"
 	"mime/multipart"
 	"net/http"
 	"net/textproto"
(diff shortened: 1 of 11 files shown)


pkg/client/client.go
```diff
@@ -70,32 +71,19 @@ func (c *Client) Close() error {
...
-		}
-		files := make([]string, 0, len(fis))
-		for _, fi := range fis {
-			files = append(files, filepath.Join(dir, fi.Name()))
-		}
-
+	// writeZippedDirs a list of directories, zips them into a single zip file, and writes it on w.
+	// if toplevel=true, it will retain the toplevel directories, so if /abc, /def are passed, the resulting
+	// zip archive will contain /abc and /def.
+	// if toplevel=false, it will omit tihe toplevel directories and will place the contents of each
```
nonsense, line 84:
```suggestion
	// if toplevel=false, it will omit the toplevel directories and will place the contents of each
```

pkg/cmd/build.go
```diff
@@ -161,7 +162,25 @@ func doBuild(c *cli.Context, comp *api.Composition) ([]api.BuildOutput, error) {
 	}
 
 	req := &api.BuildRequest{Composition: *comp}
-	resp, err := cl.Build(ctx, req, planDir, sdkDir)
+
+	// if there are extra sources to include for this builder, contextualize
+	// them to the plan's dir.
+	builder := comp.Global.Builder
+	extra, ok := manifest.ExtraSources[builder]
+	if ok && len(extra) != 0 {
```
nonsense, line 170: I think this `if` is redundant. You can `range` over a `nil` slice, it is just a no-op.

PR #1087, opened 2020-06-25
trying squash of docker image layers

nonsense:


nonsense:
This doesn't work in its current form, as squashing happens in the end, but we use a pointer to an intermediate layer to do the rolling go cache...

PR #1089, opened 2020-06-26
compositions: trickle default configs to all groups.

### LINKED ISSUE #521: Composition field cascading/inheritance
Example:

```toml
[global.run]
test_params = { foo = 100, bar = "xyz" }
```

Test parameters defined in the global block would propagate down to all groups. Any test parameters defined in groups would override the global ones. The challenge becomes how to _UNSET_ a globally set parameter within a group, given that TOML does not support nil values 🤷‍♂️

raulk:
Compositions can now specify `[build]` and `[run]` subtables in the `[global]` table.

Build and run settings in groups will inherit the default configuration. The coalescing will happen in this manner:

* `build.dependencies` => the default dependency set will be merged into the local one, with the local entries taking precedence.
* `build.selectors` => no merging; the local selector list will override the default selector list entirely.
* `run.artifact` => local artifacts take precedence over default.
  * (known issue: if a global artifact is set, there's no way to "unset" it for a group, so that a build is triggered.)
* `run.test_params` => the default test parameters will be merged into the local map, with the local entries taking precedence.

---

Fixes #521.
Closes #648 (fix included here).

### DIFF
--- a/pkg/api/composition.go
+++ b/pkg/api/composition.go
@@ -16,6 +16,20 @@ var compositionValidator = func() *validator.Validate {
 	return v
 }()
 
+type Groups []*Group
+
+func (gs Groups) Validate() error {
+	// validate group IDs are unique
+	m := make(map[string]struct{}, len(gs))
+	for _, g := range gs {
+		if _, ok := m[g.ID]; ok {
+			return fmt.Errorf("group ids not unique; found duplicate: %s", g.ID)
+		}
+		m[g.ID] = struct{}{}
+	}
+	return nil
+}
+
 type Composition struct {
 	// Metadata expresses optional metadata about this composition.
 	Metadata Metadata `toml:"metadata" json:"metadata"`
@@ -25,7 +39,7 @@ type Composition struct {
 
 	// Groups enumerates the instances groups that participate in this
 	// composition.
-	Groups []*Group `toml:"groups" json:"groups" validate:"unique=ID"`
+	Groups Groups `toml:"groups" json:"groups" validate:"required,gt=0"`
 }
 
 type Global struct {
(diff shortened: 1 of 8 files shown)


raulk:
Fixes https://github.com/filecoin-project/oni/issues/62.

pkg/api/composition.go
```diff
@@ -45,11 +45,21 @@ type Global struct {
...
 	// Runner is the runner we're using.
 	Runner string `toml:"runner" json:"runner" validate:"required"`
 
 	// RunConfig specifies the run configuration for this run.
 	RunConfig map[string]interface{} `toml:"run_config" json:"run_config"`
+
+	// Run applies global run defaults that trickle down to all groups, such as
+	// test parameters or build artifacts. Groups can override these in their
+	// local run definition.
+	Run *Run `toml:"run" json:"run"`
```
nonsense, line 62: I wonder if we can come up with slightly better names here, as it is starting to get confusing - Build and BuildConfig, Run and RunConfig...

Maybe:

```
Build -> GroupBuildParams
Run -> GroupRunParams
```

The idea is that Build and Run are applied to the Groups, whereas BuildConfig and RunConfig are global for the given build and run.

In any case that can be done as a follow up or not at all.

raulk: Yup, we need to deduplicate between build, builder, run, runner, parameters, config...

nonsense:
@raulk I think this can go straight to `master` actually.

raulk:
@nonsense SGTM; alternatively we could merge into `oni`, and `oni` into `master`, but admittedly, `oni` is still in flux until we're fully satisfied with the build cache.

nonsense:
@raulk `oni` is not merged to `master` due to https://github.com/testground/testground/pull/1082

I think the build cache functionality is mostly on master already, but is opt-in, so it is fine when it is not perfect.

PR #1092, opened 2020-07-01
custom exposed ports

nonsense:


pkg/build/docker_go.go
```diff
@@ -44,14 +44,15 @@ type DockerGoBuilder struct {
...
-	PostRuntimeCopy string `toml:"post_runtime_copy"`
+	PreModDownload  string   `toml:"pre_mod_download"`
+	PostModDownload string   `toml:"post_mod_download"`
+	PreSourceCopy   string   `toml:"pre_source_copy"`
+	PostSourceCopy  string   `toml:"post_source_copy"`
+	PreBuild        string   `toml:"pre_build"`
+	PostBuild       string   `toml:"post_build"`
+	PreRuntimeCopy  string   `toml:"pre_runtime_copy"`
+	PostRuntimeCopy string   `toml:"post_runtime_copy"`
+	ExposedPorts    []string `toml:"exposed_ports"`
```
nonsense, line 55: @raulk actually that appears to be the wrong place to add the `exposed_ports`, as I really need them as part of the `runtime` configuration. The current impl. works for `local:docker` (due to `PublishAll`), but not for `cluster:k8s`.

I think I will opt for specifying the exposed ports as part of the run configuration somehow, so that we expose them when starting the container, and not in the Dockerfile as EXPOSE arguments.

nonsense:
Closing in favour of https://github.com/testground/testground/pull/1093

Issue #1094, opened 2020-07-06
make exposed ports isomorphic across runners


raulk:
https://github.com/testground/testground/pull/1093 introduced the ability to expose local ports through the runner's environment. This is used in `local:docker` and `cluster:k8s` to expose arbitrary ports, such as API endpoints and the like to the host environment in order to facilitate debugging.

The way it was implemented, using static ports, and assumes that the test plans will be exposing those ports. This is fine in environments that rely on container isolation, but it doesn't work for `local:exec`.

For example, launching 5 instances of a test plan that exposes port 6060 will work in `local:docker` and `cluster:k8s`, but will fail on `local:exec` due to port collision.

We need to fix this, because I would really like to use `local:exec` for tests that do not require traffic shaping or sidecar functionality.

raulk:
In the same spirit as `extra_sources`, a fix would be to scope mappings to runners, and label them with a name, borrowing the idea from `docker compose`:

<img width="381" alt="image" src="https://user-images.githubusercontent.com/1101242/86585564-908a7900-bf7e-11ea-8750-5c6365d74b06.png">

This would inject `PPROF_PORT`, `NODE_RPC_PORT` and `MINER_RPC_PORT` environment variables in the runenv for `local:docker` and `cluster:k8s`. They would be undefined for `local:exec`, and we would fall back to 0 (random), to avoid port collisions.

raulk:
Actually, it's simpler than that. We don't need per-runner configs. Because `exposed_ports` are mapped into runner-specific configurations, the runners that don't support this will simply not map them into env variables. What we need to make this work is the second part (injecting these values into env vars).

raulk:
This was done in:

* https://github.com/testground/testground/pull/1095
* https://github.com/testground/testground/pull/1096

And merged in to the `oni` branch, which will be eventually merged into `master`.

PR #1095, opened 2020-07-06
inject port numbers as env variables.

raulk:
This facilitates isomorphism across runners, because test workloads can now rely on port numbers being set via environment variables, instead of hard-coded values in source.

- port numbers are now specified as a map `label => port`, _à la Docker compose_.
- `local:docker` and `cluster:k8s` support port mappings and thus inject the port numbers as env variables with the name `${LABEL}_PORT`.
- `local:exec` will not inject anything because it doesn't support this feature (TODO/UX: print a warning in the runner informing that port mappings are not applied?).
- https://github.com/testground/sdk-go/pull/24 allows us to load such port numbers, and specify fallbacks if they're not set (e.g. `0` -- random).

---

Fixes https://github.com/testground/testground/issues/1094.

### DIFF
--- a/pkg/runner/cluster_k8s.go
+++ b/pkg/runner/cluster_k8s.go
@@ -115,7 +115,7 @@ type ClusterK8sRunnerConfig struct {
 	CollectOutputsPodMemory string `toml:"collect_outputs_pod_memory"`
 	CollectOutputsPodCPU    string `toml:"collect_outputs_pod_cpu"`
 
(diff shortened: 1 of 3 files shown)


nonsense:
> port numbers are now specified as a map label => port, à la Docker compose.

I believe this is Docker in general, not just Docker Compose. We do have all these ports as env variables for all Docker runners right now.

raulk:
> I believe this is Docker in general

Docker itself does not attach tags/labels/service names to ports.

PR #1097, opened 2020-07-09
add _star exclude pattern to docker context

nonsense:


pkg/docker/image.go
```diff
@@ -36,7 +36,9 @@ func defaultBuildOptsFor(name string) *types.ImageBuildOptions {
 // the Name, and the constructed options are sent to the docker client.
 // The build output is directed to stdout via PipeOutput, and also returned from this function.
 func BuildImage(ctx context.Context, ow *rpc.OutputWriter, client *client.Client, opts *BuildImageOpts) (string, error) {
-	buildCtx, err := archive.TarWithOptions(opts.BuildCtx, &archive.TarOptions{})
+	buildCtx, err := archive.TarWithOptions(opts.BuildCtx, &archive.TarOptions{
+		ExcludePatterns: []string{"$_*"},
```
raulk, line 40: What's the dollar sign for?

nonsense: Got confused, I meant to add ^ as a start ($ is for ending)...

In any case this is not working, as we need to prepend the path for the `BuildCtx` passed in, such as `/root/testground/data/work/requests/e636068c/plan/_*`

PR #1099, opened 2020-07-16
feat: `docker:go` cross-repo dependency replacement

### LINKED ISSUE #48: `docker:go` builder: Allow cross-repo replace directives
Currently, they only allow selecting a branch. Ideally, I'd be able to point at a different repo or even a local directory.

hacdias:
This PR should close #48 by allowing cross-repository replacement directives for dependencies on the `docker:go` builder. I followed @raulk instructions on [this comment](https://github.com/testground/testground/issues/48#issuecomment-536328534). Some thoughts:

> 1. Adopt go.mod notation in the CLI `--dep` parameter: `<dependency>=[<target>]@<version>`. When `target` is missing, we assume no remapping of repos.

Done.

The current iteration is forcing the `@` to be present just like the specification. Should it be optional when we only have a version?

>2. Adjust the components mentioned above. The `build.Input.Dependencies` field turn into a `map[string]DependencyTarget`, where `DependencyTarget` is a struct containing the target and version.

Done.

>3. In the future, to preserve correctness, builders not supporting this feature (e.g. a hypothetical `docker:js` builder that relies on `npm install` and `package.json`) will need to return an error if DependencyTarget.Target is not empty.

Wouldn't it be better to just parse whatever is after the `=` signal on each builder, allowing for a different syntax per builder, according to the target languate?

### DIFF
--- a/pkg/api/builder.go
+++ b/pkg/api/builder.go
@@ -51,11 +51,11 @@ type BuildInput struct {
 	// builder. In the case of go builders, this field maps to build tags.
 	Selectors []string
 
-	// Dependencies are the versions of upstream dependencies we want to build
-	// against. For a go build, this could be e.g.:
+	// Dependencies are the versions and target of upstream dependencies we wants
+	// to build against. For a go build, this could be e.g.:
 	//  github.com/ipfs/go-ipfs=v0.4.22
-	//  github.com/libp2p/go-libp2p=v0.2.8
-	Dependencies map[string]string
+	//  github.com/libp2p/go-libp2p=github.com/user/fork@v0.2.8
+	Dependencies map[string]DependencyTarget
 
 	// BuildConfig is the configuration of the build job sourced from the test
 	// plan manifest, coalesced with any user-provided overrides.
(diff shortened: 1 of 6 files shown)


nonsense:
I haven't tested this, but LGTM!

dgrisham:
Hey! Should this allow using local versions of packages as well? Seems like that'd be more involved than simply pointing to a different git repo, but wasn't sure of the scope here. Haven't been able to get it to work using local packages, but it does seem to work with remote ones :ok_hand:

nonsense:
@dgrisham if you want to use local version of packages you could use go.mod replace directives and the extra sources feature added in https://github.com/testground/testground/pull/1085

We do that in the `lotus-soup` test plan at:
https://github.com/filecoin-project/oni/blob/master/lotus-soup/go.mod#L38
https://github.com/filecoin-project/oni/blob/master/lotus-soup/manifest.toml#L2

nonsense:
I just noticed that #1085 is actually merged to our current in-progress branch `oni` - we are working on pushing upstream to `master` all the new features in the coming weeks. You will have to build Testground from `oni` for now, or wait for Testground v0.5.3

dgrisham:
Ah awesome, I'll try that. Thanks @nonsense!

PR #1102, opened 2020-07-17
informative instance tags in pretty log printer.

raulk:
* In `local:exec`, we now identify instances in the pretty log output with: `<< group_id[idx] >>`
* In `local:docker`, we use the tag: `<< group_id[idx] (container_id[:6])`.

Since it's not obvious that composition-less runs via CLI are mapped to a singleton group named `single`, I've added a log statement for UX purposes.

This is what it looks like on `local:exec`:

image

---

Fixes https://github.com/filecoin-project/oni/issues/152.

---

cc @schomatis.


schomatis:
LGTM (the output, don't understand the code), thanks for addressing this so quickly.

PR #1103, opened 2020-07-17
fix: add ulimit memlock=-1

### LINKED ISSUE #701: Add ulimit to all testground docker builds
go 1.14 does not build on some kernels unless certain ulimits are applied.

See https://github.com/docker-library/golang/issues/320

I think we should apply the ulimit suggestions in our builds.

hacdias:
This closes #701 by adding `memlock=-1` to all ulimits across all `local:docker` runs and `docker:go` builds thus allowing go 1.14 to build on all kernels.

1. I added it in all runs and it **cannot** be overridden by flags. If you think it's best, I can update it to be the default, i.e., if the ulimit option is empty.
2. Should we add to `docker:generic` builds too in case people want to use Go?

License: MIT

### DIFF
--- a/pkg/runner/local_docker.go
+++ b/pkg/runner/local_docker.go
@@ -257,6 +257,8 @@ func (r *LocalDockerRunner) Run(ctx context.Context, input *api.RunInput, ow *rp
 				}},
 			}
 
+			// Updates the ulimit option to allow go 1.14 to build on all kernels.
+			// Ref: https://github.com/docker-library/golang/issues/320
 			if !strings.Contains(strings.Join(cfg.Ulimits, " "), "memlock=-1") {
 				cfg.Ulimits = append(cfg.Ulimits, "memlock=-1")
 			}
--- a/pkg/runner/local_docker.go
+++ b/pkg/runner/local_docker.go
@@ -257,13 +257,15 @@ func (r *LocalDockerRunner) Run(ctx context.Context, input *api.RunInput, ow *rp
 				}},
 			}
 
-			if len(cfg.Ulimits) > 0 {
-				ulimits, err := conv.ToUlimits(cfg.Ulimits)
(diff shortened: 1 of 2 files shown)


pkg/runner/local_docker.go
```diff
@@ -257,13 +257,17 @@ func (r *LocalDockerRunner) Run(ctx context.Context, input *api.RunInput, ow *rp
...
-			if len(cfg.Ulimits) > 0 {
-				ulimits, err := conv.ToUlimits(cfg.Ulimits)
-				if err == nil {
-					hcfg.Resources = container.Resources{Ulimits: ulimits}
-				} else {
-					ow.Warnf("invalid ulimit will be ignored %v", err)
-				}
+			// Updates the ulimit option to allow go 1.14 to build on all kernels.
+			// Ref: https://github.com/docker-library/golang/issues/320
+			if !strings.Contains(strings.Join(cfg.Ulimits, " "), "memlock=-1") {
```
hacdias, line 262: I don't like this. Suggestions are welcome.

nonsense:
I am not sure if we need this.

Has anyone complained that test plan containers are not working for them? I see the desire to support all kernels, but this is also adding complexity to Testground, for something that might break for some people, but at the same time noone has reported.

Furthermore this is introducing another change between `local:docker` and `cluster:k8s`.

hacdias:
@nonsense 

> Has anyone complained that test plan containers are not working for them?

Honestly, I don't know. I was just picking backlog issues. But I understand your point and I think it makes sense not to create more discrepancies between runners.

hacdias:
@nonsense should we close this? The linked issue (#701) is tagged as "Important but unplanned" and as per our previous discussion I think we can close this and the related issue.

nonsense:
Yes, I think we should close this.

PR #1106, opened 2020-07-20
feat: upgrade to latest sdk changes

hacdias:
This includes the upgrades made on the sdk. They ended up being less than I thought they would but it's good.

---

_You believe me it's faster to commit & wait for CI than building locally and waiting for Docker to download all dependencies? Bad connection..._


hacdias:
Updated this to be built on top of `oni` instead.

PR #1107, opened 2020-07-20
local:docker: add traffic routing policies

### LINKED ISSUE #1104: Allow test instances to signal that they want the sidecar to whitelist all traffic through the data plane
## Description

This is follow up to: https://github.com/testground/testground/pull/1082 (currently not merged to `master`, and only present in the `oni` branch, but at some point we will want to merge to `master`).

We want to make it possible for test instances to signal that they want the sidecar to whitelist all traffic through the data plane (except the control traffic, of course). Maybe add a field in `network.Config`? `WhitelistAll bool`?

At the moment test instances in `local:docker` have access on the data plane only over the data network, but don't have access to the internet as routes are blocked by default.

---

In the mentioned example, we need to have access from `local:docker` to the internet, so that Lotus processes have access to the Drand network.

hacdias:
- Requires https://github.com/testground/sdk-go/pull/27 ✅
- Closes #1104

For what I understood, we need a flag to disable the code commented on #1082. That flag is `WhitelistAll` and if true, the code is not run. If false, the code runs, removing all the existent routes that are not control routes.

I did not test this and I would like to see some validation from someone to check if this seems right. I might've not understood something right.

What I did:

1. Added the `WhitelistAll` flag to `network.Config` on the [sdk](ttps://github.com/testground/sdk-go/pull/27).
2. Added the `controlRoutes` to the `DockerNetwork` struct so they are available during `ConfigureNetwork`.
	- Also extracted this code into a `getControlRoutes` function.
3. If the flag is set to `false`, go over the all links and remove the routes that are not control routes.


pkg/sidecar/docker_network.go
```diff
@@ -138,5 +142,44 @@ func (dn *DockerNetwork) ConfigureNetwork(ctx context.Context, cfg *sdknw.Config
...
+					return fmt.Errorf("failed to remove existing route: %w", err)
+				}
+			}
+		}
+	}
+
 	return nil
 }
+
+func (dn *DockerNetwork) isControlRoute(route netlink.Route) bool {
```
hacdias, line 177: This function looks inefficient to me. Any ideas?

pkg/sidecar/docker_reactor.go
```diff
@@ -293,13 +254,60 @@ func (d *DockerReactor) handleContainer(ctx context.Context, container *docker.C
...
+			break
+		}
+		logging.S().Infow("successfully resolved route to host", "container_id", id)
+		controlRoutes = append(controlRoutes, hostRoutes...)
+	}
+
+	return controlRoutes, nil
+}
+
+func getNetworkHandlers(pid int) (netns.NsHandle, *netlink.Handle, error) {
```
hacdias, line 299: Not sure if this is the right file to put this function on.

hacdias:
@nonsense the tests are failing because of the RTT being slightly higher than it should. I don't think it's related to the changes I made, but maybe it is.

nonsense:
@hacdias I think tests are failing because packets go over a different interface and don't adhere to the traffic shaping:

```
Jul 20 14:51:58.259633	INFO	302.4608s       FAIL << single[001] (4a7322) >> expected an RTT between 200ms and 215ms, got 3.158262ms	{"req_id": "3d81051e"}
Jul 20 14:51:58.262355	INFO	302.4639s       FAIL << single[000] (827bad) >> expected an RTT between 200ms and 215ms, got 3.740121ms	{"req_id": "3d81051e"}
 ```

Basically we don't seem to be getting the 100ms delay at all (200ms roundtrip).

**Comment by hacdias** [comment]
🤦 I didn't notice the difference was so big. I'll try to understand what's going on then.

**Comment by hacdias** [comment]
So this ended up taking much more time than I thought 😅 finally made it pass the tests and I'm positive it works. This is what it does:

1. It stores the values from the non-control routes in the beginning.
2. Every time there's a change in the `WhitelistAll`, the routes are either added or removed from the IP configuration.

I also took the  opportunity to encapsulate two blocks of code in functions.

**Comment by nonsense** [comment]
@hacdias I see there are no additional tests, so it would be good to confirm that this is actually addressing the problem, if not in an automated test, then manually:

1. Could you run a test plan that fetches something from the internet (gist from Github or whatever), and confirm `WhitelistAll` works when set. (Maybe that could be an automated test, even if we have to rely on Github being up?)
2. Could you run a test plan where 1 fails, due to WhitelistAll not being set?
3. Could you run `splitbrain` test plan and make sure it continues to work? We don't have it as part of CI at the moment, so it would be good to confirm it is not being broken. (Maybe add it to CI if it is not too slow?).

WDYT?

File: pkg/sidecar/sidecar_handler.go [hunk 3]
```diff
@@ -24,8 +24,9 @@ func handler(ctx context.Context, instance *Instance) error {
 
 	// Network configuration loop.
 	err := instance.Network.ConfigureNetwork(ctx, &network.Config{
-		Network: defaultDataNetwork,
-		Enable:  true,
+		Network:      defaultDataNetwork,
+		Enable:       true,
+		WhitelistAll: true,
```
nonsense, line 29: I believe this is changing the behaviour we have until now: so far we did not have `WhitelistAll` for test plans (therefore I added the hack on `oni`, without an intention of it going to `master`).

I think we should set it by default to `false`, and have test plans be explicit when they want to `WhitelistAll`. This however makes it more difficult as routes are initially blocked, and only later allowed, but worth confirming.

hacdias:
@nonsense so I have both great and bad news:

> 1. Could you run a test plan that fetches something from the internet (gist from Github or whatever), and confirm `WhitelistAll` works when set. (Maybe that could be an automated test, even if we have to rely on Github being up?)

Done, it works.

> 2. Could you run a test plan where 1 fails, due to WhitelistAll not being set?

Done, it works. Added them both to CircleCI.

> 3. Could you run `splitbrain` test plan and make sure it continues to work? We don't have it as part of CI at the moment, so it would be good to confirm it is not being broken. (Maybe add it to CI if it is not too slow?).

This does not work. Something I changed broke this which I think is weird, but I'll try to understand what.

hacdias:
Note to self: I should also store the initial routes **per network**.

hacdias:
Update: I think I fixed everything. Won't be around for a few hours, but will update this afterwards. I have some questions and feedback to leave here.

hacdias:
@nonsense the updates I just pushed tick all the three points. I was removing/adding all the routes (even the ones from the "default" network). I missed completely that we were only removing the ones from [other networks](https://github.com/testground/testground/pull/1082/files#diff-dee5533a8419f2bdc79313bbed502441R279). Now the external traffic rules work.

My problem is: semantics. The `network.Config` struct has a `Name` and a `Policy`, making us think the policy is tightly related to the `Name`, but it's not! We're not changing the routes from the `default` network, we're changing the routes from external networks that provide **external** access.

The code now works but I'm not completely sure if this is the right way to tell sidecar this is what we want.

/cc @raulk

plans/network/traffic.go
```diff
@@ -0,0 +1,86 @@
...
+	"net/http"
+	"time"
+
+	"github.com/testground/sdk-go/network"
+	"github.com/testground/sdk-go/run"
+	"github.com/testground/sdk-go/runtime"
+	"github.com/testground/sdk-go/sync"
+)
+
+func makeTest(whitelistTraffic bool) run.TestCaseFn {
```
nonsense, line 17:
```suggestion
func routingPolicyTest(allowTraffic bool) run.TestCaseFn {
```

nonsense:
Overall LGTM. @hacdias we have to apply this to the `k8s_reactor` as well - probably worth spinning up a cluster and confirming the behaviour on Kubernetes with respect to the routing policies.

pkg/sidecar/docker_network.go
```diff
@@ -140,3 +147,18 @@ func (dn *DockerNetwork) ConfigureNetwork(ctx context.Context, cfg *sdknw.Config
...
+func (dn *DockerNetwork) handleRoutingPolicy(ctx context.Context, policy sdknw.RoutingPolicyType) error {
+	var err *multierror.Error
+
+	for _, routing := range dn.externalRouting {
+		switch policy {
+		case sdknw.AllowAll:
+			err = multierror.Append(err, routing.enable(dn.nl))
+		default:
+			err = multierror.Append(err, routing.disable(dn.nl))
+		}
```
nonsense, line 160: I think it'd be nice to be explicit about the policy, and not have a default case - maybe add `DenyAll`, and always make sure one of the two values is set. WDYT?

plans/network/traffic.go
nonsense: Maybe even better to just send the actual policy under test, rather than a `bool` that we then infer on.

pkg/sidecar/docker_network.go
hacdias: So I added the `DenyAll` rule. However, I left the default case because returning an error instead would make all other tests fail since none of them is setting the option. I think for now we should just let `DenyAll` as default as it was before to avoid breaking test cases compatibility.

hacdias:
> Overall LGTM. @hacdias we have to apply this to the k8s_reactor as well - probably worth spinning up a cluster and confirming the behaviour on Kubernetes with respect to the routing policies

@nonsense I'm merging this one into `oni` as soon as the CI turns green and I'll create a follow-up issue for k8s 👍

Issue #1112, opened 2020-07-24
add authentication functionality


nonsense:
### Authentication (must)
-   Introduce an Authenticator abstraction in the daemon. Initially backed by a simple toml-based file that maps user => token.
-   HTTP calls to the daemon should carry an `Authentication` HTTP request header with value `Token <token>`.
-   One special user will be the GitHub automation [R07].

### Authentication (nice-to-have)
-   We should be able to stamp each job with the user that requested it.

#### Usages
-   Client-side, the token should be set in the `.env.toml` file.
-   Initially, informational. Each job will be stamped with its requestor.

PR #1113, opened 2020-07-24
basic auth functionality

nonsense:
Partially addresses https://github.com/testground/testground/issues/1112

We will need something like this, once we spin up ingress for the daemon in the CI cluster.

---

FOLLOW UP TODO:
- [ ] maybe add another X-header that is `username` which will be attached to `jobs` in Testground, so that it is clear who started a given job.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -15,6 +15,7 @@ import (
 	"net/textproto"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 
 	"github.com/logrusorgru/aurora"
@@ -396,8 +397,10 @@ func (c *Client) request(ctx context.Context, method string, path string, body i
 	req, err := http.NewRequest(method, "http://"+c.endpoint+path, body)
 	req = req.WithContext(ctx)
 
-	// add auth token
-	req.Header.Add("X-Token", c.cfg.Client.Token)
+	token := strings.TrimSpace(c.cfg.Client.Token)
+	if token != "" {
+		req.Header.Add("Authorization", "Bearer "+token)
+	}
 
 	for i := 0; i < len(headers); i = i + 2 {
 		req.Header.Add(headers[i], headers[i+1])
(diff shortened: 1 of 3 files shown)


pkg/daemon/daemon.go
```diff
@@ -38,6 +38,18 @@ func New(cfg *config.EnvConfig) (srv *Daemon, err error) {
 
 	r := mux.NewRouter()
 
+	// Authentication
+	r.Use(func(next http.Handler) http.Handler {
```
raulk, line 42: We should add this handler only if there are tokens in the cfg.

pkg/daemon/daemon.go
```diff
@@ -38,6 +38,18 @@ func New(cfg *config.EnvConfig) (srv *Daemon, err error) {
...
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
raulk, line 43: We can initialise a map[string]struct{} inside the closure, and close over it (capture it) in the nested closure. Then we can get rid of the contains method. BTW -- when dealing with critical config values like this, I really recommend cleaning up the user-input strings using strings.TrimSpace.

pkg/client/client.go
```diff
@@ -396,6 +396,9 @@ func (c *Client) request(ctx context.Context, method string, path string, body i
 	req, err := http.NewRequest(method, "http://"+c.endpoint+path, body)
 	req = req.WithContext(ctx)
 
+	// add auth token
+	req.Header.Add("X-Token", c.cfg.Client.Token)
```
raulk, line 400: BTW, I'd recommend using the standard `Authorization: Bearer` convention: https://swagger.io/docs/specification/authentication/bearer-authentication/.

pkg/daemon/daemon.go
nonsense: Right, I added check for len, but inside the handler. You're right that it should be for the handler itself.

PR #1115, opened 2020-07-26
docker:go: replace rolling cache by static cache

### LINKED ISSUE #1088: rethink the rolling docker:go cache build
Currently if we set `enable_go_build_cache = true` build config for docker:go builder, we do a rolling build cache using `func (b *DockerGoBuilder) parseBuildCacheOutputImage(output string) string {`which looks for an intermediate image ID, and caches it, so that we don't have to rebuild Go source-code multiple times.

However overlay2 storage driver has a max depth of 125 layers, and by selecting an intermediate image ID, we can't really use the docker build `--squash` option.

We have to come up with a way to squash the cache image, so that we stop hitting the 125 layers limit.

hacdias:
This PR closes #1088 by removing the rolling cache mechanism and replacing it by a more manual cache process. Now, the cached image is only built once and the developer decides when to reset/clear the cached image.

To enable the cache, the option `enable_go_build_cache` must be set to true, just like it is done right now. When it's true, two things can happen:

1. If there's no cached image, we cache the intermediate later right before the runtime phase.
2. If there's a cached image, use it as a base.

To clear up the cache, you can run:

```
testground build purge --builder docker:go --plan <test-plan-name>
```

### DIFF
--- a/pkg/build/docker_go.go
+++ b/pkg/build/docker_go.go
@@ -111,6 +111,8 @@ type DockerGoBuilderConfig struct {
 	//    that assumption, we have found that the end-to-end build is faster.
 	EnableGoBuildCache bool `toml:"enable_go_build_cache"`
 
+	ResetGoBuildCache bool `toml:"reset_go_build_cache"`
+
 	// DockefileExtensions enables plans to inject custom Dockerfile directives.
 	DockerfileExtensions DockerfileExtensions `toml:"dockerfile_extensions"`
 }
@@ -230,7 +232,7 @@ func (b *DockerGoBuilder) Build(ctx context.Context, in *api.BuildInput, ow *rpc
 
 	var baseimage string
 	var alreadyCached bool
-	if cfg.EnableGoBuildCache {
+	if cfg.EnableGoBuildCache && !cfg.ResetGoBuildCache {
 		baseimage, err = b.resolveBuildCacheImage(ctx, cli, in, cfg, ow)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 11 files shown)


hacdias:
Ping @nonsense @raulk

nonsense:
@hacdias I think this part should be a command and not a builder configuration:

```
When reset_go_build_cache=true, the previously stored cached image is cleared and we start from a clear image, not from the the previous cache to avoid hitting the 125 layers limit of the storage driver.
```

Users should be able to clear the stored cached image with a command, rather than having to fiddle with the configs for their test plan. I also think we shouldn't have such an option in the config in the first place, rather just `EnableGoBuildCache`.

hacdias:
I completely agree. I was just worried that we would start creating command for specific actions for specific runners. Will do something like:

```
testground cache remove --runner <name> --testplan <name>
```

hacdias:
@nonsense just moved it to its own command. Since it's tightly coupled with the builder mechanism, I added it as a builder function. The command is like this:

```
testground build purge --builder docker:go --plan network
```

It can be extended to purge any additional resources of other builders. Perhaps `purge` is not the best name. Can rename if you have suggestions.

pkg/build/docker_go.go
```diff
@@ -228,12 +211,15 @@ func (b *DockerGoBuilder) Build(ctx context.Context, in *api.BuildInput, ow *rpc
 		args["RUNTIME_IMAGE"] = &cfg.RuntimeImage
 	}
 
+	cacheimage := fmt.Sprintf("tg-gobuildcache-%s", in.TestPlan)
 	var baseimage string
+	var alreadyCached bool
```
nonsense, line 216: I am not sure I understand the purpose of this variable... it seems that in the `if condition` below it is always set to `true`.

I fail to see how `cfg.EnableGoBuildCache && !alreadyCached` can happen.

nonsense:
@hacdias have you tested this? Can we update the description of the PR for future reference, it seems to be out-dated compared to the current impl.

pkg/build/docker_go.go
hacdias: If `EnableGoBuildCache` is set for the first time and there's no cached image, we cache it. The other command is just to remove the cache. By removing that condition, we'd be updating the cache image every time, making it a rolling cache.

hacdias:
@nonsense 

> have you tested this?

Yeah! It seems to work.

> Can we update the description of the PR for future reference, it seems to be out-dated compared to the current impl.

Sure, makes sense

hacdias:
Should behave correctly now

pkg/build/docker_go.go
```diff
@@ -436,54 +424,40 @@ func (b *DockerGoBuilder) setupGoProxy(ctx context.Context, ow *rpc.OutputWriter
...
+		return true, nil
 	}
 
-	ow.Infow("tagging initial go build cache image", "cache_image", cacheimage, "base_image", baseimage)
-
-	if err := cli.ImageTag(ctx, baseimage, cacheimage); err != nil {
-		return "", fmt.Errorf("failed to tag %s as %s", baseimage, cacheimage)
-	}
-	return cacheimage, nil
+	ow.Info("build cache image not found", "cache_image", cacheImage)
```
nonsense, line 442:
```suggestion
	ow.Infow("build cache image not found", "cache_image", cacheImage)
```

PR #1125, opened 2020-08-07
feat: supervisor and async tasks

### LINKED ISSUE #1022: runners: support long-running tests
Right now some runners have explicit cancellations following a timeout (20min - https://github.com/testground/testground/blob/master/pkg/daemon/daemon.go#L58 ; as well as https://github.com/testground/testground/blob/master/pkg/runner/cluster_k8s.go#L573). Some users want to be able to run long-running tests with Testground.

### LINKED ISSUE #1043: introduce concept of jobs to Testground

### LINKED ISSUE #1044: introduce a supervisor component

hacdias:
This PR closes a multitude of other PRs by implementing the Supervisor and async tasks described on #1044, #1043 and #1022. Each task is described bellow.

- Closes #1059.
- Closes #1044.
- Closes #1043.
- Closes #1022.

## The Big Todo

- [x] Keeps state of `Scheduled` and `InProgress` tasks.    
- [x] Moves tasks to `Archive` store when they get to a `Completed` state.
- [x]  Runs `n` tasks concurrently (we start with `n=1`)
	- [x] `n` is customizable (const or var in .env.toml)
- [x] Tracks and supervises test runs and all test instances
- [ ] Should be upgradable in between test runs (draining on OS signal)
- [x] Doesn’t determine whether a test run is `Successful` or `Failed`, just `Completed`, `Scheduled`, or `InProgress`
- [x] As a developer, when I submit a job (run or build) to Testground, I should:
	- [x] Get back a unique <JOB_ID> from Testground so that I can query job status while it is running.
	- [x] Get back an error if there are too many jobs scheduled.
- [x] Add `--wait` flag to `testground run` and `testground build`, which will now:
	- [x] Submit a job and get back a unique <JOB_ID>
	- [x] Busy-wait until the job reaches the Completed state
	- [x] On context cancel, the job is canceled.
- [x] Add `testground status -t <id>` command that would query and return the current state of a job.
- [x] Add `testground tasks` (right now shows all tasks of the past 24 hours, can be further customized in a different PR)
- [x] Add `testground logs -t <id> [-f]`
- [x] Persist tasks to disk
- [ ] Documentation on variant `master` at https://docs.testground.ai/

## Nice to have

- [ ] Should be able to survive restart, ideally even if there are `InProgress` tasks


pkg/api/rpc.go
```diff
@@ -58,3 +64,14 @@ type CollectResponse struct {
...
+type TaskStatusResponse struct {
+	Priority   int
+	ID         string
+	Type       string
+	Input      interface{}
+	Result     task.TaskResult
+	Created    string
+	LastUpdate string
+	LastState  string
+}
```
hacdias, line 77: I would prefer to directly send `task.Task` but there's an issue when mapstructure decodes `time.Time` which can be [easily fixed](https://github.com/mitchellh/mapstructure/issues/159). But there's another, more complicated issue, to decode enums (for `TaskType` and `TaskState`) which I don't know how to solve yet. We could use reflection, but reflecting the enum type will only give me `string` which is not very helpful.

hacdias:
@nonsense I implemented most of the supervisor with a concurrency of 2 tasks. The queue has a limit of 100 tasks at the time. I mostly used the task and queue implementation that was there already. Could you take a quick look at the code, please?

I have implemented a rough version of `--wait`: basically just created an endpoint that wait for the task to finish and then replies with the data. However, I was wondering if it wouldn't be better to print the output of the task on the client. I'm just unsure on how to do that. That and how to connect the cancellation between ctrl-c and a running task (maybe store the `cancel` function somewhere accessible?). What do you think?

pkg/cmd/status.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	defer r.Close()
+
+	res, err := client.ParseTaskStatusResponse(r)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("ID:\t\t%s\n", res.ID)
+	fmt.Printf("Created:\t%s\n", res.Created)
```
nonsense, line 49: This value appears to be set wrong. When I submit a job, I get the following status back:

```
ID:             b513cdd1-0153-4b6b-be87-4ddbbbd0c08f
Created:        4189-12-26 04:53:15.7134801 +0000 UTC
Type:           run
Status:         processing
Last update:    2020-08-17 13:50:47.2888156 +0000 UTC
```

pkg/engine/engine.go
```diff
@@ -134,270 +144,64 @@ func (e *Engine) ListRunners() map[string]api.Runner {
...
-	}
-
-	// Wait until all goroutines are done. If any failed, return the error.
-	if err := errgrp.Wait(); err != nil {
-		return nil, err
-	}
-
-	return ress, nil
+func (e *Engine) QueueBuild(request *api.BuildRequest, sources *api.UnpackedSources) (string, error) {
+	id := uuid.New().String()
```
nonsense, line 148: I think it'd be best to stay with `uuid.New()[:8]` because these IDs are part of image names and we quickly hit docker tag name limits if we use a full-fledged UUID (36 chars). I think 8 chars random UUID is fine for Testground for the foreseeable future.

nonsense:
One potential problem with the current impl. is that we are losing context in terms of the execution log from the `testground daemon` when queuing jobs and waiting for them to be processed. It should be possible to query and get the current execution log for a given job, for example with `testground logs <RUN_ID>`. As a user only knowing that my job is `scheduled`, or `processing` is not enough I think.

pkg/task/task.go
```diff
@@ -3,29 +3,29 @@ package task
...
 import "time"
 
-// TaskState (kind: int) represents the last known state of a task.
+// TaskState (kind: string) represents the last known state of a task.
 // A task can be in one of three states
 // StateScheduled: this is the initial state of the task when it enters the queue.
 // StateProcessing: once work begins on the task, it is put into this state.
 // StateComplete: work is no longer being done on this task. client should check task result.
-type TaskState int
+type TaskState string
```
nonsense, line 11: Nitpick: this is already in the `task` package, maybe just rename to `State`? No need to `task.TaskState`.

pkg/task/task.go
```diff
@@ -56,3 +61,11 @@ func (t *Task) Created() time.Time {
 	}
 	return time.Unix(u.Time().UnixTime())
 }
+
+func (t *Task) LastState() DatedTaskState {
```
nonsense, line 65: This could be named just `State()` - I think it is clear that this is the current/latest state.

pkg/api/rpc.go
```diff
@@ -43,6 +47,11 @@ type BuildPurgeRequest struct {
 	Testplan string `json:"testplan"`
 }
 
+type TaskStatusRequest struct {
+	ID                string `json:"id"`
+	WaitForCompletion bool   `json:"wait-for-completion"`
```
nonsense, line 52: Seems like you are using both camelCase and other naming scheme - I suggest we `snake_case` everywhere - I think this is what Testground is already using (i.e. "run_config", "build_config", "run_id", etc.)

pkg/api/rpc.go
nonsense: Yeah, that struct seems weird. Why not just define a JSON Marshal/Unmarshal methods, rather than store time and state as strings? Also maybe rename `LastUpdate` to `Updated`, and `LastState` to `State`.

pkg/engine/supervisor.go
```diff
@@ -0,0 +1,381 @@
...
+	)
+
+	for {
+		tsk, err := queue.Pop()
+		if err == task.ErrQueueEmpty {
+			time.Sleep(time.Second)
+			continue
+		}
+
+		ctx, _ := context.WithTimeout(context.Background(), time.Minute*30)
```
nonsense, line 52: I think we should `cancel` this context if the task completes successfully, before next iteration.

pkg/engine/supervisor.go
```diff
@@ -0,0 +1,381 @@
...
+	comp, err := input.Composition.PrepareForBuild(&input.Manifest)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := comp.ValidateForBuild(); err != nil {
+		return nil, fmt.Errorf("invalid composition: %w", err)
+	}
+
+	// TODO: how to pass this to the client with --wait? Should we?
```
nonsense, line 90: Yes, we should. We should also persist the execution trace for inspection with `testground logs`.

hacdias:
This PR is almost done. There are two issues:

1. `storage.Range` is not working for the time ranges I'm giving it. So I'm not being able to finish the `testground tasks` command yet.
2. Some Go tests seem to fail on CI randomly. I think it may be related to the way I'm reading the output file for streaming via the log command?

hacdias:
> storage.Range is not working for the time ranges I'm giving it. So I'm not being able to finish the testground tasks command yet.

Already know the answer: we're relying on IDs being formatted as UUIDs so we can get the time easily. But that's not how we're storing them actually so the time ranges do not work. Gotta find a new way.

hacdias:
I just started using [xid](https://github.com/rs/xid) library instead because the IDs are shorter than UUIDs and still encode the timestamp, which is necessary for the ranges.

hacdias:
The tests are running on my computer without any issue. On CI they are failing with:

```
invalid character 'P' looking for beginning of value
```

By restarting the tests many times, we can verify that it can happen on virtually any test that calls the CLI `testgroud run` or `testground build`. My suspicion is that this is related to the server side logs logic and the way I'm fetching the logs as they're still being written to the file system.

hacdias:
OMG. Just fixed!

hacdias:
So there's something going on with the tests that I don't know yet what it is. It seems that using the permanent storage is making something break during the integration tests for k8s which doesn't seem to make much sense. Needs investigation.

The permanent storage for tasks is required so I can finish the last todo on this PR: "Should be upgradable in between test runs (draining on OS signal)".

hacdias:
@nonsense did you manage to take a look? I'm not understanding what's causing the error during the k8s test because it works during the Docker test! It doesn't seem to make much sense.

nonsense:
@hacdias I added a few changes to the PR, so that it makes debugging the failing tests easier:
1. More powerful machine, so that tests are faster.
2. Waiting for `testground daemon` to start listening on 8080 (in case of races).
3. Added `set -x` so that we print out the temporary dir where we save the output of the testground daemon, and we can debug easily when we attach with SSH to the CircleCI machine.

---

It seems that the `testground daemon` is panicking on occasion with the following error, which results in the tests deadlocking (as we can't connect to the daemon, hence the `connection refused` message):

```
circleci@default-d190a968-3291-425e-ba2b-ed9fb2cfbf50:/tmp/tmp.fs4Qe4GlOw$ cat daemon.out
Sep 11 11:50:03.976989  INFO    using home directory: /home/circleci/testground
Sep 11 11:50:03.977361  INFO    .env.toml loaded from: /home/circleci/testground/.env.toml
Sep 11 11:50:03.990389  INFO    supervisor worker started       {"worker_id": 0}
Sep 11 11:50:03.990459  INFO    supervisor worker started       {"worker_id": 1}
Sep 11 11:50:03.991295  INFO    listen and serve        {"addr": "[::]:8080"}
Sep 11 11:50:03.991371  INFO    daemon listening        {"addr": "[::]:8080"}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x15cf864]

goroutine 15 [running]:
github.com/testground/testground/pkg/engine.(*Engine).worker.func1(0xc0001bf500, 0x0, 0xc0005bbf90, 0x0)
        /home/circleci/project/pkg/engine/supervisor.go:59 +0xd4
github.com/testground/testground/pkg/engine.(*Engine).worker(0xc0001bf500, 0x0)
        /home/circleci/project/pkg/engine/supervisor.go:105 +0x147
created by github.com/testground/testground/pkg/engine.NewEngine
        /home/circleci/project/pkg/engine/engine.go:110 +0x37f
```

hacdias:
@nonsense it seems there was a missing error verification. I checked at the logs and couldnn't see the error printed anywhere. I'm going to restart the tests 2 or 3 times to see if it happens.

hacdias:
For "Should be upgradable in between test runs (draining on OS signal)" I'm thinking about the following:

1. Adding a signal handler inside the engine;
2. On receiving the signal, waits for the current tasks to complete; shuts down.
3. On turning on, checks the storage for queued tasks, and runs them.

Sounds good?

nonsense (approved):
I think this is a great first step! Thanks a lot @hacdias .

I tested most/all of the new features and they work very well.

Let's discuss how to continue from here!

nonsense:
> For "Should be upgradable in between test runs (draining on OS signal)" I'm thinking about the following:

> 1. Adding a signal handler inside the engine;
> 2. On receiving the signal, waits for the current tasks to complete; shuts down.
> 3. On turning on, checks the storage for queued tasks, and runs them.

> Sounds good?

---

For 2. I think we should have a deadline
- some tasks will complete in ~5-6min.
- some might be broken and die only on timeout (which is quite long right now, I think hours)...
- some tasks might be legit, but take 5-6 hours (such as Drand outage tests on Lotus).

As an operator, I might be okay to have a deadline of 5-10min, but I wouldn't want to wait longer for current tasks to finish executing (or timing out), so if they don't complete within the deadline, just cancel them and shut down.

pkg/cmd/run.go
```diff
@@ -138,61 +139,66 @@ func doRun(c *cli.Context, comp *api.Composition) (err error) {
...
-	_, manifest, err = resolveTestPlan(cfg, comp.Global.Plan)
+	planDir, manifest, err := resolveTestPlan(cfg, comp.Global.Plan)
 	if err != nil {
 		return fmt.Errorf("failed to resolve test plan: %w", err)
 	}
 
-	// Check if we have any groups without an build artifact; if so, trigger a
-	// build for those.
-	var buildIdx []int
+	// Check if this the daemon will need to build the project.
```
nonsense, line 147:
```suggestion
	// Check if the daemon needs to build the test plan.
```

Issue #1130, opened 2020-08-19
Error using `local:docker` and `docker:go` in Linux.


adlrocha:
#### Version information:
`commit 75d9c9a681afc0e8564d9388c0c7d4cb51216388`
x86_64 GNU/Linux 5.4.0-7634-generic

#### Description:
I am trying to run a single testcase using `local:docker` and `docker:go` and I am unable to start  the test. If I run the quickstart testcase [from the docs](https://docs.testground.ai/v/master/writing-test-plans/quickstart) everything works fine and the testcase runs successfully.
```
testground run single --plan=quickstart \
                                  --testcase=quickstart \
                                  --runner=local:docker \
                                  --builder=docker:go \
                                  --instances=2
```
With the following output:
```
Aug 19 07:15:51.351875  INFO    1.1849s    MESSAGE << eb6967be7f0d >> Hello, Testground!        {"req_id": "e125f502"}
Aug 19 07:15:51.352008  INFO    1.1866s    MESSAGE << a5d3f048b22c >> Hello, Testground!        {"req_id": "e125f502"}
Aug 19 07:15:51.352126  INFO    1.1849s         OK << eb6967be7f0d >>   {"req_id": "e125f502"}
Aug 19 07:15:51.352252  INFO    1.1867s         OK << a5d3f048b22c >>   {"req_id": "e125f502"}
Aug 19 07:15:51.352359  INFO    1.1892s    MESSAGE << eb6967be7f0d >> waiting for pushgateway to become accessible      {"req_id": "e125f502"}
Aug 19 07:15:51.352694  INFO    1.1894s    MESSAGE << a5d3f048b22c >> waiting for pushgateway to become accessible      {"req_id": "e125f502"}
Aug 19 07:15:51.353481  INFO    1.1930s    MESSAGE << a5d3f048b22c >> io closed {"req_id": "e125f502"}
Aug 19 07:15:51.353823  INFO    1.1930s    MESSAGE << eb6967be7f0d >> io closed {"req_id": "e125f502"}
Aug 19 07:16:31.365358  INFO    41.2047s    MESSAGE << eb6967be7f0d >> metrics done     {"req_id": "e125f502"}
Aug 19 07:16:51.368113  INFO    61.2075s    MESSAGE << a5d3f048b22c >> metrics done     {"req_id": "e125f502"}
[...]

PR #1132, opened 2020-08-19
add flag to control cgo

### LINKED ISSUE #1131: add `enable_cgo` option in `docker:go` builder
To fix #1130.

---

Dropping `CGO_ENABLED=0` by default in the `docker:go` builder, commit https://github.com/testground/testground/commit/950238a9f41ed184eddeabf9876bb5a681beee0a, caused a regression.

Dynamic linking is necessary for the Lotus tests, but it forces developers to use custom images that contain the dynamically-linked libraries. This is in turn inconvenient for testing things like IPFS.

We should add an option `DockerGoBuilderConfig#EnableCGO` that defaults to false (i.e. implicit boolean default).

* When false, we pass `CGO_ENABLED=0` (default).
* When true, we pass `CGO_ENABLED=1` (we'll need to update the Oni test plans to set this flag).

nonsense:
This PR is adding a flag to control cgo.

Note that it is also changing the default behaviour in Testground when using the `docker:go` builder, as by default now Cgo will be disabled. See https://github.com/testground/testground/commit/950238a9f41ed184eddeabf9876bb5a681beee0a#r41587533

---

cc @adlrocha 

---

Fixes #1131 #1130.

### DIFF
--- a/pkg/build/docker_go.go
+++ b/pkg/build/docker_go.go
@@ -95,6 +95,13 @@ type DockerGoBuilderConfig struct {
 	EnableGoBuildCache bool `toml:"enable_go_build_cache"`
 
 	// Cgo enables the creation of Go packages that call C code. By default it is disabled.
+	// Enabling CGO also enables dynamic linking. Disabling CGO (default) produces statically
+	// linked binaries.
+	//
+	// If you ever see errors like the following, you are probably better off
+	// disabling CGO (and therefore enabling static linking).
+	//
+	//   /testplan: error while loading shared libraries: libdl.so.2: cannot open shared object file: No such file or directory
 	// If you pass `true` to this flag, your test plan will be built with CGO_ENABLED=1
 	EnableCGO bool `toml:"enable_cgo"`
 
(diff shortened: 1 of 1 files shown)


pkg/build/docker_go.go
```diff
@@ -94,6 +94,10 @@ type DockerGoBuilderConfig struct {
 	// cached image.
 	EnableGoBuildCache bool `toml:"enable_go_build_cache"`
 
+	// Cgo enables the creation of Go packages that call C code. By default it is disabled.
```
raulk, line 97: Leaving a note to ourselves, because I know we're probably going to trip on this one again.

```suggestion
	// Cgo enables the creation of Go packages that call C code. By default it is disabled.
	// Enabling CGO also enables dynamic linking. Disabling CGO (default) produces statically
	// linked binaries.
	//
	// If you ever see errors like the following, you are probably better off
	// disabling CGO (and therefore enabling static linking).
	//
	//   /testplan: error while loading shared libraries: libdl.so.2: cannot open shared object file: No such file or directory
```

raulk (approved):

PR #1140, opened 2020-09-17
initial notification to slack on completed task ; basic html pages for tasks and logs

nonsense:
This PR is:
1. Adding `SlackWebhookURL` to which we publish a message when a task is completed. We don't do anything if this field is empty.
2. Adds 3 GET endpoints:
```
/tasks - lists all tasks
/logs - used to fetch logs for a given task
/outputs - returns a download of the outputs tgz for a given task
```

TODO:
- [x] make slack notifications optional
- [x] add testplan/testcase in the message
- [x] fix `outputs` link on `tasks` html page
- [x] fix tests

### DIFF
--- a/pkg/task/task.go
+++ b/pkg/task/task.go
@@ -67,6 +67,10 @@ func (t *Task) Name() string {
 	return t.Plan + ":" + t.Case
 }
 
+func (t *Task) Took() time.Duration {
+	return t.State().Created.Sub(t.Created()).Truncate(time.Second)
+}
+
 func (t *Task) State() DatedState {
 	if len(t.States) == 0 {
 		panic("task must have a state")
--- a/pkg/task/task.go
+++ b/pkg/task/task.go
@@ -63,6 +63,10 @@ func (t *Task) Created() time.Time {
 	return t.States[0].Created
 }
 
+func (t *Task) Name() string {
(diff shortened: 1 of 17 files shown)


pkg/task/task.go
```diff
@@ -45,8 +45,10 @@ type Result struct {
 // state of a running or scheduled task.
 type Task struct {
 	Version  int          `json:"version"`  // Schema version
-	Priority int          `json:"priority"` // scheduling priority
+	Priority int          `json:"priority"` // Scheduling priority
 	ID       string       `json:"id"`       // Unique identifier for this task
+	Plan     string       `json:"plan"`     // Test plan
+	Case     string       `json:"case"`     // Test case
```
nonsense, line 51: Maybe these two would move to a `Metadata` struct, so that we include more relevant metadata (such as `CreatedBy`, `Commit`, `Branch`, `Repo`, `Dependencies`, etc.

I imagine some of these won't be filled by all users, but if they are, it would be nice to be able to filter based on them...

hacdias (approved):

pkg/task/task.go
hacdias: I think that'd be a great idea and give us more flexibility in the future.

PR #1152, opened 2020-09-22
prepend pod name to logs

nonsense:
This PR is an attempt to include `stdout` from test plan instances (such as logs emitted by various loggers in Lotus), to the logs returned to the testground client, so that users can access them when using TaaS.

---

We've had a conversation about propagating `stdout` and `stderr` from test plan instances back to Testground and for them to be available to developers. We still haven't decided if we want to do that, so this is a hack that could work for small-sized test plans.

Ultimately test plans should be good citizens and using the run environment provided by Testground SDK. However problems are possible when test plans have to use libraries that are not configurable (for example with respect to loggers) and emit valuable information to stdout/stderr.

---

Related: https://github.com/filecoin-project/oni/issues/73


pkg/runner/cluster_k8s.go
```diff
@@ -341,7 +342,7 @@ func (c *ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow *rpc
 						return err
 					}
 
-					fmt.Print(logs)
+					ow.WriteProgress([]byte(logs))
```
nonsense, line 345: We know this won't scale well for a large number of instances, but for the purposes of the Lotus tests it will be fine.

pkg/runner/cluster_k8s.go
```diff
@@ -341,8 +342,8 @@ func (c *ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow *rpc
 						return err
 					}
 
-					fmt.Print(logs)
```
nonsense, line 344: Previously we were just printing the last 2 log lines from every container on the daemon process, as a shortcut.

Obviously this doesn't work anymore, as we are generally interested in all logs, and also want to propagate them back to the user.

Issue #1153, opened 2020-09-24
logs are not always flushed to output file


nonsense:
When running test plans on testground as a service, logs are saved in a file within the testground daemon datadir.

Logs are supposed to be a replication of what is returned to the daemon with `ow *rpc.OutputWriter` (https://github.com/testground/testground/blob/master/pkg/runner/cluster_k8s.go#L151), but is not the case.

This is reproducible with the `deals-stress-concurrent-natural-k8s.toml` composition from Oni, or just by triggering an hour build on the Oni repo.

Issue #1147, opened 2020-09-18
testground task status improvements


nonsense:
At the moment we only report if a Testground task is `scheduled`, `processing` or `completed`.

- [ ] We should include an **exitcode** for the task run in the task schema and display it on the dashboard
- [ ] When we get a failed run (see definition of fail below), testground should fetch extra information such as:
a) number of `outcome: ok` events / vs total instances per composition group
b) testground should inspect runner env events, for example in `cluster:k8s` - `ErrImagePull`, or `OOM`, or instances blocked in `Pending` state (which would result in global timeout for the run) and attach them to the task in case of failures.

---

Definition of `exitcode`:
1. If a task failed to compile, the exitcode should be >0 and task should be marked as failed.
2. If a task compiles and its run completes, and all instances are succeeded and all logs include `outcome: ok` event, then the run is successful.
3. For now, every other case, means a run failed.

---

Note that `partial successes` are not in scope for this iteration. In the future we might want to split the `failed` runs into more categories, and include `partial success` (for example a test plan with 1000 instances, which are mostly independent and 999 manage to succeed with `outcome: ok` event). This is why it is important to also include extra information as part of runs that are not 100% successful.

PR #1154, opened 2020-09-24
add result field and propagate and store errors in tasks

nonsense:
This PR is:
1. Adding `Result` field in run outputs and starting to use it as part of `Task` - Result is an interface and depends on the runner that produces it, it should contain all information that is relevant to a given testplan run executed on a given runner (i.e. not just status, and exit codes for test instances, but also enough debug information for developers to understand what happened in their run).
2. Adding `Error` field as part of `Task` - this is an error returned by Testground. You can think of it as a higher level than `Result`.. could be "global timeout", "not enough resources on the cluster", "build failed", etc.
3. `cluster:k8s` - Subscribing to all WARN events and pod statuses relevant to a given run, while executing it. Also collecting these events and storing them inside `Result` for reference for developers.
4. Fixing bugs with respect to error propagation.
5. Fixing double-JSON marshalling for logs.
6. Fixing long line log parsing due to `bufio.ReadBytes`.
7. Functionality to post updates to Slack and Github
8. Slight refactor on the `Task` model and separation of concerns (DB operations, vs business logic, etc.)
9. Adding `StateCanceled` to task states
10. Extracting timeouts - currently we have `run` and `task` timeouts on `cluster:k8s`, with the idea to unify them into one timeout in the future.
...

---

TODO:
- [x] solve https://github.com/testground/testground/issues/1153
- [x] make sure `outcome:ok` is taken into account when deciding if status is ok/fail (i.e. support strict case fully for now) ; leniency mode should be added as parameter to manifest.toml
- [x] make global timeout configurable
- [x] introduce some schema for run journal (events from k8s + pod statuses)
- [x] dedup journal entries
- [x] fix tests
- [x] godoc
- [x] refactor post to github

---

FOLLOW UP:
- [ ] consider removing `run_timeout` in `cluster:k8s` runner, and relying only on `task_timeout`
- [ ] gracefully handle testground daemon shutdown without leaving tasks in store (we wouldn't need `delete` task func)
- [ ] fix journaled Kubernetes statuses in Result
- [ ] support multiple github repos for status check updates
- [ ] some kind of ACL maybe - userA should be able to kill userB's task?
- [ ] admin control panel (for tasks such as `delete`)
- [ ] fix `took` on dashboard at https://ci.testground.ipfs.team/ to not include `scheduled` time, but only `processing` time.
- [ ] remove `plan`, `case`, since we already have a `Composition` in `Task`

---

Fixes: https://github.com/testground/testground/issues/1147
Fixes: https://github.com/testground/testground/issues/1153
Fixes: #1163


pkg/client/client.go
```diff
@@ -505,21 +505,6 @@ func ParseLogsRequest(w io.Writer, r io.ReadCloser) (api.LogsResponse, error) {
 				return err
 			}
 
-			var data struct {
```
nonsense, line 508: @hacdias removing this, as I don't think we need to double-encode log lines.

pkg/daemon/tasks.go
```diff
@@ -53,12 +55,64 @@ func (d *Daemon) listTasksHandler(engine api.Engine) func(w http.ResponseWriter,
...
 	}
 }
+
+func (d *Daemon) redirect() func(w http.ResponseWriter, r *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		http.Redirect(w, r, "/tasks", 301)
+	}
+}
+
+func ByteCountSI(b int64) string {
```
nonsense, line 97: @raulk I remember you mentioned a library that you used for converting bytes to human-readable strings, which one is it?

nonsense:
@hacdias @raulk opening this for review, even though there are still a few TODOs, as it is getting quite large, and already addressing a few bugs that it'd be nice to merge to master.

pkg/engine/engine.go
```diff
@@ -420,23 +426,16 @@ func (e *Engine) Logs(ctx context.Context, id string, follow bool, cancel bool,
 		}
 		defer file.Close()
 
-		scanner := bufio.NewScanner(file)
```
nonsense, line 423: @hacdias no need to read the file line by line (which already includes `ndjson` and encoding once again further down with `ow.WriteProgress`.

pkg/engine/engine.go
```diff
@@ -420,23 +426,16 @@ func (e *Engine) Logs(ctx context.Context, id string, follow bool, cancel bool,
...
-
-		for scanner.Scan() {
-			_, err = ow.WriteProgress(scanner.Bytes())
-			if err != nil {
-				return nil, err
-			}
-
-			if err := scanner.Err(); err != nil {
```
nonsense, line 431: Note that this should have been outside of the `for scanner.Scan()` func.

pkg/engine/engine.go
```diff
@@ -450,14 +449,31 @@ func (e *Engine) Logs(ctx context.Context, id string, follow bool, cancel bool,
 		}
 	}
 
-	file, err := os.Open(path)
+	stop := make(chan struct{})
+	file, err := newTailReader(path, stop)
```
nonsense, line 453: @hacdias @raulk if you have better ideas on how to mimic `tail -f` behaviour, I'd be happy to refactor this.

The idea here is that we continuously read the file (which includes logs for a given task), and we sleep a bit if we encounter `io.EOF`.

`engine.Logs` is blocking, and `e.signals[id]` governs when we return from this function - this is used for the `--wait` flag, as well for the `testground logs --follow` command.

pkg/runner/cluster_k8s.go
```diff
@@ -216,7 +246,43 @@ func (c *ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow *rpc
...
+		var err error
+		err = c.monitorTestplanRunState(ctx, ow, input, &journal)
+
+		lastId := "0"
+
+		for {
+			notifications, newId, errn := notification.Fetch(&template, lastId)
+			if errn != nil {
+				return errn
+			}
```
nonsense, line 258: This is a major change. While previously we just returned from `monitorTestplanRunState` which waits for pods to complete (either in `Success` state or in `Failed`, doesn't matter), we now just wait for the test plan processes to exit, and then determine the outcome based on the `notifications`: currently we expect all test plan instances to report `{"outcome": "ok"}` event, and if they don't, we have a failure.

pkg/runner/cluster_k8s.go
```diff
@@ -318,46 +391,66 @@ func (c *ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow *rpc
...
+						ow.Debugw("fetching logs", "pod", podName)
+						logs, err := c.getPodLogs(ow, podName)
+						if err != nil {
+							return err
+						}
+						ow.Debugw("got logs", "pod", podName, "len", len(logs))
 
-					logs, err := c.getPodLogs(ow, podName)
-					if err != nil {
+						_, err = ow.WriteProgress([]byte(logs))
```
nonsense, line 417: For now we forward `stdout` logs from test plan instances to the `logs` output for given task. If test plans are well-implemented, in the future we might not need to do that.

pkg/daemon/tasks.go
```diff
@@ -53,12 +55,64 @@ func (d *Daemon) listTasksHandler(engine api.Engine) func(w http.ResponseWriter,
 			return
 		}
 
+		cr, _ := engine.RunnerByName("cluster:k8s")
```
nonsense, line 58: This whole handler will need to handle the `view` of the tasks (i.e. the tasks dashboard) properly soon, but for now I just printout an HTML table. It is out of scope for this PR to address the UI properly.

pkg/daemon/tasks.go
hacdias: I'm no Raul, but this came to mind: https://github.com/dustin/go-humanize

hacdias (approved):
After going over the 21 changed files, it looks good. Didn't spot any glaring flaw! I didn't try running the code though.

nonsense:
@raulk @hacdias this is ready for review again (having merged another set of changes in order not to have 2 huge PRs).

@hacdias you have reviewed most of this already, the changes until your previous review are:
```
git diff dfc34b642afa232a4acbbde7c02df7eb7b18d4a3..HEAD
```

hacdias:
Quickly went through it and looks good! Just to confirm, what's the difference between `run_timeout_limit` and `task_timeout_limit`? Is the first only for k8s _runs_, while the second works for everything? What do we need a specific one for k8s?

nonsense:
@hacdias the timeout on the `cluster:k8s` side has been there since the initial implementation, and is measured since the test plan run is scheduled. The task timeout includes both run AND build processes.

Ultimately we can have only one timeout that unifies build and run processes (i.e. the task timeout) and not have two timeouts, but I didn't want to introduce even more changes in this PR. This has been added as a follow up.

pkg/engine/engine.go
```diff
@@ -450,14 +444,32 @@ func (e *Engine) Logs(ctx context.Context, id string, follow bool, cancel bool,
...
+			e.signalsLk.RUnlock()
+			if !running {
+				time.Sleep(2 * time.Second)
+				close(stop)
+				return
+			}
+
+			time.Sleep(1 * time.Second)
+		}
+	}()
```
raulk, line 467: This might be racy, and I think we can get rid of the polling. Can't we get the signal channel outside the goroutine, and then receive from it inside the goroutine?

pkg/runner/cluster_k8s.go
raulk: With proper run events in the sync service now, we should stop doing this. The fact that the runner behaves different with different pod counts is confusing. We should also get rid of the tailing.

pkg/engine/engine.go
raulk: I think we should get rid of the tailing entirely. The way I see it:
1. we have run events in the sync service to update the user view.
2. we should consume those run events live, and update the UI accordingly. IMO that's what the new "tail" should be.
3. results aren't really actionable/useful until the run has finished, so those should be downloaded as outputs.

plans/network/pingpong.go
```diff
@@ -64,8 +57,9 @@ func pingpong(runenv *runtime.RunEnv) error {
 	ipC := byte((seq >> 8) + 1)
 	ipD := byte(seq)
 
-	config.IPv4 = &runenv.TestSubnet.IPNet
+	config.IPv4 = runenv.TestSubnet
 	config.IPv4.IP = append(config.IPv4.IP[0:2:2], ipC, ipD)
+	config.IPv4.Mask = []byte{255, 255, 255, 255}
```
raulk, line 62: Is this correct?

plans/example/go.mod
```diff
@@ -2,4 +2,16 @@ module github.com/ipfs/testround/plans/example
...
+	github.com/pkg/errors v0.9.1 // indirect
+	github.com/testground/sdk-go v0.2.6-0.20201013134246-8d8414e32da5
+	golang.org/x/net v0.0.0-20200301022130-244492dfa37a // indirect
+	golang.org/x/sync v0.0.0-20200317015054-43a5402ce75a // indirect
+	golang.org/x/sys v0.0.0-20200625212154-ddb9806d33ae // indirect
+	golang.org/x/tools v0.0.0-20191125144606-a911d9008d1f // indirect
+	google.golang.org/protobuf v1.25.0 // indirect
+	gopkg.in/check.v1 v1.0.0-20200227125254-8fa46927fb4f // indirect
+	gopkg.in/yaml.v2 v2.2.8 // indirect
+)
```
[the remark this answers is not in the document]
raulk: (Same for all other plans.)

plans/splitbrain/go.mod
```diff
@@ -2,4 +2,4 @@ module github.com/testground/testground/plans/splitbrain
 
 go 1.14
 
-require github.com/testground/sdk-go v0.2.2
+require github.com/testground/sdk-go v0.2.6-0.20201013134246-8d8414e32da5
```
raulk, line 5: This looks saner.

pkg/task/task.go
```diff
@@ -77,3 +91,40 @@ func (t *Task) State() DatedState {
...
+		return fmt.Sprintf(`<a href="%s" target="_blank">%s</a>`, url, title)
+	}
+
+	return t.CreatedBy
+}
+
+// CreatedByCommit parses the task.CreatedBy field in case it contains a ;
+// https://github.com/filecoin-project/oni/commit/d8b5a13f1fc758c674ffcee6dbd7f13540997283
+// Currently CircleCI utilises the following format for the `user` field in the `.env.toml`
+// user="filecoin-project/oni -> '$CIRCLE_BRANCH';https://github.com/filecoin-project/oni/commit/'$CIRCLE_SHA1'
```
raulk, line 110: Can't we break this down into different fields? The `RunEnv` from prototype versions of Testground had fields for branch, commit, etc. It doesn't need to be part of the `RunEnv`, but I think modelling those source control attributes in Testground is the way forward.

pkg/task/storage.go
```diff
@@ -145,8 +167,23 @@ func (s *Storage) ChangePrefix(dst string, src string, id string) error {
...
+	case StateProcessing:
+		prefix = prefixProcessing
+	case StateComplete:
+		prefix = prefixComplete
+	}
+
+	return s.rangeIter(prefix, start, end)
+}
+
+// range returns []*Task with all tasks between the given time ranges.
```
raulk, line 185:
```suggestion
// rangeIter returns []*Task with all tasks between the given time ranges.
```

raulk (approved):
This are looking really good! 😍 Honestly, most of my comments are just nits. I plan to review the whole thing more holistically once it's on master, before it's released.

The two sticking points for me are:
* very direct dependency on k8s runner output/result. This _might_ be fine, if we consider that TaaS will only ever run with a k8s runner, but I don't think that should be the case. I would like to be able to set up Testground quickly on a remote machine with Docker, and use it to schedule jobs with `local:docker`. We don't need to resolve this right here, right now, but we should before we release TaaS as beta/GA.
* the log tailing. Now that we've landed run events via the sync service, I think we should not be tailing container/pod outputs at all. (1) Live updates will take the form of push run events, and (2) outputs should be downloaded from the runner.

pkg/config/env.go
```diff
@@ -36,15 +36,18 @@ type DockerHubConfig struct {
...
-	TasksInMemory   bool     `toml:"tasks_in_memory"`
-	SlackWebhookURL string   `toml:"slack_webhook_url"`
+	Listen                string   `toml:"listen"`
+	Tokens                []string `toml:"tokens"`
+	Workers               int      `toml:"workers"`
+	QueueSize             int      `toml:"queue_size"`
+	TasksInMemory         bool     `toml:"tasks_in_memory"`
+	SlackWebhookURL       string   `toml:"slack_webhook_url"`
+	GithubRepoStatusToken string   `toml:"github_repo_status_token"`
+	TaskTimeoutMin        int      `toml:"task_timeout_min"`
```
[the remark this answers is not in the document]
nonsense: `services` is a bit of overloaded term, i like `auth` and `scheduler`. what about `hooks` instead of `services`?

pkg/daemon/kill.go
```diff
@@ -0,0 +1,42 @@
...
+			fmt.Fprintf(w, "cannot kill tsk")
+			return
+		}
+
+		redirect := `
+      <script>
+         setTimeout(function(){
+            window.location.href = 'https://ci.testground.ipfs.team/tasks';
+         }, 1000);
+      </script>
```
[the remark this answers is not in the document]
nonsense: This is going to go away in the next 1-2 iterations, this is why I didn't bother. The whole view part needs to be done properly.

pkg/daemon/tasks.go
```diff
@@ -53,12 +56,54 @@ func (d *Daemon) listTasksHandler(engine api.Engine) func(w http.ResponseWriter,
...
+		allocatableCPUs, allocatableMemory, _ := rr.GetClusterCapacity()
+
+		_, _ = w.Write([]byte("<strong>cluster resources</strong><br/>"))
+		_, _ = w.Write([]byte(fmt.Sprintf("capacity cpus: %d<br/>", allocatableCPUs)))
+		_, _ = w.Write([]byte(fmt.Sprintf("capacity memory: %s<br/>", humanize.Bytes(uint64(allocatableMemory)))))
+
 		tf := "Mon Jan _2 15:04:05"
 
-		fmt.Fprintf(w, "<table><th>task id</th><th>type</th><th>name</th><th>state</th><th>created</th><th>updated</td><th>outputs tgz</th><th>task logs</th><th>took</th>")
+		fmt.Fprintf(w, "<table><th>task id</th><th>type</th><th>name</th><th>created</th><th>updated</td><th>outputs tgz</th><th>task logs</th><th>task journal</th><th>took</th><th>status</th><th>outcomes</th><th>error</th><th>actions</th><th>created by</th>")
```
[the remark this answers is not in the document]
nonsense: Let's decide on how to handle the view part in a follow up PR, it is not just this table, but also other handlers.

pkg/engine/engine.go
```diff
@@ -208,12 +212,13 @@ func (e *Engine) QueueRun(request *api.RunRequest, sources *api.UnpackedSources)
...
-		Version:  0,
-		Priority: request.Priority,
-		Plan:     request.Composition.Global.Plan,
-		Case:     request.Composition.Global.Case,
-		ID:       id,
-		Type:     task.TypeRun,
+		Version:     0,
+		Priority:    request.Priority,
+		Plan:        request.Composition.Global.Plan,
+		Case:        request.Composition.Global.Case,
```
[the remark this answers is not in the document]
nonsense: No, will remove these.

plans/example/go.mod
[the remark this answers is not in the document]
nonsense: `go mod tidy` is run as part of CI, I think this is correct. We might need to remove the go.sum entirely and recreate it.

plans/network/pingpong.go
nonsense: I think yes, because the ping-pong is requesting specific IPs. @hacdias that's correct, right?

pkg/task/task.go
nonsense: Yes, that's an absolute hack right now. I wasn't even sure that we want this metadata about run to be propagated as part of `user`, but didn't want to add more flags, so piggy-backed on the `user` field.

pkg/engine/engine.go
nonsense: This might be running before the signal is even created, so we would need another channel to propagate signals... basically a subscription for all signals that govern the lifecycles of the tasks.

Why do you think it is racy? It might be delayed a few seconds, but that's fine.

pkg/daemon/journal.go
```diff
@@ -0,0 +1,61 @@
...
+			return
+		}
+
+		tsk, err := engine.GetTask(taskId)
+		if err != nil {
+			fmt.Fprintf(w, "cannot fetch tsk")
+			return
+		}
+
+		result := decodeResultK8s(tsk.Result)
```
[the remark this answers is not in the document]
nonsense: Yes, but it will be difficult to get the right abstraction now, without having used a few runners for a while with all the necessary outputs stored in `result`. So I'd rather we make it specific, until we nail the correct abstraction, rather than work with a bad abstraction.

---

I guess once we introduce more runners in TaaS, or once we want to use these dashboards locally, we will have to solve that problem.

pkg/task/task.go
nonsense: This is about propagating the `commit`, `branch`, etc. from the `testground client` to the `testground daemon` however, I don't see how `RunEnv` applies here.

nonsense:
> I think we should get rid of the tailing entirely. The way I see it:

> we have run events in the sync service to update the user view.
we should consume those run events live, and update the UI accordingly. IMO that's what the new "tail" should be.
results aren't really actionable/useful until the run has finished, so those should be downloaded as outputs.

This is used when a user schedules a task, and then wants to follow its progress:
a) from the CLI
b) from the web interface (via the /logs handler)

---

Run events (from testplan instances) is not the only relevant information at that point, as a user you care about the logs from the actual runner, and these are stored in a log file, so we seem to need to do the tailing of the log file, unless we store these aggregate runner + testplan instances outputs in another store, and not a file.

pkg/runner/cluster_k8s.go
nonsense: I both agree and disagree with you:
a) agree because this is really hacky and confusing and not consistent across runners.
b) disagree about removing it now, because we will lose stdout and it won't be available to users, and this will make it much harder to debug test plans, that are incorrectly using stdout for various debug messages and errors.

I'd rather have a better UX at the moment, and slightly harder and complicated codebase, than well architected runner and bad UX.

pkg/engine/engine.go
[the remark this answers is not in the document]
nonsense: Actually this will introduce a circular dependency, so needs a larger refactor. I'd rather not do it in an already big PR.

pkg/api/engine.go
```diff
@@ -54,9 +56,11 @@ type Engine interface {
 	QueueBuild(request *BuildRequest, sources *UnpackedSources) (string, error)
 	QueueRun(request *RunRequest, sources *UnpackedSources) (string, error)
 
-	Status(id string) (*task.Task, error)
-	Logs(ctx context.Context, id string, follow bool, cancel bool, ow *rpc.OutputWriter) (*task.Task, error)
 	Tasks(filters TasksFilters) ([]task.Task, error)
+	GetTask(id string) (*task.Task, error)
+	Logs(ctx context.Context, taskId string, follow bool, cancel bool, w io.Writer) (*task.Task, error)
+	Kill(taskId string) error
+	DeleteTask(taskId string) error
```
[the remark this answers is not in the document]
nonsense: Adding the `interface` now, but will do a refactor of the Engine later - it is not straight-forward to just remove the methods, the signals needs to be abstracted, and methods like `Kill` are using it, so the interface will grow.

pkg/config/env.go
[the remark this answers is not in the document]
nonsense: I extracted `SchedulerConfig` as it has 4 fields already. `Auth` is a bit premature right now - there is only 1 config (`Tokens`). Decided to postpone `Github` and `Slack` extraction, as Github will need to support multiple repos maybe, so might be refactored.

pkg/daemon/journal.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/testground/testground/pkg/api"
+	"github.com/testground/testground/pkg/logging"
+)
+
+func (d *Daemon) getJournalHandler(engine api.Engine) func(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
nonsense: I also don't like the `journal` name. I am a bit hesitant to call it `RunnerOutput`, because it sounds like `run output` and this is something else... could be the `outputs` we collect or the `api.RunOutput`...

pkg/task/task.go
nonsense: Refactoring this, so that we can include `metadata-commit`, `metadata-branch` and `metadata-repo` from the CLI when submitting a composition. These would be used by CircleCI for example.

nonsense:
@raulk I've addressed some of your comments, but not all. While I agree with pretty much all of them, they require larger refactors, that I'd rather not do in an already big PR. Will go over the comments and the TODOs in the PR, and create issues for each of them, so that we address in easier-to-digest PRs in the coming week.

---

Thanks for the multiple reviews @hacdias @raulk

PR #1156, opened 2020-10-02
adding StateCanceled for task ; adding kill cmd ; refactor Task ; extract timeouts

nonsense:
Relies on:
- https://github.com/testground/testground/pull/1154
- https://github.com/testground/sdk-go/pull/31

We should either first merge these PRs, or merge this into `tasks-status` and do an overall review.

---

This PR is:
1. Adding a `kill` command, so that we can cancel tasks that are currently processing or scheduled.
2. Refactoring `Task` package - making impl. details private (such as `prefixes` for the LevelDB schema) and unifying the names between the in-memory Queue and the Task.
3. Explicitly adding a `StateCanceled` for tasks.
4. Extracting `task` timeout as a config. We now have two timeouts:
a) `task` includes both the `build` and `run` duration of a task. Consider it as a global timeout.
b) `run` is the timeout of the run part of a test plan. It should be less that `task` timeout.
5. Adding `user` field to the `[client]` testground config. This field is attached to tasks so that we identify who triggered a task.


nonsense:
@raulk actually I am merging this, because it is getting difficult to maintain a working build between various branches on:
`go-sdk`
`testground`
`lotus`
`lotus-soup` test plan

I am reducing a bit complexity here, and will address comments directly on https://github.com/testground/testground/pull/1154 after we make review and merge `go-sdk`.

PR #1158, opened 2020-10-09
feat: builder docker/node and sdk-js

hacdias:
This PR is aligned with the efforts in https://github.com/testground/sdk-js/issues/1.

- Adds a `docker/node` builder with the following base Dockerfile:
	```docker
	ARG BASE_IMAGE
	FROM ${BASE_IMAGE} AS builder
	ENV PLAN_DIR /plan
	WORKDIR /plan
	COPY . /
	RUN npm ci
	EXPOSE 6060
	ENTRYPOINT [ "npm", "start"]
	```
- Adds a test plan `example-js` with the following tests:
	- pingpong
	- sync
	- failure
	- output (not complete because metrics aren't finished yet)
- Adds `example-js/pingpong` to CI


hacdias:
We have to make [`network.Config`](https://github.com/testground/sdk-go/pull/34/files) the most compatible possible. Next week, I will tackle on what @raulk mentioned in the linked PR and will also implement an Unmarshaller/Decoder for that struct.

Some things such as `net.IPNet` are not easily converted to the same thing as Go. My suggestion is to just pass around the string CIDR. That will virtually work on all implementations since there's always a function to convert CIDRs to native structs.

We can keep using `net.IPNet` in Go but I'll need to implement an unmarshaller.

raulk:
Yeah, I think struct tags will do for the casing. And for IPNet, there's actually a marshaller/unmarshaller you can use. Let me try digging it up!

raulk:
Here it is: https://github.com/testground/sdk-go/blob/master/runtime/runparams.go

hacdias:
Just throwing it out again: I think we should collect all this types in some place! 😆

hacdias:
I'm changing the base of this PR to #1154 because that one includes the newer SDK with the updates to all tests. However, I'd prefer if we could merge #1154 into master when possible, since that one is not even based on master.

hacdias:
Update: ping-pong test working!

hacdias:
@nonsense @raulk I feel this PR is finished. Even though sdk-js isn't complete (missing metrics before releasing 1.0), I feel this PR is ready for review and should be merged. Later, we can make further changes if needed.

pkg/build/docker_node.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	cliopts := []client.Opt{client.FromEnv, client.WithAPIVersionNegotiation()}
+
+	var (
+		basesrc  = in.UnpackedSources.BaseDir
+		cli, err = client.NewClientWithOpts(cliopts...)
+	)
+
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Minute)
```
nonsense, line 45: I see this is copied from `docker:go` builder, but we don't need it anymore, as we are always passing through the supervisor, which applies the `task timeout` (by default 10min, but configurable). So I suggest we delete this (maybe also directly delete in `docker:go` builder as well.

pkg/build/docker_node.go
```diff
@@ -0,0 +1,133 @@
...
+	defer cancel()
```
nonsense, line 46:
```suggestion
```

pkg/build/docker_node.go
```diff
@@ -0,0 +1,133 @@
...
+
+	if err != nil {
+		return nil, err
+	}
```
nonsense, line 50: Nitpick, but I find it weird that this error check is separated from the declaration of `cli`... not sure why we want to group `basesrc` and `cli` under the same `var` group.

plans/example-js/pingpong.js
```diff
@@ -0,0 +1,186 @@
...
+    const oldAddrs = os.networkInterfaces()
+
+    const config = {
+      network: 'default',
+      enable: true,
+      default: {
+        latency: 100 * 1000 * 1000, // 100ms in nanoseconds
+        bandwidth: 1 << 20 // 1 Mib
+      },
+      callbackState: 'network-configured', // Fix this small and big letters thingies.
```
nonsense, line 32: You have fixed that, right? Update comment?

plans/example-js/pingpong.js
```diff
@@ -0,0 +1,186 @@
...
+
+    // Make sure that the IP addresses don't change unless we request it.
+    const newAddrs = os.networkInterfaces()
+    if (!sameAddrs(oldAddrs, newAddrs)) {
+      throw new Error('interfaces changed')
+    }
+
+    runenv.recordMessage(`I am ${seq}`)
+
+    const ipC = (seq >> 8) + 1
```
nonsense, line 49: Considering that `ping-pong` is always using 2 instances (and doesn't work with more than 2 instances), this is a really weird way to update the C and D parts of the IP. I know this is copying the network/ping-pong Go impl., just saying we might want to refactor this soon :)

nonsense (approved):
Overall LGTM!

hacdias:
I'm going to address the comments and then merge!

pkg/build/docker_node.go
hacdias: Separated! I didn't think much of that. I used `docker:go` as a boilerplate and basically just removed what wasn't needed. MIssed that!

PR #1171, opened 2020-11-06
add route to gateway, so that we dont blackhole it

nonsense:
This PR is explicitly adding a route to the gateway, because otherwise packets to DNS (and other services) are dropped:

```
/ # ip route
28.116.0.0/16 dev eth1 scope link  src 28.116.128.0 # data network
30.0.0.0 dev eth1 scope link # data network
blackhole 100.64.0.0/10 # we want to drop packets to the services CIDR apart from DNS
100.64.0.10 via 100.96.4.1 dev eth0 # this is DNS
100.96.1.9 via 100.96.4.1 dev eth0 # this is influxdb
100.96.3.3 via 100.96.4.1 dev eth0 # this is redis
blackhole 100.96.4.0/24 # we want to drop packets to all destinations at 100.96.4.x except 100.96.4.1
100.96.4.1 dev eth0 scope link # the route we are adding in this PR
```

---

This problem got triggered when upgrading Kubernetes, weave and flannel, as well as host OS from `debian 9 stretch` to `ubuntu` `Ubuntu 20.04.1 LTS` (kernel from `4.9.0-11-amd64` to `5.4.0-1028-aws`) at https://github.com/testground/infra/pull/64

I am not sure why packets going through 100.96.4.1 (i.e. the gateway) were not being dropped until now.

### DIFF
--- a/pkg/sidecar/k8s_reactor.go
+++ b/pkg/sidecar/k8s_reactor.go
@@ -23,6 +23,7 @@ import (
 	"github.com/vishvananda/netlink"
 	"github.com/vishvananda/netlink/nl"
 	"github.com/vishvananda/netns"
+	"golang.org/x/sys/unix"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
@@ -302,6 +303,7 @@ func (d *K8sReactor) manageContainer(ctx context.Context, container *docker.Cont
 		}
 	}
 
+	var addedGwRoute bool
 	for _, r := range routesToBeDeleted {
 		// Don't route to the default route. Blackhole these routes.
 		bh := netlink.Route{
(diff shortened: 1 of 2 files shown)


hacdias (approved):
LGTM

pkg/sidecar/k8s_reactor.go
```diff
@@ -313,6 +315,25 @@ func (d *K8sReactor) manageContainer(ctx context.Context, container *docker.Cont
 			routeDst = r.Dst.String()
 		}
 
+		// detect the gateway => we know that we are going to delete the services CIDR, which are routed via the gateway,
+		// so we use they to figure out the gateway IP as well as the link index
+		if !addedGwRoute && routeDst == "100.64.0.0/10" {
```
raulk, line 320: Let's move this out to a configuration variable and make it settable through a CLI flag?

nonsense: Totally agree about extracting it - done.

I don't think we need a CLI flag for this - this configuration is highly coupled with the exact configuration of the Kubernetes cluster, so chances are it won't work if you have a different configuration (this PR is a good example of how fragile networking updates are).

PR #1173, opened 2020-11-12
metrics autodiscover / automated dashboards for testplans

nonsense:
This PR is:
1. Refactoring the rendering of tasks and dashboards, that is part of TaaS
- adding bootstrap for better looking tables and other UI
2. Adding `/dashboard` and `/data` endpoints that render dashboards for testplans (autodetecting metrics emitted by the testplan), and visualising them. `/data` endpoint is currently returning data from a specific metric to be visualised with a library such as Highcharts https://www.highcharts.com/
3. Making sure both `local:docker` and `cluster:k8s` runners are supported in the UI, by default on http://localhost:8042/

TODO:
- [x] replace Highcharts with another open charting library.
Resolution: for now replaced with `plotly`.

FOLLOW UP:
- [ ] update docs on http://docs.testground.ai/ with respect to - supervisor, asynchronous tasks, auto-discover for metrics, dashboards, --wait flag

TAAS:
- [ ] add retention rates to InfluxDB otherwise it will blow up at some point.

OPEN QUESTIONS:
- [x] do we want to bundle all frontend code together with Testground (for example `bootstrap`). I've been thinking whether we want to host Testground and Testground as a Service within the same app / image, and am still not decided. Input is welcome!
Resolution: for now I am bundling the assets as part of the Docker image, it makes it very easy to deploy Testground.


nonsense:
Marking this as ready for review, as the general improvements (dashboards, tasks list, autodiscovered metrics, etc.) are now tested on TaaS and mostly ready.

Tasks lists and dashboards are also working for `local:docker`, which I think is handy now that by default we schedule asynchronous tasks/jobs to Testground. Even if one works locally and iterates on a test plan, it is handy to have the dashboard open.

pkg/api/engine.go
```diff
@@ -59,7 +59,7 @@ type Engine interface {
 	QueueRun(request *RunRequest, sources *UnpackedSources) (string, error)
 
 	DoBuildPurge(ctx context.Context, builder, plan string, ow *rpc.OutputWriter) error
-	DoCollectOutputs(ctx context.Context, runner string, runID string, ow *rpc.OutputWriter) error
```
nonsense, line 62: `DoCollectOutputs` now retrieves a `Task` based on the `runID`, and depending on where this task was executed, fetched the outputs. This makes it possible for the dashboard to list all tasks (no matter which runner was used to run them), and provide links to their outputs.

pkg/config/env.go
```diff
@@ -41,6 +41,8 @@ type DaemonConfig struct {
 	Tokens                []string        `toml:"tokens"`
 	SlackWebhookURL       string          `toml:"slack_webhook_url"`
 	GithubRepoStatusToken string          `toml:"github_repo_status_token"`
+	RootURL               string          `toml:"root_url"`
```
nonsense, line 44: RootURL is used for some links on the dashboard. When run locally it should be set to:
```
root_url = "http://localhost:8042"
```

Whereas on a given deployment, such as TaaS, it is set to:
```
    root_url = "https://ci.testground.ipfs.team"
```

pkg/daemon/daemon.go
```diff
@@ -70,6 +77,11 @@ func New(cfg *config.EnvConfig) (srv *Daemon, err error) {
 		})
 	})
 
+	staticDir := "/static/"
+	r.PathPrefix(staticDir).Handler(http.StripPrefix(staticDir, http.FileServer(http.Dir("."+staticDir))))
+
+	r.HandleFunc("/data", srv.dataHandler(engine)).Methods("GET")
```
nonsense, line 83: `/data` returns data visualised by dashboards. It returns CSV data for a single panel.

pkg/daemon/tasks.go
```diff
@@ -58,38 +62,90 @@ func (d *Daemon) listTasksHandler(engine api.Engine) func(w http.ResponseWriter,
 
 		cr, _ := engine.RunnerByName("cluster:k8s")
 		rr := cr.(*runner.ClusterK8sRunner)
-		allocatableCPUs, allocatableMemory, _ := rr.GetClusterCapacity()
 
-		_, _ = w.Write([]byte("<strong>cluster resources</strong><br/>"))
-		_, _ = w.Write([]byte(fmt.Sprintf("capacity cpus: %d<br/>", allocatableCPUs)))
-		_, _ = w.Write([]byte(fmt.Sprintf("capacity memory: %s<br/>", humanize.Bytes(uint64(allocatableMemory)))))
+		var allocatableCPUs, allocatableMemory int64
+		if rr.Enabled() {
```
nonsense, line 67: This is something worth refactoring in the future. Not all Testground daemons will support all runners, and right now we don't have a good way of defining which runners a given deployment supports, i.e. which runners are enabled.

---

The purpose of this bit here is to make sure that we show relevant information when Testground is working in cluster mode (i.e. on TaaS), and hide this information when running Testground locally and we don't have a cluster available.

hacdias (approved):
LGTM! Nice dashboard

PR #1177, opened 2020-11-17
simplify network test plan

nonsense:
This PR is updating the network test plan to use `InitializedTestCaseFn`, reduce the code and update parts of the output.


hacdias (approved):
LGTM!

PR #1187, opened 2021-01-15
feat: http sync service

### LINKED ISSUE #1160: refactor sdk-go FetchAllEvents and SubscribeEvents
Follow up to https://github.com/testground/sdk-go/pull/31 and https://github.com/testground/testground/pull/1154

---

At the moment we have 2 methods, `FetchAllEvents` and `SubscribeEvents`, that consumers use to get events from the sync service via the `go-sdk`.

We can probably unify them into just one method which stops consuming events when it's context is canceled (i.e. `SubscribeEvents`), but that would also require a slight refactor on Testground side where we use `FetchAllEvents`.

### LINKED ISSUE #1186: Sync service
To build the new testground SDK for JavaScript, we will need to support browsers.  Thus, it is very difficult to make a Redis client work from the browser since Redis doesn't really use HTTP. The solution seems to be to convert the sync library we have on the Go SDK to a Go service that will work as an HTTP service where all tests can connect to and perform the necessary operations. By doing this, we will also remove the direct access to Redis from the tests, which was never the goal.

hacdias:
Closes #1186.

**Read more on: https://github.com/testground/sync-service/issues/4**

---

This PR introduces a new web service to Testground: the sync service. The sync service will act as the central service to which the tests should connect to get sync features (barriers, publishing, subscribing, entry signaling, event signaling). Previously, the tests were connecting directly to the Redis instance through the SDK. However, that brings us some issues:

- ~~In case we want to switch from Redis to a different technology (unlikely in the short term), we would need to change the implementation of all SDKs in all languages.~~ https://github.com/testground/sync-service/pull/8
- The support of Redis on non-backend technologies (such as browsers) is non-existent, which would not allow us to successfully build a JavaScript SDK that successfully works in the browser.
- By accessing Redis directly, a test may, on purpose, mess up with the purpose of Redis.

This new service is built to battle those issues, by:

- Providing an HTTP server to which each instance should connect. From that point on, a WebSocket connection should be created. This WebSocket will live as long as the test instance is alive. All connections to the sync service shall be done through this single WebSocket to avoid useless connections.
- The implementation of the sync service is abstracted by the `sync.Service` interface, which makes it easy to swap to a different backend in the future.

Regarding **testing**, I simply moved the tests from `sdk-go` that now belong here, and adapted them. Do you think we should do more unit tests or are the current integration tests enough?

---

## To do

- [x] Update runners/k8s to use the sync service (CTRL-F for `WatchClient`)
	- [x] local:docker
	- [x] local:exec
	- [x] cluster:k8s
- [x] Check locally if the new service works with all example tests locally.
- [x] Tests (most migrated from `sdk-go`)
- [x] Fix #1160
- [x] `sdk-go` https://github.com/testground/sdk-go/pull/37
- [x] It keeps environment variables and Redis healthcheck for support with older SDKs for now.

### Before merging

- [x] Update `sdk-go` version to definite one on tests and tg
- [ ] ~~Update infra repository to use the new sync service~~ Updating TaaS is deferred.


hacdias:
@nonsense I implemented the WebSockets as we discussed. Tomorrow, I will start updating the client in order to support the new sync server. Plus, I still need to add tests here.

I wanted to ask you to take a quick look to validate the architecture. A small recap of what I did so far was:

- `sync.Service` is an interface that should implement the actual sync actions using a certain backend. The current implementation, `DefaultService` uses Redis and it is mostly based on the original sync package from the Go SDK.
- We have an HTTP server that accepts connections, which must upgrade to WebSockets. For each connection, we have a `connection` that is responsible for consuming the new requests and consuming (sending) the responses that we generate. When we receive a new request, we spin a new goroutine to handle that request. Usually, those are short lived goroutines except for `Subscribe` and `Barrier`.

One of my concerns was spinning up goroutines for each request but, as I mentioned, most of them are short-lived. And for the long-lived ones, it's unlikely that a test is calling Barriers and Subscribes at the same time.

hacdias:
Independent binary: https://github.com/testground/sync-service

- Useful for creating docker images and publishing individually.
- Avoids weird dependencies between testground and sdk-go.
- TODO: Testground simply will spin a Docker image with the service.

hacdias:
I now noticed testground/pkg/runners and testground/pkg/sidecar both depend on sdk-go for the sync client. The usage of the watch client is making my life complicated as I'm not sure what's the use for it (or even the need).

I plan on moving the GC collection to the sync service.

@nonsense do you have any recommendation on what to do with the Watch Client? What was it used for? is it useful? Should the sidecar connect directly to Redis? Should I create some interface through the sync service for the watching thing?

nonsense:
The `WatchClient` is used by the `testground daemon` to fetch information about the run of individual testplan instances for a given run, and display that information on TaaS on the dashboard, i.e. `clients 998/1000 ; miners 10/10` in terms of `failed/ok` testplan instances.

The `testground daemon` would have to connect to the new `sync service` to fetch that information, when we introduce the sync service, that would replace the sync library and the direct connection to Redis.

hacdias:
## Current issues with this PR

### Issue 1

Every time a test fails, I get something like this on the console.

```
Feb 16 15:31:37.247686	INFO	1.3661s INCOMPLETE << sidecar      >>
2021-02-16 16:31:37.24772 +0100 CET m=+263.132244532 write error: write /Users/henriquedias/testground/data/daemon/XXXXXXXXXXXX.out: file already closed
```

So I guess I'm still somehow writing to that file while I shouldn't (see logs).

### Issue 2 (splitbrain)

Splitbrain **hangs** and then fails with deadline exceeded. It appears that pubsub might be only sending data to one sub.

---

PubSub system is made for only one sub. Make it support more subs. Take into account that multiple subs might've started in different points in time, hence lastid might differ.

hacdias:
Circle CI is failing with `context deadline exceeded` while building images. The other day I noticed something similar when using a specific commit from the `sdk-go` repository.

hacdias:
@nonsense this is almost finished. As far as I can see, the only missing thing is to add the new sync service to the cluster k8s. I tried and it failed on 59937b1. I commented the relevant code on 7a1ce46 that made it fail. I'm not an expert with k8s and I used it very few times. I will dig into it more on Thursday or Friday.

Nevertheless, I would like to receive a review on this so far as it is basically done, as well as for the sdk: https://github.com/testground/sdk-go/pull/37

nonsense:
@hacdias could you explain a bit more what's the problem with Kubernetes? We have to make sure this works on the `cluster:k8s` runner, as we are running many tests there - `lotus-soup`, `graphsync`, `ipfs`, etc.

Maybe you could start a new cluster, following https://docs.testground.ai/runner-library/cluster-k8s/how-to-create-a-kubernetes-cluster-for-testground so that we make sure the `cluster:k8s` runner continues to work, and we don't mess with https://ci.testground.ipfs.team/ before we are confident it works?

pkg/runner/local_docker.go
```diff
@@ -391,6 +429,14 @@ func (r *LocalDockerRunner) Run(ctx context.Context, input *api.RunInput, ow *rp
 	ctxContainers, cancel := context.WithCancel(ctx)
 	defer cancel()
 
+	eventsWg := sync.WaitGroup{}
+	eventsCancel, err := r.collectOutcomes(ctxContainers, result, &eventsWg, &template)
```
nonsense, line 433: I am a bit confused with the flow here. Why are we passing a WaitGroup to `collectOutcomes` and waiting for it only in the `select` below? I don't understand the changed flow here... also the WaitGroup has a single entry, so I am not sure of its purpose.

pkg/runner/local_docker.go
```diff
@@ -171,36 +171,82 @@ func (r *LocalDockerRunner) Healthcheck(ctx context.Context, engine api.Engine,
...
+
+	return cancel, nil
 }
 
 func (r *LocalDockerRunner) Run(ctx context.Context, input *api.RunInput, ow *rpc.OutputWriter) (runoutput *api.RunOutput, err error) {
+	log := ow.With("runner", "local:docker", "run_id", input.RunID)
+
+	err = r.setupSyncClient()
+	if err != nil {
+		r.lk.Unlock()
```
nonsense, line 245: This is also a bit confusing. we are doing an `Unlock` and it is not at all obvious where the `Lock` happened?

pkg/sync/go.mod
```diff
@@ -0,0 +1,12 @@
+module github.com/testground/testground/pkg/sync
+
+go 1.15
+
+require (
+	github.com/go-redis/redis/v7 v7.4.0
+	github.com/google/uuid v1.1.1
+	github.com/testground/testground v0.5.3
```
nonsense, line 8: Why is the `sync` service requiring Testground? We should create a separate repo for the sync service, so that we don't have nested go mods - versioning would be a mess otherwise I think.

pkg/runner/local_docker.go
hacdias: I am using the flow of `SubscribeEvents` instead of fetching them all in the end as described here: https://github.com/testground/testground/issues/1160. Maybe a channel is better to indicate that it finished instead of a waiting group.

If you think it's better to just collect them in the end as we did before, please let me know.

pkg/runner/local_docker.go
hacdias: Shouldn't be there. Thanks for noticing!

pkg/sync/go.mod
hacdias: Hmm... it's just the logging. But it needs to be a different package anyways, because the sdk-go needs to import some types... and testground imports sdk-go.

I'll work on moving it to a different repository then.

nonsense:
@hacdias I suggest we create the repo for the sync server - review it, merge it (without it being merged into sdk-go and testground), and continue with the rest of the issues (integrating the sync service) on a much smaller PR.

This PR is already rather big, and it is not complete, because we cannot merge to master without making sure `cluster:k8s` is working, with most of our internal testplans (`lotus-soup`, `gossipsub`, `go-graphsync`, etc.) and that `infra` repo is amended to run the sync service.

hacdias:
I moved the sync service to its own repository: https://github.com/testground/sync-service.

- Docker Hub integration needs to be set up so we can pull the docker images on the tests that run on this repository. Right now, I can't set it up: Docker Hub is not allowing me to choose the correct GitHub organization - it is connected to @testgroundbot.

hacdias:
**Thought**: to keep supporting tests that use the OLD sync service, we just need to keep exporting the `REDIS_HOST` variable.

hacdias:
Note for self: SSH into CircleCI; try check logs k8s.

.circleci/sync-service.yaml
```diff
@@ -0,0 +1,36 @@
...
+      hostPID: true
+      hostNetwork: true
+      dnsPolicy: ClusterFirstWithHostNet
+      containers:
+        - name: testground-sync-service
+          image: iptestground/sync-service:latest
+          command: ["/service"]
+          env:
+            - name: REDIS_HOST
+              value: "172.18.0.1"
```
hacdias, line 24: The network created for "kind" is 172.18.0.0/16, where the host of the containers is 172.18.0.1. We need to be able to access Redis which is running on a Docker container on host from the pod where the sync service is running.

hacdias:
I managed to update the testing environment to use the new sync service and tests are now passing.

@nonsense I will need your help with upgrading the 'infra' repository once this is approved or merged.

pkg/runner/cluster_k8s.go
```diff
@@ -277,14 +275,12 @@ func (c *ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow *rpc
...
-		erru := c.updateRunResult(&template, result)
-		if erru != nil {
-			ow.Errorw("could not update run result", "err", erru)
+		err := c.collectOutcomes(ctx, result, &template)
+		if err != nil {
+			ow.Errorw("could not start collecting outcomes", "err", err)
 		}
 
-		return err
+		return c.watchRunPods(ctx, ow, input, result, &template)
```
nonsense, line 283: @hacdias could you explain the change here, just so that we are on the same page? We no longer have `updateRunResult`, so I guess `collectOutcomes` collects results for individual test plan instances, while they are running, and not at the end (as previously), right?

pkg/runner/cluster_k8s.go
```diff
@@ -1187,31 +1188,40 @@ func (c *ClusterK8sRunner) GetClusterCapacity() (int64, int64, error) {
...
-		if result.Outcomes[g].Total != result.Outcomes[g].Ok {
+		result.Outcome = task.OutcomeSuccess
+		if len(result.Outcomes) == 0 {
 			result.Outcome = task.OutcomeFailure
-			break
 		}
-	}
+
+		for g := range result.Outcomes {
+			if result.Outcomes[g].Total != result.Outcomes[g].Ok {
```
nonsense, line 1219: @hacdias I believe we have a race here - how do we know that this has completed in the main thread, so that we return `result` only after this has been run?

pkg/runner/local_docker.go
```diff
@@ -172,36 +173,75 @@ func (r *LocalDockerRunner) Healthcheck(ctx context.Context, engine api.Engine,
...
-			break
 		}
-	}
+
+		for g := range result.Outcomes {
+			if result.Outcomes[g].Total != result.Outcomes[g].Ok {
+				result.Outcome = task.OutcomeFailure
+				break
+			}
+		}
```
nonsense, line 230: @hacdias same comment here - I think we need a channel/notification to the runner when this is done, so that we return only after `result` has been updated, no?

pkg/runner/cluster_k8s.go
hacdias: Yes, as they run.

pkg/runner/local_docker.go
hacdias: Fixed on c5fd5d3.

pkg/runner/cluster_k8s.go
hacdias: Fixed on 7b7f16e. Thanks :D

.circleci/sync-service.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: apps/v1
+kind: DaemonSet
```
nonsense, line 2: Nit, but this should be a deployment, not a daemon set - daemon set is something that is available on every single node of the cluster. I guess it is fine for tests, as we have only 1 node.

On the `infra` repo, this should be a deployment.

Makefile
```diff
@@ -55,7 +58,7 @@ test-integ-local-docker:
 	./integration_tests/09_docker_splitbrain_accept.sh
 	./integration_tests/10_docker_splitbrain_reject.sh
 	./integration_tests/11_docker_splitbrain_drop.sh
-	./integration_tests/12_docker_example-js_pingpong.sh
+	# ./integration_tests/12_docker_example-js_pingpong.sh
```
hacdias, line 61: SDK JS is a work in progress. We can re-add the tests afterwards.

PR #1234, opened 2021-05-22
feat: emit periodic goroutine dumps

### LINKED ISSUE #1232: add functionality to testground to periodically emit goroutine dumps via the sdk-go
At the moment most of the failures we encounters with Testground testplans on TaaS is that they timeout for various reasons, due to barrier not being hit, because of various bugs.

It'd be very helpful if we can add a parameters to the compositions files and request the `sdk-go` to trigger goroutine dumps every N seconds, so that we can review goroutine dumps for test runs that timed-out.

hacdias:
Fixes #1232.

Depends on https://github.com/testground/sdk-go/pull/41.

Composition example:

```toml
[metadata]
name    = "test"
author  = "test"

[global]
plan    = "network"
case    = "ping-pong"
builder = "docker:go"
runner  = "local:docker"
emit_dumps = 5
total_instances = 2

[global.build_config]
go_version = "1.16"

[[groups]]
id = "all"
instances = { count = 2 }
```

Emits goroutine dumps every 5 seconds on each test instance. Plan `network` needs to be upgraded to SDK in linked PR in order for this to work.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -47,7 +47,7 @@ require (
 	github.com/stretchr/testify v1.5.1
 	github.com/syndtr/goleveldb v1.0.0
 	github.com/testground/plan-templates v1.0.3
-	github.com/testground/sdk-go v0.2.8-0.20210522074938-6aa84502788f
+	github.com/testground/sdk-go v0.2.8-0.20210527111726-b1de9869438c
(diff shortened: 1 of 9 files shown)


nonsense (approved):
LGTM!

PR #1250, opened 2021-06-29
fix: enable Redis on k8s test

hacdias:
Enables Redis using the docker healthcheck on the first k8s integration test. Something similar is already being done on the second test. However, the first one was silently failing to connect to Redis.

### DIFF
--- a/integration_tests/01_k8s_kind_placebo_ok.sh
+++ b/integration_tests/01_k8s_kind_placebo_ok.sh
@@ -14,6 +14,7 @@ docker tag $ARTIFACT testplan:placebo
 kind load docker-image testplan:placebo
 
 pushd $TEMPDIR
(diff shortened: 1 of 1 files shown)


integration_tests/01_k8s_kind_placebo_ok.sh
```diff
@@ -14,6 +14,7 @@ docker tag $ARTIFACT testplan:placebo
 kind load docker-image testplan:placebo
 
 pushd $TEMPDIR
+testground healthcheck --runner local:docker --fix
```
nonsense, line 17:
```suggestion
// enable redis using the docker healthcheck on the first k8s integration test
testground healthcheck --runner local:docker --fix
```

nonsense (approved):

Issue #1213, opened 2021-03-21
extra_sources not available in docker con


dgrisham:
#### Version information:

I'm running the most recent version of `testground` on `master`:

```
$ git log | head -n1
commit 3da7c61944ad273856c45d6d90a20eaf9e0dfdb6
```

Arch Linux, x64

#### Description:

Hey! I'm attempting to run tests in Docker using the `extra_sources` feature. I have the following lines in my manifest set up to include my local versions of two repositories as extra sources:

```
[extra_sources]
docker_go = [
    "../test-datasets",
    "../extra/go-bitswap",
    "../extra/go-peertaskqueue"
]
```

and I've tried a few corresponding replacement directives in the test plan's `go.mod`, such as

```
replace github.com/ipfs/go-bitswap => ./extra/go-bitswap
replace github.com/ipfs/go-peertaskqueue => ./extra/go-peertaskqueue
```

and

```
replace github.com/ipfs/go-bitswap => /extra/go-bitswap
replace github.com/ipfs/go-peertaskqueue => /extra/go-peertaskqueue
```

and

```
replace github.com/ipfs/go-bitswap => /go-bitswap
replace github.com/ipfs/go-peertaskqueue => /go-peertaskqueue
```

and

```
replace github.com/ipfs/go-bitswap => ./go-bitswap
replace github.com/ipfs/go-peertaskqueue => ./go-peertaskqueue
```

When I run the test using the `local:docker` runner and `docker:go` builder, I get the following error:

```
go: github.com/ipfs/go-bitswap@v0.2.20: parsing /extra/go-bitswap/go.mod: open /extra/go-bitswap/go.mod: no such file or directory
```

I've tried various combinations of `go.mod` replace directives, but nothing seems to work. I'm fairly certain that my extra sources directories are being found, because when I change them to invalid values I get an error when testground tries to read the directories.

Am I doing something wrong here, or is this feature not quite complete? I know it was only on the `oni` branch awhile back, but was under the impression that it had been merged into master.

Thanks!

max-i-am:
I get the same issue when trying to add extra_sources to my test case. Inside the docker container the files are not available and probably not added at all. Therefore, I get the same error `../test-datasets: no such file or directory`.
Does somebody have a workaround or hot fix?

nikosft:
The same here. Any progress on that?

nikosft:
[...]

PR #1259, opened 2021-08-12
fix: extra_sources builder reference

daonb:
extra_sources wasn't working for me and when I dug deeper I found out there was a missing conversion. The builder had to be changed from `docker:generic` to `docker_generic` .

### DIFF
--- a/pkg/cmd/build.go
+++ b/pkg/cmd/build.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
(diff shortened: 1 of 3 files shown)


hacdias (approved):
This looks correct to me, as per [the docs](https://docs.testground.ai/concepts-and-architecture/builders#include-extra-sources) we can see that the extra sources are supposed to be declared as `snake_case`.

Perhaps this fixes #1213.

/cc @nonsense

nonsense (approved):

PR #1286, opened 2022-01-07
Allow connecting to additional containers on the control network

hannahhoward:
# Goals

Allow uses to connect to their own containers that they've spun up on the control network when using local docker.

I wanted to be able to ship spans to a jaeger instance I started on the testground-control network. Originally I thought I'd just use the fact that it was portmapped to host to accomplish this, but that didn't work. Instead, I spun it up on the control network manually then added this code and an entry in my env.toml to allow connecting to it.

I'm not sure if this is the best way to do this -- it'd be great if you could simply specify additional containers to start with the daemon as part of the healthcheck. Note that I want to go looks at spans after the test run is complete, so it can't simply start with the test run.

### DIFF
--- a/pkg/runner/local_docker.go
+++ b/pkg/runner/local_docker.go
@@ -3,7 +3,6 @@ package runner
 import (
 	"context"
 	"fmt"
-	"github.com/testground/testground/pkg/logging"
 	"io"
 	"io/ioutil"
 	"net"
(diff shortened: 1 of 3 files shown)


kerzhner:
[Our team](https://github.com/statechannels) building state channels has been using this feature to run a local Ethereum network that is used by testground instances. We would love to see this feature incorporated into the master branch!

kerzhner:
https://github.com/testground/testground/pull/1481 has been opened.

laurentsenta:
Closing in favor of https://github.com/testground/testground/pull/1481, massive thank you for sharing this work that state-channels reused @hannahhoward

Issue #1263, opened 2021-08-19
Stuck at `waiting for network to initialize` after Sidecar worker failure


hannahhoward:
#### Version information:

Commits all exhibiting the issue:
8e76ab86474215293d647a9d03927cdde4948b61 (master), 71ab494b5ea8a8c54b528aeafb9b6260e62257cf (fix/1230), and eaf986c5bdeb126870217a03b86b8a7cab871b38 ( a few commits back)

Environment:
Apple Silicon M1 + Docker Desktop 3.6.0 (Docker Engine 20.10.18)

#### Description:

Unable to run any testground test plans.
Used https://github.com/protocol/beyond-bitswap/tree/master/testbed/testbed
Used https://github.com/ipfs/go-graphsync/tree/master/testplans/graphsync
Replicated even with just the ping pong test -- `testground run single --plan network --testcase ping-pong --builder=docker:go --runner=local:docker --instances=2`

After docker images build, I see this
`sidecar worker failed: failed to initialise the container: failed to get handle to network namespace: protocol family not supported`
followed by some json that looks something like this:
`	{"host": "unix:///var/run/docker.sock", "container": "97626bb09939f15a9059b0abfa18fd33d7ed6f2c671375d293fb365d1e045254"}`

Then all containers get permaneantly stuck at `waiting for network to initialize`

I tried wiping running docker images and volumes
I tried various settings Docker Desktop for Mac
I tried updating  github.com/vishvananda/netlink to v1.1.0 since that's where the error occurs
I tried rebuilding / reinstalling testground (I.e. rebuilding the sidecar docker image)

My machine used to work for testground. Older sha was one that already was running. Have no idea what might have changed other than:
go 1.15 - go 1.16
whatever upgrades to Docker Desktop / Docker Machine / Linux Kernel since I last ran it (did install an update to docker)

nonsense:
I am unable to reproduce this on current master `0b997a1` with Docker Desktop 4.1.1 on macOS (Intel CPU).

<img width="747" alt="Screenshot 2021-11-11 at 16 51 12" src="https://user-images.githubusercontent.com/50459/141327851-d7783864-520b-4e83-b662-765024e73494.png">

I will attempt to try on M1 and see if I can reproduce there.

galargh:
I have experienced the exact same behaviour on M1 Pro mac running Monterrey with the following versions:
- testground commit: 84e9dff
- docker client: 20.10.9
- go: 1.16.9
- docker engine: 20.10.10
[...]

PR #1297, opened 2022-03-12
Build for arm64

hannahhoward:
This is a fix for #1263 similar to #1296. I'm pretty sure this is non-ideal but it's what I use to make it work on my machine.


ackintosh:
It seems that this can be closed since https://github.com/testground/testground/pull/1372 has been merged.

PR #1321, opened 2022-04-22
fix: builder result parsing in web server

laurentsenta:


pkg/daemon/tasks.go
```diff
@@ -164,6 +169,27 @@ func decodeResult(result interface{}) *runner.Result {
...
+		return TaskOutcome{Outcome: result.Outcome, Content: result.String()}
+	default:
+		logging.S().Errorw("Unknown task type", "type", t.Type)
+		return TaskOutcome{Outcome: task.OutcomeFailure, Content: ""}
+	}
+}
+
 func unescape(s string) template.HTML {
 	return template.HTML(s)
 }
```
laurentsenta, line 195: (creating a Thread here)

It's a simple fix, it looks like `runner.Result` and the `TaskBuild` type have been added later. Might be worth revisiting how we manage results later.

laurentsenta:
Moving this back into Draft,
We're not using `testground build` anymore, so this is much less urgent, I might come back to this with a better fix after we fixed other task-status-related tasks.

PR #1327, opened 2022-05-02
plan/example-rust: update SDK and optimize build

laurentsenta:
- [x] update the SDK,
- [x] use a two-step "download package" - "build the main.rs" file in Dockerfile to speed-up iterations,
- [ ] Replace the temporary sdk-rust rewrite by a release


laurentsenta:
@mxinden @ackintosh ready for review!

plans/example-rust/empty.rs
```diff
@@ -0,0 +1,2 @@
+
+fn main() {}
```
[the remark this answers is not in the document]
laurentsenta: It's a placeholder main file we use to speed up the docker build:
https://github.com/testground/testground/pull/1327/files#diff-ea696a48937940f71a6d8064e6ca41eb74b1ff17bc0e18d475c83cadaf77f1ceR4-R6

I moved it into the `Dockerfile`

.circleci/config.yml
```diff
@@ -127,19 +136,54 @@ jobs:
...
+    steps:
+      - setup-install-bins
       - run:
-          name: "run local:docker integration tests"
-          command: make test-integ-local-docker
+          name: "build testground"
+          command: make install
+      - run:
+          name: "run cluster:k8s integration tests"
+          command: make test-integ-cluster-k8s
```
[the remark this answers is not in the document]
laurentsenta: Nice catch, I amended the commit, thanks!

laurentsenta:
Note:
This PR adds the rust example to CI tests,
so I'm also using it as an opportunity to split the different tests (exec, docker, k8s, examples) and do some cleaning.

plans/example-rust/Dockerfile
```diff
@@ -1,9 +1,20 @@
-FROM rust:1.57-bullseye as builder
+FROM rust:1.59-bullseye as builder
 WORKDIR /usr/src/testplan
+
+# Cache dependencies between test runs,
+# See https://blog.mgattozzi.dev/caching-rust-docker-builds/
+# And https://github.com/rust-lang/cargo/issues/2644
+
+RUN mkdir -p ./plan/src/
+RUN echo "fn main() {}" > ./plan/src/main.rs
```
mxinden, line 9: :+1:

plans/example-rust/Dockerfile
```diff
@@ -1,9 +1,20 @@
-FROM rust:1.57-bullseye as builder
+FROM rust:1.59-bullseye as builder
 WORKDIR /usr/src/testplan
+
+# Cache dependencies between test runs,
```
mxinden, line 4: :rocket:

plans/example-rust/src/main.rs
```diff
@@ -2,9 +2,10 @@ use std::net::{Ipv4Addr, TcpListener, TcpStream};
 
 const LISTENING_PORT: u16 = 1234;
 
-#[async_std::main]
+#[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let mut sync_client = testground::sync::Client::new().await?;
+    println!("TODO: why is this necessary to make the test go green?");
+    let (client, _run_parameters) = testground::client::Client::new().await?;
```
mxinden, line 8:
```suggestion
    let (client, _run_parameters) = testground::client::Client::new().await?;
    client.wait_network_initialized().await?;
```

Would this resolve the failures as well @laurentsenta?

laurentsenta: That looks much more stable, thanks for the patch @mxinden 

A quick note, the tests are still failing because the `run.out` file is empty, it's a file we collect (via `--collect`) at the end of the run. It looks like testground expects the SDK to write logs to a predefined folder,
Related code in the go-sdk:

https://github.com/testground/sdk-go/blob/d78f61eccaa37b0d941f98650bce5dbb8fd7b7f1/runtime/runenv_logger.go#L29

I'm going to skip this check:

- The rust SDK doesn't implement this logger yet as far as I can tell,
- More importantly, it might be a good time to discuss logging, do we even need that parser, why not "just" rely on stdout and stderr? (and get rid of the pretty.go file at the same time).
- When we fix https://github.com/testground/testground/issues/1349 simply `--wait`'ing for a run will detect eventual failures.

mxinden: > The rust SDK doesn't implement this logger yet as far as I can tell,

The rust-sdk does have the stdout hack:

https://github.com/testground/sdk-rust/blob/00c280fd23b75f19f80a45bcfa752298f109640d/src/background.rs#L375-L388

It does not support `run.out`.

> More importantly, it might be a good time to discuss logging, do we even need that parser, why not "just" rely on stdout and stderr? (and get rid of the pretty.go file at the same time).

Off the top of my head, that would be the cleaner solution.

@laurentsenta what do you think of simply adding a `touch run.out &&` before the entrypoint in the `Dockerfile`?

laurentsenta: It's a fair solution, 

The way we test success with `run.out` made sort of sense without the `--wait` fix but it's arbitrary: https://github.com/testground/testground/blob/master/integration_tests/header.sh#L33-L38

If we both think it's worth looking into getting rid of `run.*` files, I'm in favor of not cluttering the images at all and starting relying on actual exit codes.

mxinden: > If we both think it's worth looking into getting rid of `run.*` files, I'm in favor of not cluttering the images at all and starting relying on actual exit codes.

Off the top of my head I don't have a use-case for it. That said, I am not a testground expert. :+1: for proposing to remove it in this repository.

Issue #1337, opened 2022-05-17
Add additional parameters to groups in composition files


laurentsenta:
## Description

We're working on interoperability tests with the go & rust libp2p projects,
One feature we need is the ability to test different versions of the go-library which means having the ability to customize the docker build image and the dockerfile extensions **per group**.

Example of a desired composition file:

```yaml
[metadata]
  name = "ping-interop-cross-version"

[global]
  plan = "libp2p/ping-interop/go"
  case = "ping"
  total_instances = 4
  builder = "docker:go"
  runner = "local:docker"

[[groups]]
  id = "v0.11"
  instances = { count = 1 }

  [groups.build]
    docker_build_image  = 'golang:1.14-buster' # <= This is new
    selectors = ['compatv1'] # implements a shim for compat with different API versions
    dockerfile_extension = { pre_mod_download = 'mv ./v0.11/go.* ./' }
      
[[groups]]
  id = "v0.17"
  instances = { count = 1 }

  [groups.build]
    docker_build_image  = 'golang:1.16-buster'
    selectors = ['compatv2']
    dockerfile_extension = { pre_mod_download = 'mv ./v0.17/go.* ./' }

[[groups]]
  id = "v0.19"
  instances = { count = 1 }

  [groups.build]
    docker_build_image  = 'golang:1.18-buster'
    selectors = ['compatv2']
    dockerfile_extension = { pre_mod_download = 'mv ./v0.19/go.* ./' }
```

## What defines this endeavour to be complete?

- [ ] create a corresponding integration test with a similar manifest,
- [ ] add a field build configuration to a composition's group (same content as the global group),
- [ ] make sure the overrides manifest > composition.global > composition.groups are working as expected,
- [ ] make sure the build caching & deduping are taking into account these parameters,
- [ ] have the test passing & use in libp2p.

laurentsenta:
This is related to:
https://github.com/testground/testground/issues/1305
https://github.com/libp2p/test-plans/pull/23 (we're going to implement something similar to the "end goal" described)

PR #1338, opened 2022-05-17
feat: Add additional parameters to groups in composition files

laurentsenta:
Fixes https://github.com/testground/testground/issues/1337

This PR allows build_config override in groups (in composition files). See the example below which has a group with a custom docker build image.

It also introduces a `modfile` option for the docker:go builder that lets you override the go.mod (and go.sum) used to build the application, similar to the `-modfile` option in go build tools.

```yaml
[metadata]
  name = "issue-1337-override-builder-configuration"

[global]
  plan = "integrations"
  case = "issue-1337-override-builder-configuration"
  total_instances = 2
  builder = "docker:go"
  runner = "local:docker"

[[groups]]
  id = "v14"
  instances = { count = 1 }

  # --- NEW ---
  [groups.build_config]
    docker_build_image  = "golang:1.14-buster"
    modfile = 'go.v14.mod'
  # --- END NEW ---

  [groups.build]
    selectors = ["v14"]

  [groups.run.test_params]
    expected_version = "v14"
```

## Todo

- [x] create a corresponding integration test with a similar manifest,
- [x] add a field build configuration to a composition's group (same content as the global group),
- [x] make sure the overrides manifest > composition.global > composition.groups are working as expected,
- [x] make sure the build caching & deduping are taking into account these parameters,
- [x] find the "right" way to update the BuildKey key,
- [x] have the test passing,
- [x] test with libp2p interop.

### DIFF
--- a/pkg/build/docker_go.go
+++ b/pkg/build/docker_go.go
@@ -631,10 +631,6 @@ RUN cp ${PLAN_DIR}/go.mod ${PLAN_DIR}/go.sum /tmp/
 COPY . /
 RUN cp /tmp/go.mod /tmp/go.sum ${PLAN_DIR}/
 
-# Copy again the modfiles in case there where overwritten.
-COPY /plan/${MODFILE} ${PLAN_DIR}/go.mod
-COPY /plan/${MODFILE_SUM} ${PLAN_DIR}/go.sum
-
 {{.DockerfileExtensions.PostSourceCopy}}
 
 {{.DockerfileExtensions.PreBuild}}


pkg/build/docker_go.go
```diff
@@ -596,8 +626,14 @@ RUN echo "Using go proxy: ${GO_PROXY}" \
...
 {{.DockerfileExtensions.PreSourceCopy}}
 
-# Now copy the rest of the source and run the build.
+# Now copy the rest of the source and run the build. Make sure we backup modfiles first.
+RUN cp ${PLAN_DIR}/go.mod ${PLAN_DIR}/go.sum /tmp/
 COPY . /
+RUN cp /tmp/go.mod /tmp/go.sum ${PLAN_DIR}/
+
+# Copy again the modfiles in case there where overwritten.
+COPY /plan/${MODFILE} ${PLAN_DIR}/go.mod
```
brdji, line 635: Are the lines 629 - 632 necessary? Since lines 635-636 will overwrite the `go.mod` and `go.sum`, I don't see the point in copying the files to `/tmp/`, and then back again.

brdji (approved):

pkg/build/docker_go.go
laurentsenta: Nice catch, thanks for the note!

It looks like this now:

https://github.com/testground/testground/pull/1338/files#diff-9b5fa505e6278f70a7479cc5cc065c37fd0c374023b8b89b124d948ffc1025e7R630-R632

We do keep a backup of the modfile, instead of overwriting it.
Because earlier in the Dockerfile, when we `go mod download`, the command updated the `go.mod` file with the effective module versions.

Issue #1220, opened 2021-04-19
tasks page isn't working when starting daemon from a non-project-root directory 


huiscool:
#### Version information:

commit be5f89d28d3d38bdc279a455657abcfd4760b2ab

macOS Big Sur Version 11.2.2
#### Description:

**What you were doing when you experienced the bug:**
I start the testground deamon from a directory that is not the testground code root. Say that `~/code/testground` is the directory where I cloned the testground code, and `make install` has been run. Then I start daemon from `~` or `~/testground`, open `localhost:8042/tasks` in my browser, and I get:
![image](https://user-images.githubusercontent.com/30804350/115183419-4be75d00-a10e-11eb-9c93-bf40464202e1.png)
**Any error messages you saw, *where* you saw them, and what you believe may have caused them (if you have any ideas)**
The log in daemon says:
```
2021/04/19 12:53:39 http: panic serving 127.0.0.1:58012: cannot ParseFiles with tmpl/tasks: open tmpl/tasks.html: no such file or directory
goroutine 35 [running]:
net/http.(*conn).serve.func1(0xc0000cf860)
	/usr/local/go/src/net/http/server.go:1801 +0x147
panic(0x25cf340, 0xc0000a9b60)
	/usr/local/go/src/runtime/panic.go:975 +0x47a
github.com/testground/testground/pkg/daemon.(*Daemon).listTasksHandler.func1(0x2b53500, 0xc000022000, 0xc000124200)
	/Users/huyingcong/code/testground/pkg/daemon/tasks.go:142 +0x1671
net/http.HandlerFunc.ServeHTTP(0xc00000ee00, 0x2b53500, 0xc000022000, 0xc000124200)
	/usr/local/go/src/net/http/server.go:2042 +0x44
github.com/testground/testground/pkg/daemon.New.func2.1(0x2b53500, 0xc000022000, 0xc000124200)
	/Users/huyingcong/code/testground/pkg/daemon/daemon.go:76 +0x14f
net/http.HandlerFunc.ServeHTTP(0xc0004c0100, 0x2b53500, 0xc000022000, 0xc000124200)
	/usr/local/go/src/net/http/server.go:2042 +0x44
github.com/gorilla/mux.(*Router).ServeHTTP(0xc0004ac0c0, 0x2b53500, 0xc000022000, 0xc000124000)
	/Users/huyingcong/go/pkg/mod/github.com/gorilla/mux@v1.7.3/mux.go:212 +0xd3
[...]

PR #1339, opened 2022-05-19
Add tmpl_dir to daemon env.

brdji:
Fix for #1220 
Because the daemon is not running in the same directory where the `tmpl` folder is located, it cannot access the template files.

tmpl_dir is the path to the `tmpl` directory in the testground root project.
It will be read from the env.toml file, if available; if not, the default ('tmpl') will be used.

This fix does not affect the default case (ie. running the daemon from the project root); in this case, the env var can be unset, and the templates will still be read correctly

### DIFF
--- a/env-example.toml
+++ b/env-example.toml
@@ -40,6 +40,8 @@ ulimits = [
 
 [daemon]
 listen                    = ":8080"
+# Path to the testground `tmpl` directory (can be ignored if you run testground from its root directory)
+tmpl_dir                  = "my/testground/dir/tmpl"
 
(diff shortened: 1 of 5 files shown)


nonsense:
I don't have strong feelings on how to implement this, but given how thin the HTML/CSS are for Testground, maybe embedding these files in the binary is also a good option, to avoid having more configuration options.

brdji:
My intention was to avoid modifying the installation, but I agree with your suggestion. I'll take a look into embedding the templates.

nonsense (approved):

Issue #1304, opened 2022-04-07
go examples fail because InitializedTestCase is not used


laurentsenta:
#### Version information:

```bash
› git log | head -n1
commit 6ae5c95b246ce61ec4382f260ad99ae739967786
```

macOS, 12.3.1, Intel,
Running on local docker,

#### Description:

When I run the go example it ends with failed outcome:

```sh
testground run single --plan=example --testcase=output --builder=docker:go --runner=local:docker --instances=1
```

This is somewhat expected: these examples do not use the [`InitializedTestCase` wrapper](https://github.com/testground/sdk-go/blob/d78f61eccaa37b0d941f98650bce5dbb8fd7b7f1/run/invoker.go#L169-L173).
Without this wrapper, the test won't send the "success message" to testground and it'll assume the test failed.

See also: https://github.com/testground/testground/issues/1303#issuecomment-1090143102

One fix is to add the wrapper to the tests:

- tests: https://github.com/testground/testground/blob/master/plans/example/main.go
- example: https://github.com/testground/testground/blob/master/plans/network/main.go#L9

A more thoughtful fix might be to make `InitializedTestCase` the new default, see:
https://github.com/testground/sdk-go/blob/d78f61eccaa37b0d941f98650bce5dbb8fd7b7f1/run/invoker.go#L46-L65

brdji:
I now understand that `INCOMPLETE` is unrelated to the actual job outcome, thanks!

Yes, using `InitializedTestCase` does solve the issue, and the job's outcome is correctly shown as successful in the dashboard.

>A more thoughtful fix might be to make `InitializedTestCase` the new default, see

This would be my preferred solution, however I'm not sure if it will break any existing test plans that are using it already.

laurentsenta:
What if we release a new go-sdk version where the default case does the init, and we provide a new uninitialized case?

Something like:
https://github.com/testground/sdk-go/compare/master...feat/default-initialized-test-case

laurentsenta:
(removing myself until I have time to take this)

PR #1347, opened 2022-06-06
fix: testground run with --wait exits with code = 0 on some failures

laurentsenta:
Fixes https://github.com/testground/testground/issues/1349
Contributes to https://github.com/testground/testground/issues/1304

## Context:

Testground defines a `task.Task` object with an anonymous Result field. The content of this field depends on the builder or runner used, for example:

- if the task is Build => Result contains an array of artifacts (probably docker images),
- if the task is Run, with a local_exec runner => Result contains nil,
- if the Task is Run, with a k8s or local docker runner => Result contains a struct with different Outcomes field.

That's fine for storage / serialization (when used as a wire format), but any piece of code that "receive" a task object needs to interpret the `Result` field by knowing about the runners. There are hints of the code required in the `daemon/tasks.go` code, see related PR: https://github.com/testground/testground/pull/1321. It create cyclic dependencies and bugs (because not all parts of the code use the same interpretation, see issue #1349).

Later it would be nice to rethink this data model.

## Tasks

- [x] introduce a `data.DecodeTaskOutcome` function, (a non-intrusive way to make sure we have ONE reusable way to compute task outcomes),
- [x] make sure running a job with `--wait` will fail with exit code != 0 if the outcome is not a success.
- [x] Return with an exit code 3 if the command failed during collect.
- [x] Update tests: now that we check for actual outcomes, most test are failing


laurentsenta:
I'll revisit https://github.com/testground/testground/pull/1321 when we approve & merge this.

laurentsenta:
(temporarily adding the fix from https://github.com/testground/testground/pull/1353 to make sure the CI is passing)

pkg/data/result_test.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	r1 := DecodeRunnerResult(result1)
+	assert.NotNil(t, r1)
+
+	result2 := [2]string{"artifact", "artifact2"}
+	r2 := DecodeRunnerResult(result2)
+	assert.NotNil(t, r2)
+}
+
+func TestDecodeTaskOutcome(t *testing.T) {
```
brdji, line 29: I'd rather split this large test into several smaller unit tests, each one testing a specific outcome (and the decoding of the outcome).
E.g. `TestDecodeGenericUnknownOutcome`, `TestDecodeGenericSuccessOutcome`, etc.

laurentsenta: Fixed!

Issue #1357, opened 2022-06-15
Add builders parameters to groups in composition files


laurentsenta:
## Description

Subtask for https://github.com/testground/testground/issues/1340

We're working on interoperability tests with the go & rust libp2p projects, see https://github.com/testground/testground/issues/1340 and https://github.com/testground/testground/issues/1305.

We need the ability to write a test that uses different builders (like `docker:go` and `docker:generic`).

Example of a desired composition file:

```toml
[global]
  plan = "integrations_mixed_builders"
  case = "issue-1357-mix-builder-configuration"
  total_instances = 2
  runner = "local:docker"

[[groups]]
  id = "go"
  instances = { count = 1 }
  builder = "docker:go" # override builder here

  [groups.build_config]
    path = "go/" # define base path here

  [groups.run.test_params]
    expected_implementation = "go"

[[groups]]
  id = "generic"
  instances = { count = 1 }
  builder = "docker:generic" # override builder here

  [groups.build_config]
    path = "generic/" # define base path here
    build_args = { image = "golang:1.16-buster" }

  [groups.run.test_params]
    expected_implementation = "generic"
```

## What defines this endeavor to be complete?

- [x] create a corresponding integration test with a similar manifest,
- [x] move the build configuration fields to the groups,
- [x] make sure the overrides manifest > composition.global > composition.groups are working as expected,
- [ ] have the test passing & use in libp2p.

Issue #1305, opened 2022-04-08
Run plan with two instances based on two different builds


mxinden:
:wave: libp2p steward here.

## Description

My larger goal is to test that go-libp2p and rust-libp2p are compatible.

As a first step I would like to:

- Run two instances, one based on go-libp2p, one based on rust-libp2p.
- Connecting to each other.
- Exchanging libp2p ping protocol messages.

I have a first proof-of-concept here: https://github.com/libp2p/test-plans/pull/20

Here I define two test plans, one Golang based, one Rust based:

- https://github.com/libp2p/test-plans/pull/20/files#diff-e604ac0c6c3c5649b5554487d4c69f1344a947bfb512f458af7c476d36e8378f
- https://github.com/libp2p/test-plans/pull/20/files#diff-13de5a9ebadea1c629f4768d0c538dcfdf3bd36b323e902f560add44d79359a8

I then define a *compositional* test plan which uses the outputs of the two standard plans above and overrides the `[groups.run] artifact` with the corresponding Docker images

https://github.com/libp2p/test-plans/pull/20/files#diff-2266c0d0d968091b0c145b03fb719583bc664631ccda4e844e07b1852e15cfe1

This is a very manual process. I first have to build both standard test plans, scan the output for the Docker image names, update the *compositional* test plan and only then execute it.

Can the same be achieved in an automatic fashion, excluding the manual steps? If not, is supporting this use-case on the roadmap?

Feedback on https://github.com/libp2p/test-plans/pull/20 directly is also much appreciated.

//CC @nonsense

laurentsenta:
@mxinden It looks like there is no testground-native feature for composition right now, but we can automate your approach and reuse a few patterns found across the testground repo.

I built on top of your branch and poc'd automating composition files creation & dependency overrides:
https://github.com/libp2p/test-plans/pull/22/files#diff-4d2a8eefdf2a9783512a35da4dc7676a66404b6f3826a8af9aad038722da6823

I'm still working on it, but once I get the test to pass on a dev machine and the CI I'll probably need your input regarding the test content itself.

mxinden:
@laurentsenta instead of a shell script, should this functionality not be a build-in feature of testground? I expect this feature to be useful beyond the libp2p compatibility use-case.

laurentsenta:
@mxinden Definitely, I started looking at the builder implementation, it looks pretty doable, and I'm sure @nonsense might have some input on how to implement this.

[...]

PR #1367, opened 2022-06-27
interop: add builder and path option in grousp

laurentsenta:
Draft: Simpler approach to interop testing.

Fixes https://github.com/testground/testground/issues/1357 and https://github.com/testground/testground/issues/1305

## Tasks

- [ ] tbd


plans/_integrations_mixed_builders/generic/Dockerfile
```diff
@@ -0,0 +1,14 @@
+ARG image
+FROM ${image} AS builder
+
+WORKDIR /build
+ENV CGO_ENABLED 0
+COPY . .
+
+ARG PLAN_PATH
+RUN cd plan/${PLAN_PATH} && go build -a -o /testplan
```
laurentsenta, line 9: This is how you use `PLAN_PATH` in docker:generic builder.

laurentsenta:
@mxinden asking for your review, if you have time to take a look at the API (the manifest in the PR description) and how we use it in the `Dockerfile`.

pkg/api/composition_test.go
```diff
@@ -80,6 +89,57 @@ func TestValidateGroupBuildKey(t *testing.T) {
...
+			{
+				ID: "another-id",
+			},
+		},
+	}
+
+	require.EqualValues(t, []string{"docker:generic", "docker:go"}, c.ListBuilders())
+}
+
+func TestBuildKeyWithoutGroupPanics(t *testing.T) {
```
galargh, line 115: Aren't we testing that the opposite here? That calling `buildKey` panics if there is no builder?

pkg/build/docker_go.go
```diff
@@ -578,8 +582,11 @@ ARG RUNTIME_IMAGE=busybox:1.31.1-glibc
 #:::
 FROM ${BUILD_BASE_IMAGE} AS builder
 
+# PLAN_PATH is the path of our test's source code.
+ARG PLAN_PATH
```
galargh, line 586: Just to confirm. This defaults to empty string, right?

galargh (approved):
🥇

pkg/api/composition_test.go
laurentsenta: Good catch, renamed!

pkg/api/composition_test.go
```diff
@@ -80,6 +89,57 @@ func TestValidateGroupBuildKey(t *testing.T) {
...
+
+	g2 := &Group{
+		ID:      "with-go",
+		Builder: "docker:go",
+	}
+
+	k1 := g1.BuildKey()
+	k2 := g2.BuildKey()
+
+	require.NotEqualValues(t, k1, k2)
```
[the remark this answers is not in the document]
laurentsenta: Thanks for double-checking! I updated the test to answer your question,

The detail is: no, we don't rely on `ID` to compute the `BuildKey`, we use this key to deduplicate builds, so if 2 groups have the same build parameters but different IDs, we want them to have the same build key.

pkg/build/docker_go.go
laurentsenta: It defaults to `./` (in `docker_generic.go`), this way `Dockerfile` authors don't need to change their code on move.

plans/_integrations_mixed_builders/manifest.toml
```diff
@@ -0,0 +1,21 @@
+name = "integrations"
```
mxinden, line 1: Is this `manifest.toml` still needed even though we have the composition file?

plans/_integrations_mixed_builders/generic/Dockerfile
```diff
@@ -0,0 +1,14 @@
+ARG image
+FROM ${image} AS builder
```
mxinden, line 2: I am guessing that this is how I would be passing feature flags from the composition file to the `cargo build` command? Works for me. Great.

mxinden (commented):
:rocket: Interface looks good to me. That should be all we need to build both Go and Rust.

plans/_integrations_mixed_builders/generic/Dockerfile
laurentsenta: Yes! And we'd use the `PLAN_PATH` below to dispatch between go and rust code.

brdji (approved):

PR #1370, opened 2022-06-28
Upgrade to go 1.18

lalexgap:
I'm attempting to stress test our [client](https://github.com/statechannels/go-nitro) which was written with go 1.18. Since the docker containers only use go 1.16 building the test via docker would fail. 

To fix this I've upgraded our fork of testground to go 1.18 (mostly done by finding and replacing). I figured I'd raise a PR in case it's helpful in any way.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ version: 2.1
 parameters:
   go-version:
     type: string
-    default: "1.16"
+    default: "1.18"
   workspace-dir:
     type: string
     default: "/home/circleci"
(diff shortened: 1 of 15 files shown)


laurentsenta:
Hi @lalexgap thanks for sharing this PR, that would update the whole testground project and all its content to go 1.18,
I'm not against it, but it looks like you /just/ want to use a different version in your test, have you tried changing the build image?

In your manifest:

```toml
[builders."docker:go"]
build_base_image      = "golang:1.18-buster"
```

A more complete example is available here in the plan/dockercustomize folder: https://github.com/testground/testground/blob/master/plans/dockercustomize/manifest.toml

You can also customize build config with compositions:
https://github.com/testground/testground/blob/master/plans/_integrations_interop/_compositions/issue-1337-groups-builder-configuration.toml#L19-L20

PR #1372, opened 2022-06-29
Do not specify GOARCH when building to support arm64

### LINKED ISSUE #1298: Support more platforms/archs
At the moment Testground is compiled against `amd64`. Given that Testground depends heavily on Docker, we should extract the arch component and make sure Testground can be configured to work on other platforms, such as `arm64`, too.

Related: https://github.com/testground/testground/pull/1297 ; https://github.com/testground/testground/issues/1263

lalexgap:
Fixes #1298 

This is the same change as https://github.com/testground/testground/pull/1296 but has been rebased on the latest `master`.

It removes `GOARCH=amd64` when calling go build which would fail on `arm64` machines (like M1 macs). We've been used this in our fork to use testground on m1 macs.

### DIFF
--- a/Dockerfile.sidecar
+++ b/Dockerfile.sidecar
@@ -19,7 +19,7 @@ COPY . /
 # Testground version
 ARG TG_VERSION
 
-RUN cd / && CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags "-X github.com/testground/testground/pkg/version.GitCommit=${TG_VERSION}" -o testground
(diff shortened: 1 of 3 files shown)


laurentsenta:
That's awesome, thanks for creating this PR @lalexgap 
Note I created a [follow-up task](https://github.com/testground/testground/issues/1373) to make arm support more official.

@brdji, @galargh, @nonsense I'll merge by the end of the day if there's no issue with this,

brdji (approved):

PR #1375, opened 2022-07-06
Update env-example.toml

dektech:
Adding a comment to explain how users can connect to the tg-daemon running inside a cluster, instead of setting up a LB in front.

### DIFF
--- a/env-example.toml
+++ b/env-example.toml
@@ -45,7 +45,7 @@ listen                    = ":8080"
 task_timeout_min          = 20
 task_repo_type            = "disk"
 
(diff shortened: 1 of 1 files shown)


sysrex:
NITPICK - While this has helped me greatly, I would specify in the documentation that .env.toml endpoint actually refers to the testground-daemon endpoint (Kubernetes Service Object) just for clarity.

dektech:
> NITPICK - While this has helped me greatly, I would specify in the documentation that .env.toml endpoint actually refers to the testground-daemon endpoint (Kubernetes Service Object) just for clarity.

Added a commit further clarifying what the endpoint is and how to target it.

laurentsenta (approved):

PR #1381, opened 2022-07-08
Fix HTTP static file serving

lalexgap:
When trying to access the http server for the daemon I noticed that the static css/js files didn't seem to load:

image

Removing the trailing slash in the path passed into `http.Dir` seemed to resolve this and allow the HTTP server to properly serve up the static files.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -77,8 +77,8 @@ func New(cfg *config.EnvConfig) (srv *Daemon, err error) {
 		})
 	})
 
-	staticDir := "/static/"
-	r.PathPrefix(staticDir).Handler(http.StripPrefix(staticDir, http.FileServer(http.Dir("."+staticDir))))
+	staticPrefix := "/static/"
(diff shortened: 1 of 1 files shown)


laurentsenta:
Thanks for creating this PR @lalexgap, I finally had time to take a look at it: I can reproduce the 404 when I start the daemon outside the testground folder (which contains the static assets I guess). But I'm sure this PR would fix the issue.

What happens if you run `git checkout master && make goinstall && testground daemon` from the testground folder? do you still have the 404 error?

The solution might be something around embedding the assets in the binary, it was discussed in https://github.com/testground/testground/pull/1339

lalexgap:
> Thanks for creating this PR @lalexgap, I finally had time to take a look at it.
> 
> I can reproduce the 404 error only if I start the daemon **outside** the testground folder (which contains the static assets I guess). I'm not sure this PR would fix the issue.
> 
> What happens if you run `git checkout master && make goinstall && testground daemon` from the testground folder? do you still have the 404 error?
> 
> The solution might be something around embedding the assets in the binary, it was discussed in #1339

Ah shoot. I was probably running from outside the testground folder. 🤦  Sorry!

PR #1393, opened 2022-07-20
Fix storm test

brdji:
The benchmarks/storm test does not work unless this parameter is renamed properly.

### DIFF
--- a/plans/benchmarks/manifest.toml
+++ b/plans/benchmarks/manifest.toml
@@ -57,6 +57,6 @@ instances = { min = 1, max = 100000, default = 5 }
   [testcases.params]
   conn_count       = { type = "int", desc = "number of TCP sockets to open", default = 5 }
   conn_outgoing    = { type = "int", desc = "number of outgoing TCP dials", default = 5 }
-  conn_delay       = { type = "int", desc = "random milliseconds jitter before TCP dial", default = 30000 }
+  conn_delay_ms       = { type = "int", desc = "random milliseconds jitter before TCP dial", default = 30000 }
   concurrent_dials = { type = "int", desc = "max number of concurrent net.Dial calls", default = 10 }
   data_size_kb     = { type = "int", desc = "size of data to write to each TCP connection", default = 128 }


laurentsenta:
Good catch, should we add this to integration tests?

laurentsenta (approved):

brdji:
> Good catch, should we add this to integration tests?

Good idea, I can add a simple integration test to run storm with a few instances.

plans/benchmarks/benchmarks.go
```diff
@@ -43,11 +44,12 @@ func NetworkInitBench(runenv *runtime.RunEnv) error {
 
 	elapsed := time.Since(startupTime)
 	runenv.R().RecordPoint("time_to_network_init_secs", elapsed.Seconds())
+	// runenv.RecordMessage("We gucci")
```
laurentsenta, line 47: dead code?

plans/benchmarks/storm.go
```diff
@@ -189,10 +190,11 @@ func Storm(runenv *runtime.RunEnv) error {
 	_ = client.MustSignalAndWait(ctx, sync.State("done writing"), runenv.TestInstanceCount)
 	runenv.RecordMessage("done writing after barrier")
 
-	runenv.RecordSuccess()
+	// runenv.RecordSuccess()
```
laurentsenta, line 193: dead code?

laurentsenta (approved):
Nice 👍 let's remove dead code and merge!
(wait for CI to succeed, you can use "re-run from failed" if needed).

plans/benchmarks/benchmarks.go
brdji: Nice catch, I thought I deleted these lines :facepalm:

integration_tests/17_docker_benchmark_storm_ok.sh
```diff
@@ -0,0 +1,24 @@
...
+    --plan=testground/benchmarks \
+    --testcase=storm \
+    --builder=docker:go \
+    --runner=local:docker \
+    --instances=2 \
+    --wait | tee run.out
+
+assert_run_outcome_is ./run.out "success"
+
+popd
```
galargh, line 24: @brdji Is there a reason not to run this in CI currently? I don't see it in https://github.com/testground/testground/blob/master/Makefile#L59

Issue #691, opened 2020-03-17
sidecar crashes if redis is unavailable


coryschwartz:
This was reported by  @aschmahmann. He demonstrated an issue in which the redis server hit its connection limit, the sidecar would also crash, and tests were left unable to move forward beyond the network initialization phase.

When this happeens, there is no automated recovery, and no obvious way for users to understand what is going on.

Although I was never able to reproduce the port exhaustion as he was seeing, I  am able to able to produce a similar outcome by simply killing the redis container as soon as it is started.

I think there are a couple of things we could do to improve the situation.
It occurs to me that the kubernetes runner will restart the pods when a crash occurs, but that same experience is unmet in the local docker runner, even though it could easily be made to do so.

1. Start local docker containers with a restart policy, so they are restarted when/if they crash.
2. Don't crash when partitioned from the redis server. Maybe implement backoff retries.

raulk:
In this case, we should fail the test run, and report what happened. We shouldn't recover automatically. It would be useless in this case, since we anyway run Redis with no persistence.

An attempt to recover will anyway be performed by a subsequent test run, which will implicitly call the healthcheck/fix logic that will notice Redis, sidecar, etc. are absent, and will start them again.

nonsense:
I'd rather we do 1, than 2. Adding recovery logic, backoff retries, etc. makes the code complicated and Redis should really be available. The sync service is critical, we should not be adding work arounds for it, while it is not there.

PR #1395, opened 2022-07-21
chores: simplify labels

laurentsenta:
* drop duplicated labels
* move dif => exp
* introduce kind/epic
* remove labels that are not directly actionable
* introduce SDK labels
* remove redundant statuses

### DIFF
--- a/.github/labels.yml
+++ b/.github/labels.yml
@@ -27,18 +27,15 @@
 - name: area/core
   color: 0bb1ed
   description: "Area: Core"
-- name: area/sdk-sync
+- name: area/sdk
   color: 0bb1ed
-  description: "Area: SDK | Sync Service"
-- name: area/sdk-runtime
-  color: 0bb1ed
-  description: "Area: SDK | Runtime"
-- name: area/sdk-testplans
-  color: 0bb1ed
-  description: "Area: SDK | Test plan API"
+  description: "Area: SDK"
 - name: area/sidecar
   color: 0bb1ed
   description: "Area: Sidecar"
+- name: area/sync-service
+  color: 0bb1ed
+  description: "Area: Sync Service"
 - name: area/observability
   color: 0bb1ed
   description: "Area: Observability"
@@ -60,9 +57,22 @@
 - name: area/testing
   color: 0bb1ed
   description: "Area: Testing"
-- name: area/test-plan
+- name: area/ci
   color: 0bb1ed
-  description: "Area: Test Plans"
+  description: "Area: Continuous Integration"
+
+###
+### SDKs
+###
+- name: sdk/rust
+  color: 0bb1ed
+  description: "SDK: Rust"
+- name: sdk/js
+  color: 0bb1ed
+  description: "SDK: Javascript"
+- name: sdk/go
+  color: 0bb1ed
+  description: "SDK: Go"
 
 ###
 ### Kinds
@@ -70,9 +80,6 @@
 - name: kind/bug
   color: c92712
   description: "Kind: Bug"
-- name: kind/problem
-  color: c92712
-  description: "Kind: Problem"
 - name: kind/investigation
   color: fcf0b5
   description: "Kind: Investigation"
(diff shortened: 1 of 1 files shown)


laurentsenta:
@galargh @brdji @nonsense 
In preparation for future triage sessions,
I am simplifying the labels, could you take a look and let me know what you think?

.github/labels.yml
```diff
@@ -91,44 +98,38 @@
 - name: kind/tracking-issue
   color: fcf0b5
   description: "Kind: Tracking Issue"
+- name: kind/epic
+  color: 3e4b9e
+  description: "Kind: EPIC"
 - name: kind/question
   color: fcf0b5
   description: "Kind: Question"
-- name: kind/enhancement
```
brdji, line 97: Maybe we should leave a tag for "Feature" issues?

.github/labels.yml
```diff
@@ -229,25 +218,13 @@
...
 
 ###
 ### Statuses
 ###
-- name: status/done
-  color: edb3a6
-  description: "Status: Done"
-- name: status/approved-waiting
-  color: edb3a6
-  description: "Status: Approved and Awaiting Merge"
```
brdji, line 250: Since we're removing the "Done" and "Approved" labels, is the following the desired flow:
- I merge my PR into the master branch
- Instead of changing the status of the issue, I simply close the issue with a comment (referencing the PR)

.github/labels.yml
```diff
@@ -202,24 +203,12 @@
 ###
 ### Hints
 ###
-- name: hint/good-first-issue
-  color: 0623cc
-  description: "Hint: Good First Issue"
```
galargh, line 207: Don't we want to keep this one around? Could be nice if we wanted to pull in someone from Launchpad, for example, to work on the project.

.github/labels.yml
```diff
@@ -229,25 +218,13 @@
 - name: hint/needs-community-input
   color: 0623cc
   description: "Hint: Needs Community Input"
```
galargh, line 220: When would we use this one?

.github/labels.yml
galargh: I think your intuition is right here and I like this flow better 👍

.github/labels.yml
```diff
@@ -229,25 +218,13 @@
...
 ###
-- name: status/done
-  color: edb3a6
-  description: "Status: Done"
-- name: status/approved-waiting
-  color: edb3a6
-  description: "Status: Approved and Awaiting Merge"
 - name: status/changes-requested
   color: edb3a6
   description: "Status: Changes Requested in Review"
```
galargh, line 230: If we're getting rid of approved, etc., do we want to remove this one too?

.github/labels.yml
```diff
@@ -269,9 +246,6 @@
 - name: status/waiting
   color: edb3a6
   description: "Status: Waiting"
```
galargh, line 248: What's this one for?

.github/labels.yml
laurentsenta: There is a "github magic label" at the top, I'd like to remove the duplicate label and use the one support by github.
https://github.com/testground/testground/blob/4dc2e51cf17f39c18bd149cfe00fc8ca83358235/.github/labels.yml#L11-L16

.github/labels.yml
laurentsenta: Whenever we're making a change that impacts users and want their feedback. Not only maintainers.

.github/labels.yml
laurentsenta: Good question, taking a deeper look, we could drop all of them but "blocked".

.github/labels.yml
laurentsenta: There is a `kind/feature` a few lines above

.github/labels.yml
laurentsenta: You can also add a tag in your PR,
"fixes #issue-id" will autoclose the issue on merge.
"contributes to #issue-id" is also useful on larger tasks (usually a smell that your issue is too big though).

.github/labels.yml
laurentsenta: see pass 2!

.github/labels.yml
brdji: My bad! Didn't notice the lines collapsed.

.github/labels.yml
galargh: I can see only one issue labeled as such so far - https://github.com/testground/testground/issues/691

It's a bit unclear to me what are the actions we're waiting for if something is labelled with this (do we go off GitHub to find testground users? do we mention someone?) and when we can remove this label (how many voices from the community is enough?).

I think the discussion I'm turning this into is out of scope for this PR (thanks for doing this btw, I think it's a great initiative).

So I'm voting for resolving/closing this thread here.

.github/labels.yml
brdji: Great. I think that's the way to go - add "fixes #issue" to auto-close it, and leave a description in the issue if further clarification is necessary.

.github/labels.yml
laurentsenta: Agreed, we dropped something like 25 labels, if we have one left, that's already a great improvement,
thanks for the feedback!

For what it's worth if/when we discuss this again: when it comes to community involvement & working in the open, I think we should rely on the experience from other projects, with more activity. We can find examples of this label in [go-libp2p](https://github.com/libp2p/go-libp2p/issues?q=is%3Aopen+is%3Aissue+label%3Aneed%2Fcommunity-input) and [kubo](https://github.com/ipfs/kubo/issues?q=is%3Aopen+is%3Aissue+label%3Aneed%2Fcommunity-input).

laurentsenta:
@galargh @brdji @nonsense, thanks for the inputs, this went faster than I thought :)
Please review soon, so we can merge and raise the discussion of having triage sessions.

galargh (approved):

laurentsenta:
@brdji Could you take one last look and approve or comment?

brdji (approved):

PR #1399, opened 2022-07-26
Limit runs per branch

### LINKED ISSUE #1204: Option to keep only 1 active job for every repo/branch that submits jobs to TaaS
At the moment a given branch can trigger multiple jobs to Testground. Ideally we don't want to run 1000 jobs, if a user pushes 1000 commits to a branch, but rather only the last one.

If a user push a new commit, it should trigger a job on TaaS, and optionally kill all the previous tasks, created by that branch.

brdji:
Fixes #1204 

- The task queue should remove tasks belonging to the same branch and repo whenever a new task is added
- For the 1st iteration, I have added this removal to the `engine`. It is possible to automatically prune the queue whenever a new task is pushed, though I feel this might obfuscate the inner workings.

Please let me know what you think.

### DIFF
--- a/integration_tests/19_limit_runs_per_branch.sh
+++ b/integration_tests/19_limit_runs_per_branch.sh
@@ -47,8 +47,8 @@ t2=$(testground run single \
 run_id2=$(echo $t2 | awk '/run is queued with ID: / {print $59}')
 
 # First run must be canceled
-check_testground_task_status "$run_id1" 'canceled'
+assert_testground_task_status "$run_id1" 'canceled'
 # Second must succeed
-check_testground_task_status "$run_id2" 'success'
+assert_testground_task_status "$run_id2" 'success'
 
 popd
--- a/integration_tests/19_limit_runs_per_branch.sh
+++ b/integration_tests/19_limit_runs_per_branch.sh
@@ -1,0 +1,54 @@
+#!/bin/bash
+my_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "$my_dir/header.sh"
+
+testground plan import --from ./plans --name testground
+
+pushd $TEMPDIR
+
+testground build single \
+    --plan testground/placebo \
+    --builder docker:go \
+    --wait | tee build.out
+export ARTIFACT=$(awk -F\" '/generated build artifact/ {print $8}' build.out)
+docker tag $ARTIFACT testplan:placebo
+
+testground healthcheck --runner local:docker --fix
+
+# Run first test case, but do not --wait
+# We must start the 2nd case at the same time
+t1=$(testground run single \
+    --plan=testground/placebo \
+    --testcase=ok \
+    --builder=docker:go \
+    --use-build=testplan:placebo \
+    --runner=local:docker \
+    --instances=1 \
+    --metadata-branch b1 \
+    --metadata-repo r1 \
+    --collect)
+
+# Very unwieldy, $59 is the ID in the log
+run_id1=$(echo $t1 | awk '/run is queued with ID: / {print $59}')
+
+# Run the second test case, and wait for it to finish
+t2=$(testground run single \
+    --plan=testground/placebo \
+    --testcase=ok \
+    --builder=docker:go \
+    --use-build=testplan:placebo \
+    --runner=local:docker \
+    --instances=1 \
+    --metadata-branch b1 \
+    --metadata-repo r1 \
+    --wait \
+    --collect)
+
+run_id2=$(echo $t2 | awk '/run is queued with ID: / {print $59}')
+
+# First run must be canceled
+check_testground_task_status "$run_id1" 'canceled'
+# Second must succeed
+check_testground_task_status "$run_id2" 'success'
+
+popd
--- a/integration_tests/header.sh
+++ b/integration_tests/header.sh
@@ -73,18 +73,32 @@ function assert_run_output_is_correct {
   test $SIZEOUT -gt 0 && test $SIZEERR -eq 0
 }
 
+# Directory where the daemon and each test will store its outputs
 TEMPDIR=`mktemp -d`
-mkdir -p ${HOME}/testground
-cp env-kind.toml ${HOME}/testground/.env.toml
 
-echo "Starting daemon and logging outputs to $TEMPDIR"
-testground daemon > $TEMPDIR/daemon.out 2>&1 &
-DAEMONPID=$!
+# Start the testground daemon, loading the .env file from the first parameter
+function start_daemon {
+  env_file=$1
 
-sleep 2
+  mkdir -p ${HOME}/testground
 
-echo "Waiting for Testground to launch on 8040..."
-while ! nc -z localhost 8040; do
-  sleep 1
-done
-echo "Testground launched"
+  cp $env_file ${HOME}/testground/.env.toml
+
+  echo "Starting daemon and logging outputs to $TEMPDIR"
+  testground daemon > $TEMPDIR/daemon.out 2>&1 &
+  DAEMONPID=$!
+
+  sleep 2
+
+  echo "Waiting for Testground to launch on 8040..."
+  while ! nc -z localhost 8040; do
+    sleep 1
+  done
+  echo "Testground launched"
+}
+
+# if the SKIP_AUTO_START flag is unset or set to 0, start the daemon immediately
+if [[ ! -n "$SKIP_AUTO_START" || $SKIP_AUTO_START -eq 0 ]]; then
+  echo "Starting daemon automatically"
+  start_daemon 'env-kind.toml'
+fi
--- a/integration_tests/header.sh
+++ b/integration_tests/header.sh
@@ -25,8 +25,18 @@ function assert_run_outcome_is {
   RUN_ID=$(awk '/run is queued with ID/ { print $10 }' "${RUN_OUT_FILEPATH}")
   echo "checking run ${RUN_ID}"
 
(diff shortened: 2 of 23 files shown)


pkg/engine/engine.go
```diff
@@ -239,8 +240,17 @@ func (e *Engine) QueueRun(request *api.RunRequest, sources *api.UnpackedSources)
 				Created: time.Now().UTC(),
 			},
 		},
-		CreatedBy: task.CreatedBy(request.CreatedBy),
-	})
+		CreatedBy: cby,
+	}
+
+	// Remove existing tasks from same branch end repo before pushing a new task
+	err := e.queue.RemoveExisting(newTask.CreatedBy.Branch, newTask.CreatedBy.Repo)
```
laurentsenta, line 247: What happens if, as a developer working on my machine, I try to submit multiple jobs to my daemon?
I guess the `cby.Branch` and `cby.Repo` are empty strings and my jobs will cancel each others.

pkg/task/queue_test.go
```diff
@@ -78,6 +79,75 @@ func TestQueueReloads(t *testing.T) {
...
+	/// Both queues will use the same storage
+	inmem := storage.NewMemStorage()
+	db, err := leveldb.Open(inmem, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	ts := &Storage{db}
+
+	// open q1 and push an item into the queue
+	q1, err := NewQueue(ts, 100, convertTask)
```
laurentsenta, line 92: let's call it q if we only have one in this test

pkg/task/queue_test.go
```diff
@@ -78,6 +79,75 @@ func TestQueueReloads(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = q1.RemoveExisting(branch, repo)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	assert.Equal(t, 2, q1.tq.Len())
```
laurentsenta, line 148: Can we check the ids too?
Can we check if we kept the latest ids? maybe push more items

pkg/engine/engine.go
```diff
@@ -239,8 +240,17 @@ func (e *Engine) QueueRun(request *api.RunRequest, sources *api.UnpackedSources)
...
+
+	if err != nil {
+		return "", err
+	}
+
+	err = e.queue.Push(newTask)
```
laurentsenta, line 253: There is a race condition + it's quite stateful,
What about we implement a single operation? (`PushWithoutDuplicate`, `PushUniqTask`, or some better naming).

pkg/task/queue.go
```diff
@@ -90,6 +90,34 @@ func (q *Queue) Pop() (*Task, error) {
...
+	keep_tasks := make([]*Task, len(keep_indexes))
+	for index, value := range keep_indexes {
+		keep_tasks[index] = (*q.tq)[value]
+	}
+
+	*q.tq = keep_tasks
+
+	return nil
+}
+
```
laurentsenta, line 120: It looks like we're dropping tasks, are we marking them as canceled?

pkg/task/queue_test.go
```diff
@@ -78,6 +79,75 @@ func TestQueueReloads(t *testing.T) {
...
+}
+
```
laurentsenta, line 150: Can we add an integration test?

For example, we submit 3 tasks with the same repo + branch and then call testground status on each (t1, t2, t3).
We expect t1 outcome = success, t2 outcome = canceled, t3 outcome = success.

pkg/engine/engine.go
brdji: Correct - I forgot to add a check that `branch` and `repo` should not be empty strings

pkg/task/queue.go
brdji: I wanted to remove the tasks from the DB  (ie. mark them as cancelled), but I wasn't sure what's the proper way to handle this.
Should I use `storage.Delete(taskId)`, or `ArchiveTask(task)` ?

pkg/task/queue_test.go
brdji: Great idea, I'll add an integration test.

pkg/engine/engine.go
brdji: I'm not sure I follow - you mean replacing the lines 247 - 253 with a single operation, e.g. `PushUniqTask` ?

laurentsenta: Yes, one way to reason about this:
- Both operations lock the same resource.
- The state at the end of Push is correct if, and only if, we called it "right after" RemoveExisting.

That's a smell that:
- if you use one part of the API (Push) without the other (RemoveExisting), you might inadvertently create bugs in the app,
- interleaving the calls might trigger a race condition
  - Thread1: RemoveExisting(A, B)
  - Thread2: RemoveExisting(A, B)
  - Thread1: Push(A, B)
  - Thread2: Push(A, B)
  - => duplicate output.

pkg/task/queue.go
laurentsenta: I'm not sure but this looks storage related,
the way to go is probably to check the `supervisor.go` and see what happens when a task is canceled, I know we have states `StateCanceled` and `OutcomeCanceled` and I expect to see these, but I don't know the exact workflow atm.

pkg/engine/engine.go
brdji: I see, thanks for the write-up. I wanted to avoid changing the Queue API, but I see that adding a new unified operation is necessary.

pkg/task/queue.go
```diff
@@ -82,14 +109,68 @@ func (q *Queue) Pop() (*Task, error) {
...
+	logging.S().Debugw("queue.pop.got-task", "id", tsk.ID, "taskname", tsk.Name())
 	err := q.ts.ProcessTask(tsk)
 	if err != nil {
 		return nil, err
 	}
 	return tsk, nil
 }
 
+// Remove all existing tasks from the queue that match the given branch/string
+func (q *Queue) RemoveExisting(branch string, repo string) error {
```
laurentsenta, line 121: Let's make it private

pkg/task/queue.go
```diff
@@ -53,12 +54,18 @@ func (q *Queue) Push(tsk *Task) error {
 	q.Lock()
 	defer q.Unlock()
 
+	return q.pushLocked(tsk)
+}
+
+// Pushes an item to the priority queue, without acquiring a lock
+func (q *Queue) pushLocked(tsk *Task) error {
```
laurentsenta, line 61: Let's either call it `unsafePush` or just `push` (and assume private functions are unsafe).

integration_tests/header.sh
```diff
@@ -25,8 +25,18 @@ function assert_run_outcome_is {
...
   echo "checking run ${RUN_ID}"
 
+  check_testground_task_status "$RUN_ID" "$EXPECTED_OUTCOME"
+}
+
+# Assert the status of a testground task, as provided by the CLI output
+#
+# Usage:
+#   check_testground_task_status "###run_id123###" "success"
+function check_testground_task_status {
```
laurentsenta, line 35: Let's stick to something like `assert_task_status_is`

integration_tests/19_limit_runs_per_branch.sh
```diff
@@ -0,0 +1,54 @@
...
+    --metadata-repo r1 \
+    --wait \
+    --collect)
+
+run_id2=$(echo $t2 | awk '/run is queued with ID: / {print $59}')
+
+# First run must be canceled
+check_testground_task_status "$run_id1" 'canceled'
+# Second must succeed
+check_testground_task_status "$run_id2" 'success'
```
laurentsenta, line 52: 👍 

I thought we couldn't cancel a task that is already running,
does it mean that Testground is so slow to pick up tasks that we haven't started `task1` when we schedule `task2`?

brdji: That's correct. The time interval between requesting the two runs  is small enough that the supervisor does not yet start `task1`.

The supervisor sleeps whenever the queue is empty:
https://github.com/testground/testground/blob/775a58896b01063ae7d9fef0afbb4faf7a5f2bf5/pkg/engine/supervisor.go#L58

```
tsk, err := e.queue.Pop()
if err == task.ErrQueueEmpty {
	time.Sleep(time.Second)
	continue
}
```
(I'm not pleased with the way this test is setup at the moment, and we should consider making the sleep window configurable)

pkg/task/storage.go
```diff
@@ -169,16 +170,19 @@ func (s *Storage) changePrefix(dst string, src string, id string) error {
...
+		logging.S().Errorf("error when trying to put the new key: '%s', err: %v", newkey, err)
 		trans.Discard()
 		return err
 	}
 	err = trans.Delete(oldkey, &opt.WriteOptions{Sync: true})
 	if err != nil {
+		logging.S().Errorf("error when trying to drop the old key: '%s', err: %v", oldkey, err)
 		trans.Discard()
 		return err
 	}
```
laurentsenta, line 188: Let's drop this, that's code from https://github.com/testground/testground/issues/1418#issuecomment-1203819103

laurentsenta:
@brdji is this ready for merge? Could you drop the leveldb debug code and merge?

integration_tests/19_limit_runs_per_branch.sh
laurentsenta: 👍 That might be flaky, let's merge as is but double-check the CI when it fails.

laurentsenta (approved):

PR #1405, opened 2022-07-27
Add support for marking runners disabled ( #1236)

### LINKED ISSUE #1236: Configure TaaS to not allow any tasks other than `cluster:k8s`
At the moment we allow all kinds of jobs on TaaS. This creates various infrastructure issues, as it starts testplans on infra nodes.

We should block the Testground daemon from starting any tasks that don't use the `cluster:k8s` runner, i.e. it should not be possible for users to start `local:docker` tasks on TaaS, as all containers related to that test run would be start on a single VMs, where the daemon is running.

brdji:
Fixes #1236 .

There is also a PR in the infra repo: https://github.com/testground/testground/pull/1405

### DIFF
--- a/integration_tests/18_block_runners.sh
+++ b/integration_tests/18_block_runners.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+# Test for https://github.com/testground/testground/issues/1236
+my_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# block the daemon from starting automatically
+AUTO_START=1
+source "$my_dir/header.sh"
+
+# start it manually with the preset env file, where the local:docker runner is blocked
+start_daemon "$my_dir/18_env.toml"
+
+testground plan import --from ./plans --name testground
+
+pushd $TEMPDIR
+
+testground healthcheck --runner local:docker --fix
+
+# Run test
+testground run single \
+    --plan=testground/placebo \
+    --testcase=ok \
+    --builder=docker:go \
+    --runner=local:docker \
+    --instances=2 \
+    --wait | tee run.out
+
+assert_run_outcome_is ./run.out "canceled"
+
+popd
--- a/integration_tests/18_block_runners.sh
+++ b/integration_tests/18_block_runners.sh
@@ -3,7 +3,7 @@
 my_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
 # block the daemon from starting automatically
-AUTO_START=1
+SKIP_AUTO_START=1
 source "$my_dir/header.sh"
 
 # start it manually with the preset env file, where the local:docker runner is blocked
--- a/integration_tests/header.sh
+++ b/integration_tests/header.sh
@@ -63,18 +63,31 @@ function assert_run_output_is_correct {
   test $SIZEOUT -gt 0 && test $SIZEERR -eq 0
 }
 
+# Start the testground daemon, loading the .env file from the first parameter
+function start_daemon {
+  env_file=$1
+
+  mkdir -p ${HOME}/testground
+
+  cp $env_file ${HOME}/testground/.env.toml
+
+  echo "Starting daemon and logging outputs to $TEMPDIR"
+  testground daemon > $TEMPDIR/daemon.out 2>&1 &
+  DAEMONPID=$!
+
+  sleep 2
+
+  echo "Waiting for Testground to launch on 8040..."
+  while ! nc -z localhost 8040; do
+    sleep 1
+  done
+  echo "Testground launched"
+}
+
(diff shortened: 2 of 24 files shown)


pkg/runner/local_docker.go
```diff
@@ -297,6 +299,11 @@ func (r *LocalDockerRunner) Run(ctx context.Context, input *api.RunInput, ow *rp
 		return
 	}
 
+	if cfg.Disabled {
+		err = ErrRunnerDisabled
+		return
+	}
+
```
laurentsenta, line 306: If we're doing the check here, is it possible for a user to send a custom `input.RunnerConfig` and overrides the flag? (see the `mergo.Merge` line above)

Could we check the flag higher up in the stack? Maybe in the `supervisor.go` module instead. It's a generic flag and it'd be safer if future maintainers don't have to remember to support this flag.

brdji: > If we're doing the check here, is it possible for a user to send a custom input.RunnerConfig and overrides the flag? 
You're right, it's possible! The merge will overwrite the default (`env.toml`) config.

> Could we check the flag higher up in the stack? Maybe in the supervisor.go module instead. 
This is the right way to go about it, though the config is a generic map[string]interface, which might not be ideal to check. I'll try to find a way to keep this check clean.

pkg/runner/local_exec.go
```diff
@@ -43,7 +44,8 @@ type LocalExecutableRunner struct {
 }
 
 // LocalExecutableRunnerCfg is the configuration struct for this runner.
-type LocalExecutableRunnerCfg struct{}
+type LocalExecutableRunnerCfg struct {
+}
 
 func (r *LocalExecutableRunner) Healthcheck(ctx context.Context, engine api.Engine, ow *rpc.OutputWriter, fix bool) (*api.HealthcheckReport, error) {
 	r.lk.Lock()
```
laurentsenta, line 51: That looks like noise, could we drop these changes?

pkg/engine/supervisor.go
```diff
@@ -549,6 +549,11 @@ func (e *Engine) doRun(ctx context.Context, id string, input *RunInput, ow *rpc.
 	// 2. Get the env config for the runner.
 	cfg = cfg.Append(e.envcfg.Runners[trunner])
 
+	var flag = cfg[0][config.RunnerDisabledFlag]
```
laurentsenta, line 552: This access pattern is going to be confusing, since we're dealing with a special, global flag, what about making it clear this is not a configuration we're going to compile, but actually a global runner config?

```suggestion
	var flag = e.envcfg.Runners[trunner][config.RunnerDisabledFlag]
```

laurentsenta:
Looks good, added a few notes.

Can we add an integration test? I guess we can use the `docker:swarm` runner as our guinea pig for this change.

brdji:
The latest commit fixes a bug where the runner/builder error does not cause the task to be canceled.

I added an integration test that uses a custom .env.toml file to block the `local:docker` runner. To facilitate this, I touched up the `header.sh` file, so other integration tests can use this in the future.

integration_tests/header.sh
```diff
@@ -63,18 +63,31 @@ function assert_run_output_is_correct {
...
-echo "Waiting for Testground to launch on 8040..."
-while ! nc -z localhost 8040; do
-  sleep 1
-done
-echo "Testground launched"
+# if the AUTO_START flag is unset or set to 0, start the daemon immediately
+if [[ ! -n "$AUTO_START" || $AUTO_START -eq 0 ]]; then
+  echo "Starting daemon automatically"
+  start_daemon 'env-kind.toml'
+fi
```
laurentsenta, line 93: 👍 nice

integration_tests/header.sh
```diff
@@ -63,18 +63,31 @@ function assert_run_output_is_correct {
...
 
-sleep 2
+  echo "Waiting for Testground to launch on 8040..."
+  while ! nc -z localhost 8040; do
+    sleep 1
+  done
+  echo "Testground launched"
+}
+
+TEMPDIR=`mktemp -d`
```
laurentsenta, line 87: `start_daemon` depends on this variable, we should define it above it.

integration_tests/18_block_runners.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+# Test for https://github.com/testground/testground/issues/1236
+my_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# block the daemon from starting automatically
+AUTO_START=1
```
laurentsenta, line 6: Usually, `0` is false-ish and `1` is truth-ish,
So when you write `AUTO_START=1` I will assume you mean "please auto start".

Let's use `SKIP_AUTO_START`.

Issue #1382, opened 2022-07-11
Testground  reports failure for instances when they appear to succeed


lalexgap:
#### Version information:
```
❯ git log | head -n1
commit 42b4355f7760d29100d971df1c94739d1b796450
```
Macbook pro (M1)

#### Description:

Testground reports that instances have failed `run finished with outcome = failure (single:0/2)` even though it looks like they have succeeded:

```
Jul 11 19:20:04.210869	INFO	2.0365s         OK << single[001] (95aa87) >>
Jul 11 19:20:04.214827	INFO	2.0369s         OK << single[000] (acfc2b) >>
```
 
I've noticed this when running my own tests but I'm also able to reproduce it by running the `ping-pong` sample test case. It doesn't happen all the time but does happen quite frequently.

```
testground run single --plan=network --testcase=ping-pong \
                        --builder=docker:go --runner=local:docker \
                        --instances=2 --wait
```

Testground collect output:
[cb67fo7semvi2emk8gog.tgz](https://github.com/testground/testground/files/9087002/cb67fo7semvi2emk8gog.tgz)

Full console output:
```
❯ ❯ testground run single --plan=network --testcase=ping-pong \
                        --builder=docker:go --runner=local:docker \
                        --instances=2 --wait
Jul 11 19:20:00.744322	INFO	created a synthetic composition file for this job; all instances will run under singleton group "single"
Jul 11 19:20:00.744352	INFO	using home directory: /Users/lagap/testground
Jul 11 19:20:00.744406	INFO	no .env.toml found at /Users/lagap/testground/.env.toml; running with defaults
Jul 11 19:20:00.744419	INFO	testground client initialized	{"addr": "http://localhost:8042"}

>>> Result:

Jul 11 19:20:00.750255	INFO	run is queued with ID: cb67fo7semvi2emk8gog

>>> Server output:

Jul 11 19:20:01.189398	INFO	performing build for groups	{"plan": "network", "groups": ["single"], "builder": "docker:go"}
Jul 11 19:20:01.222146	INFO	found existing volume	{"volume_name": "testground-goproxy-vol"}
Jul 11 19:20:01.256548	INFO	container is already running	{"container_name": "testground-goproxy"}
Step 1/29 : ARG BUILD_BASE_IMAGE
Step 2/29 : ARG RUNTIME_IMAGE=busybox:1.31.1-glibc
Step 3/29 : FROM ${BUILD_BASE_IMAGE} AS builder
 ---> 67f476ddb57c
[...]

Issue #1322, opened 2022-04-22
Every test run on local docker ends with a `write error`


laurentsenta:
#### Version information:
Using current master `08bdb03a8736218dfb24d5ec18b3a0416dba7434` on Intel

#### Description:

When I run a test, on a local machine, it ends with a write error.

- No matter the test outcome,
- With `single` run or composition alike.

```
Apr 22 13:22:17.746058  INFO    2.5546s INCOMPLETE << sidecar      >>
2022-04-22 15:22:17.746058 +0200 CEST m=+25.705691152 write error: write /Users/laurent/testground/data/daemon/c9hanvtvqc7i6rcov5tg.out: file already closed
```

Example of command:

```
testground run single --plan=network --testcase=ping-pong --builder=docker:go --runner=local:docker --instances=2
```

laurentsenta:
### Notes

The `testground/data/daemon/runId.out` is a jsonlines file of `{time, content (base64)}` logs.
It's the file we render when you visit http://localhost:8042/logs?task_id=c9hanvtvqc7i6rcov5tg (click on log in the dashboard)

Disabling the `f.Close()` here, solves the error:
https://github.com/testground/testground/blob/08bdb03a8736218dfb24d5ec18b3a0416dba7434/pkg/engine/supervisor.go#L99-L105

This `supervisor::worker` is the thing that receives tasks (build & run) and processes them.

If I log the close operation:

```
Apr 22 13:56:47.328678  INFO    worker completed task   {"worker_id": 1, "task_id": "c9hb875vqc7jrtcdemf0"}
Closing the file!
Closed the file! Error: <nil>
Apr 22 13:56:47.328829  INFO    2.4439s INCOMPLETE << sidecar      >>
2022-04-22 15:56:47.328829 +0200 CEST m=+39.359414630 write error: write /Users/laurent/testground/data/daemon/c9hb875vqc7jrtcdemf0.out: file already closed
```

So the close operation is not the problem, it's an async write. It's probably the sidecar, if you don't close the file, you get:

```
Apr 22 14:12:54.951379  INFO    run finished successfully       {"run_id": "c9hbfotvqc7le7mk1vpg", "plan": "network", "case": "ping-pong", "runner": "local:docker", "instances": 2}
Apr 22 14:12:54.952080  INFO    worker completed task   {"worker_id": 1, "task_id": "c9hbfotvqc7le7mk1vpg"}
(not) Closing the file!
[...]

PR #1406, opened 2022-07-27
Fix: testground instability due to race condition on outcomes

laurentsenta:
When the docker runner starts a test run it does a few things:
1. Wait for the instance containers to complete,
2. Setup a "pretty printer" that forwards instance's stdout, BUT ALSO tries to decode events from stdout and update the outcomes, 
3. Listen to the events sent by test instances and update the outcomes on success.

When the containers exits (1) testground stops everything related to collecting the outcomes (3).
This is fine-ish with the go-sdk because it dumps events to stdout as well, so the outcomes will be read thanks to (2).

Logging events on stdout and trying to decode the JSON looked like a legacy hack: this was not implemented in the rust-sdk. See my [notes on the pretty printer](https://github.com/testground/testground/issues/1322#issuecomment-1141209321). The rust-sdk triggers a race condition:

1. test instance sends "success event" to the sync service
2. test instance exits
3. Testgound shuts down the test (and the outcome collection part)
4.  Testground receives the "success event", but does not listen to it anymore.

I believe we've seen this error on go-sdk, less often, see [this comment](https://github.com/testground/testground/issues/1382#issuecomment-1193405557).

This PR rework the local docker runner so that a test is complete if and only if:
- All the instances have exited AND all the outcomes have been received
- OR the test timeout has been reached.

See runs:

- rust-sdk examples with regular testground: https://github.com/laurentsenta/testground-stability/actions/runs/2747857836
- rust-sdk examples with this patch: https://github.com/laurentsenta/testground-stability/actions/runs/2747849325

## Todo

- [x] Spike the patch,
- [ ] Verify that the patch doesn't break other components,
- [ ] Go through all other TODOs (dealing with errors, etc),
- [ ] Double check the outcome outputs and the integration tests results,
- [ ] Look into the testing situation.

### DIFF
--- a/pkg/runner/cluster_k8s.go
+++ b/pkg/runner/cluster_k8s.go
@@ -184,7 +184,7 @@ func (c *ClusterK8sRunner) Run(ctx context.Context, input *api.RunInput, ow *rpc
 		return nil, fmt.Errorf("could not init pool: %w", err)
 	}
 
-	result := newResult()
+	result := newResult(input)
 	runoutput = &api.RunOutput{
 		RunID:  input.RunID,
 		Result: result,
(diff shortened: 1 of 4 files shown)


laurentsenta:
Closing in favor of https://github.com/testground/testground/pull/1407 (moved to the testground repo to get CircleCI's feedbacks).

Issue #1408, opened 2022-08-01
System docker images are amd64 only. Runs in emulation on Mac M1 (arm64)


cskiraly:
#### Version information:
Version `a068f0b` on MacBook Pro M1 (arm64)

#### Description:
Many of the Docker images used are amd64 only.

<img width="510" alt="image" src="https://user-images.githubusercontent.com/1183261/182130887-c9da054b-80d4-4622-a49c-efaf3bf4e2a8.png">

#### Solution
I have tested rebuilding all of these for arm64 (results pushed to my repo).

##### testground-sync-service
```
git clone https://github.com/testground/sync-service.git
cd sync-service
docker build . -t iptestground/sync-service:latest
```
This works as an override, replaces the version testground would use.

##### testground-goproxy
```
git clone https://github.com/goproxyio/goproxy.git
cd goproxy
make
docker build . -t goproxy/goproxy:latest
```
Testground pulls anyway, replacing this. I have pushed to my Dockerhub and replaced image source in code.

##### testground-grafana
bitnami/grafana is used, which in not multi-arch. I have replaced it with grafana/grafana, which has multi-arch support.

Result
<img width="505" alt="image" src="https://user-images.githubusercontent.com/1183261/182132086-29ba0136-8d9f-4ae8-8b46-9774106f0e55.png">

PR #1409, opened 2022-08-01
use Grafana docker image from Grafana instead of Bitnami

cskiraly:
The Grafana provided image has multi-arch support, while the Bitnami version does not.

This is partial solution to #1408.

### DIFF
--- a/pkg/runner/local_common.go
+++ b/pkg/runner/local_common.go
@@ -34,7 +34,7 @@ func localCommonHealthcheck(ctx context.Context, hh *healthcheck.Helper, cli *cl
 		healthcheck.StartContainer(ctx, ow, cli, &docker.EnsureContainerOpts{
 			ContainerName: "testground-grafana",
 			ContainerConfig: &container.Config{
-				Image: "bitnami/grafana",
+				Image: "grafana/grafana",
 			},
 			HostConfig: &container.HostConfig{
 				PortBindings: exposed,


brdji:
I don't know what differences between the two images we should pay attention to, but I agree with @laurentsenta  - we need to be careful not to introduce differences or issues by switching these images.
@cskiraly Have you tried building these changes on an `amd64` machine, or just `arm64`?

cskiraly:
I have not digged into it yet too much, so take it as a discussion opener. I've mostly documented it to help others running into the same, as part of trying to move all images to arm64.

Definitely needs some cross-checking before migrating, if it's worth migrating. I've looked at the provisional images from bitnami mentioned in the thread linked by @laurentsenta at https://github.com/ZCube/bitnami-compat, but it seems there is no Grafana there at the moment.

cskiraly:
The Grafana Dockerfile is at https://github.com/grafana/grafana/blob/main/Dockerfile
- 2-stage,
- Alpine based,
- smaller than the bitnami (but of course no overlap in base layers with other bitnami images),
- comes from the Grafana developers.

I'm not saying with this that it is OK to change, just checked and sharing what I have seen.

Bitnami has the base layer, but not the things on top. They would need to 
- change the OS_ARCH sting in the Dockerfile (https://github.com/bitnami/containers/blob/main/bitnami/grafana/9/debian-11/Dockerfile)
- cache the arm64 tar.gz files
- update all the hashes in the Dockerfile
- and test .....
I suppose it will take some time.

Having said that, the amd64 grafana image seems to run on the M1 in emulation, it is just slower.

laurentsenta:
Thanks for the clear details and the contribution @cskiraly,

If you can use the `bitnami/grafana` image with emulation, I'd like to wait until we have a staging EKS cluster available to test this change:
- I'm not familiar with either of these images and I'd need to run tests to make an educated decision,
- The main risk is losing some cloud-related support which might bite us in a couple of weeks/months when we got our EKS system ready.

@brdji no action to take on your side regarding this, but do you agree?

brdji:
Agreed. We can test these images on the new EKS cluster, once we have a staging version ready.

cskiraly:
I agree, it is a good idea to delay this until it can be tested on the cluster. No urgency as emulation seems to work and manual change is easy based on this PR (see diff) if needed by someone.

laurentsenta:
(marking as blocked until we have a staging EKS setup to test this with)

PR #1421, opened 2022-08-05
feat: various usability & stabiliy improvements for interop

laurentsenta:
- [x] Introduce `split` operation in composition templates
- [x] Add a `load_resource` operation in composition templates
  - This lets a user define configuration in a file (like a list of instances) and use them during composition templating.
- [x] drop arbitrary build timeout (we have a task timeout)
- [x] drop instance count requirements when we can deduce it
- [x] provide composition templating during `build` (it was only applied during `run` before)
- [x] Add a concurrency option to build artifacts one by one
  - When working on rust interop tests, I noticed we lose the benefits of docker caching & explodes storage requirements because we start every build at the same time.

### DIFF
--- a/pkg/build/docker_generic.go
+++ b/pkg/build/docker_generic.go
@@ -47,9 +47,6 @@ func (b *DockerGenericBuilder) Build(ctx context.Context, in *api.BuildInput, ow
 		return nil, err
 	}
 
-	ctx, cancel := context.WithTimeout(ctx, 30*time.Minute)
-	defer cancel()
-
 	planPath := cfg.Path
 	basePathForPlan := path.Join("/plan", planPath)
 
--- a/pkg/cmd/fixtures/templates/with-resource-complex.toml
+++ b/pkg/cmd/fixtures/templates/with-resource-complex.toml
@@ -1,0 +1,48 @@
+[metadata]
+  name = "experiment"
+
+[global]
+  plan = "plan"
+  case = "case"
+  builder = "docker:go"
+  runner = "local:docker"
+
+{{ with (load_resource "./resource-complex.toml" ) }}
+{{- range .groups }}
+[[groups]]
+  id = "{{ .id }}"
+  instances = { count = 1 }
+
+  [groups.build]
+    selectors = ['{{ .selector }}']
+
+  [groups.build_config]
+    build_base_image = 'golang:{{ .go_version }}-buster'
+    modfile = "{{ .modfile }}"
+{{ end }}
+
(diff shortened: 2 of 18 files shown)


laurentsenta:
Marking for review,
Note each commit is one feature, I think having a single PR will save us time, but I can extract 1 PR per commit if needed.

pkg/cmd/fixtures/templates/with-resource-complex.toml
```diff
@@ -0,0 +1,48 @@
...
+{{- range .groups }}
+[[groups]]
+  id = "{{ .id }}"
+  instances = { count = 1 }
+
+  [groups.build]
+    selectors = ['{{ .selector }}']
+
+  [groups.build_config]
+    build_base_image = 'golang:{{ .go_version }}-buster'
```
brdji, line 20: This looks very unwieldy and prone to typos. I would prefer the much cleaner option of using something like `go_build_image=golang:1.18-buster` in the `resource.toml` file itself.

pkg/build/docker_generic.go
```diff
@@ -47,9 +47,6 @@ func (b *DockerGenericBuilder) Build(ctx context.Context, in *api.BuildInput, ow
 		return nil, err
 	}
 
-	ctx, cancel := context.WithTimeout(ctx, 30*time.Minute)
```
brdji, line 50: Do we have an integration test that checks the task timeout?

laurentsenta: Not really, here we're dropping a 30 minutes timeout: it's a rather trivial change, and the only integration test we could write here is to wait for 30 minutes in the CI. The benefits / cost ratio would be very low.

pkg/cmd/fixtures/templates/with-resource-complex.toml
laurentsenta: So the discussion is whether the benefit of reusing a single variable (go_version) in multiple locations is worth the complexity it adds to the template definition. That might be fine in some cases, and not fine in other cases.

But that's something a Testground User might worry about.

We're wearing a Testground Maintainer hat here: we're not using or maintaining this composition file, we're only testing the templating feature with a few fixtures that looks like real tests. I believe the test would be as good or worst if we dropped this part because it would cover a different, smaller, use case.

brdji: Understood. Feel free to close this.

brdji (approved):

Issue #1392, opened 2022-07-20
EPIC: Testground `remote:docker` and `remote:exec` runners


laurentsenta:
Currently, testground has 3 runners:
- `local:exec` which "just" starts a binary on the machine,
- `local:docker` which relies on docker containers running on the machine,
- `cluster:k8s` which relies on Kubernetes.

One of Testground promises (to be clarified) is to let a user run a test instance on a "mock network". You might use these feature to simulate custom topologies, network latency, etc (see https://docs.testground.ai/concepts-and-architecture/sidecar).

But:

- it might be too difficult to hold that promise in 100% of cases (bugs such as https://github.com/testground/testground/issues/1385),
- we might also want to run tests going through the "real" internet (benchmarks, ProbeLab, etc),
- and we might want to compare testground results on a mocked network with testground results on real networks ("testing the testground").

One solution might be to add `remote` runners to testground which would let you (the user) run a test by ssh'ing on different machines you control and use the internet and real networks as the test.

## Steps

- [ ] Clarify what we need from `remote:exec`, `remote:docker`,
- [ ] Define an architecture (are we using simple ssh, docker swarm, etc),
- [ ] Plan and implement.

Issue #1342, opened 2022-05-30
Create PR of documentation changes


KatarinaKeti:
## What defines this endeavour to be complete?

<!--Create a check list of what you expect to see as part of delivering on the ask--

- [ ] Item 1
- [ ] Item 2
- [ ] ...

dektech:
PRs already merged:
- https://github.com/testground/docs/pull/13
- https://github.com/testground/infra/pull/71
- https://github.com/testground/testground/pull/1375

PR #1425, opened 2022-08-19
feat: Add Nix Builder + Add remote runner example plan

MarcoPolo:
This PR does a couple things:

1. Introduces a Nix builder that lets us build binaries, containers, buildpacks and OSs for multiple architectures. (you can actually define what the OS looks like and deploy it as part of the test!)
  a. Adds an integration test (that will fail until we add Nix to Circle CI)
2. Tweaks `local:exec` to allow you to run any executable, not just Go binaries.
3. Adds a new example plan that demonstrates how to support remote runners with `ssh` and the Nix builder. It leverages Nix to build for different architectures and efficiently/correctly deploy the application to the remote machine – even if the application is dynamically linked (caveat that the remote machine must have Nix installed).
  a. This plan tests a simple server/client ping across machines.
  

In the long term we'll probably want some sort of first class support for remote runners in testground, but I think this is a usable solution in the short term that will let us set up some reproducible benchmark environments. But I may be missing something in my understanding of how Testground works, so apologies if that's the case.

r? @laurentsenta


p-shahi:
@laurentsenta and I discussed this PR today and what it would enable for the libp2p maintainers.

We specifically [talked about this proposal](https://github.com/libp2p/go-libp2p/pull/1804#issuecomment-1271995750)
>> That being said. I can add to the benchmark milestone and move F.1 to Q4 (or early Q1): wdyt? F.1 - Benchmarking using nix-builders (dependency: remote machines need Nix installed) F.2 - Benchmarking using first class support for remote runners (using remote:exec) in Testground

>This sound good. I think F.1 will give us >80% of the benefit of F.2. And also inform what F.2 should look like. Also worth noting that we can do F.1 _today_. We aren't blocked on anything (except time and prioritization).

For that reason, this PR has been prioritized and added to both the test-plans and go-libp2p roadmaps (for Q2 November) to bootstrap the go-libp2p benchmarking efforts.
It will also enable running [js-libp2p-transfer-performance](https://github.com/ipfs-shipyard/js-libp2p-transfer-performance) as a test-plan and CI job to benchmark transfer times across releases (to catch issues like [#1342](https://github.com/libp2p/js-libp2p/issues/1342))

The questions we had were: 
 - If this PR achieves 80% of what we need, what is the remaining 20% that needs to be addressed afterwards?[^1]

A question Laurent had with regard to the Testground team's project prioritization was:
 - Is this PR enough to unblock the libp2p teams, and allow the Testground maintainers to prioritize the set of network simulation features (used currently in CI) to do RTT testing as a part of [multi-dimenstional testing](https://github.com/testground/testground/issues/1392) and deprioritize https://github.com/testground/testground/issues/1392 [^2]

cc: @BigLep @tinytb 

[^1]: My thoughts were, I think we will likely have a better idea of the "20% gap" after merging and using this in Testground. It largely accomplishes what `remote:exec` is trying to do ["we build a binary, scp to the remote machine, and start the binary there"](https://www.notion.so/pl-strflt/Remote-Runners-c4ad4886c4294fb6a6f8afd9c0c5b73c#cd6d237911d148f18fb5620e5ce1d04f)
[^2]: I think yes it's enough to allow the Testground team to focus on the network simulation features.

plans/remote-runner/README.md
```diff
@@ -0,0 +1,67 @@
...
+the application on different machines. It's a prototype, so expect this to
+change and to be better supported in the future.
+
+At a high level, this works by:
+1. Using [Nix] to build the application for each machine's architecture
+   (e.g. One node may be `x86_64-linux` and another may be `aarch64-darwin`).
+   You may need to set up [remote
+   builders](https://nixos.org/manual/nix/stable/advanced-topics/distributed-builds.html)
+   to properly cross-build the application.
+1. Using a `local:exec` runner to run a shell script that copies the built
```
p-shahi, line 13:
```suggestion
2. Using a `local:exec` runner to run a shell script that copies the built
```
Numbering is wrong.

Also, would it be better define a new `remote:exec` runner?

MarcoPolo:
@galargh friendly ping on this

galargh:
> @galargh friendly ping on this

It's on my TODO list for this week. I'll have it reviewed on Wednesday the latest - hopefully, tomorrow. Sorry for the delay.

plans/remote-runner/README.md
galargh: If you keep using `1.`s for list items then GitHub Markdown creates a numbered list starting from `1`. That's where the numbering comes from.

As for the name, `local:exec` is an existing construct. Also, the name is fitting here because the shell script that performs the copy is executed locally.

pkg/build/nix_generic.go
```diff
@@ -0,0 +1,79 @@
...
+// Nix. The resulting artifact can be a binary or a container.
+type NixBuilder struct{}
+
+type NixBuilderConfig struct {
+	// AttributeToBuild is the attribute we're building with `nix build .#<attr>`.
+	// If empty, this will build the default package. e.g. `nix build .`
+	AttributeToBuild string `toml:"attr"`
+}
+
+// Build builds a testplan written in Go and outputs an executable.
```
galargh, line 29: Am I right in saying it builds whatever `flake.nix` specifies so it doesn't necessarily have to be Go?

pkg/build/nix_generic.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/testground/testground/pkg/api"
+	"github.com/testground/testground/pkg/rpc"
+)
+
+var (
+	_ api.Builder = &NixBuilder{}
+)
+
+// NixBuilder (id: "nix:generic") is a builder that build the test plan using
+// Nix. The resulting artifact can be a binary or a container.
```
galargh, line 20: Do we care about distinguishing between binary/container artifacts? Should we flag it somehow?

pkg/build/nix_generic.go
```diff
@@ -0,0 +1,79 @@
...
+	cmd.Stderr = &stderr
+	cmd.Stdout = &stdout
+	err := cmd.Run()
+	if err != nil {
+		return nil, fmt.Errorf("failed to build: %v. %w", string(stderr.Bytes()), err)
+	}
+
+	decoder := json.NewDecoder(&stdout)
+	type nixBuildOutput struct {
+		DrvPath string
```
galargh, line 58: What's this for?

pkg/runner/local_exec.go
```diff
@@ -44,7 +45,11 @@ type LocalExecutableRunner struct {
 }
 
 // LocalExecutableRunnerCfg is the configuration struct for this runner.
-type LocalExecutableRunnerCfg struct{}
+type LocalExecutableRunnerCfg struct {
+	// BinPath is the path of the binary relative to the artifact path. Use this
+	// if the builder doesn't put the binary as the artifact path
+	BinPath string `toml:"bin_path"`
```
galargh, line 51: Just to confirm I understand this. We need this for nix. It outputs an artifact. We pass the artifact path to local exec runner. We want to execute a specific binary within that artifact. Is it correct?

galargh (approved):
Looks good :) We'll probably want to update docs at some point but I think we can hold of for a bit and see how we end up using it in practice.

I left a few questions here and there, mostly to make sure I understand it correctly.

🚀

galargh:
@laurentsenta What was the trick for `go mod tidy` not to look into directories with different go versions? It seems that's the problem here.

laurentsenta:
@galargh I used a `_` prefix,

https://pkg.go.dev/cmd/go#hdr-Package_lists_and_patterns

> Directory and file names that begin with "." or "_" are ignored by the go tool, as are directories named "testdata".

@MarcoPolo is this still priority if we plan on moving the building out of testground in `libp2p/test-plans`?

MarcoPolo:
Closing this. I think we want to get out of the building game. And I think we need to make testground simpler to enable remote workers properly.

PR #1437, opened 2022-09-01
#1293: Engine cleans up processing tasks on startup

### LINKED ISSUE #1293: Clean-up all in-progress or completed containers on Testground daemon startup
`testground daemon` is responsible for tracking the life-cycle of testplan runs on TaaS (via the `cluster:k8s` runner).

At the moment if the daemon crashes for various reasons (OOM for example), it doesn't clean up in-progress testplan runs, which can result in containers lingering forever as `completed` in Kubernetes.

TODO
- [ ] make sure that `testground daemon` cleans up the Kubernetes namespace from in-progress / crashed / completed testplan runs on startup

brdji:
Fixes #1293 .

The engine will clean up tasks in the `Processing` state on startup. I've been trying to come up with the best way to limit the use of this function to the daemon startup:
 - Make the function private, and call it in the `New`, when the engine is created. I don't like adding too many steps to initialization / constructors, so I went with option two:
 - Make the function public, and call it from the daemon, when the engine is created. 
 
 As a safeguard (to prevent undesired side-effects), the engine will skip tasks for which a signal channel exists - these tasks are being actively processed, and should not be deleted.
 
Please let me if my reasoning is correct, thank you!

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -424,6 +425,40 @@ func (e *Engine) Kill(id string) error {
 	e.signalsLk.RUnlock()
 
 	return nil
+}
+
+// CleanUpTasks deletes all tasks in the store that are currently being processed
+func (e *Engine) CleanUpTasks() error {
+
+	allTasks, err := e.store.AllTasks()
+	if err != nil {
+		return err
+	}
+
+	deleted := 0
+	for _, aTask := range allTasks {
+		if aTask.State().State == task.StateProcessing {
+			// If the engine has a signal for the task, that means it's being tracked actively
+			if e.hasSignal(aTask.ID) {
+				logging.S().Debugw("Skipping cleanup of task, as it is still active", "task", aTask.ID, "state", aTask.State())
+				continue
+			}
+
+			logging.S().Infow("Cleaning up inactive task", "task", aTask.ID, "type", aTask.Name())
+			err = e.store.Delete(aTask.ID)
+			if err != nil {
+				return errors.New("error cleaning up task in progress")
+			}
+			deleted++
+		}
+	}
+
+	if deleted > 0 {
+		logging.S().Infof("Cleaned up %d inactive (in progress) tasks", deleted)
+	}
+
+	return nil
+
 }
 
 // UnmarshalTask converts the given byte array into a valid task
--- a/pkg/task/storage.go
+++ b/pkg/task/storage.go
@@ -230,6 +230,29 @@ func (s *Storage) rangeIter(prefix string, start time.Time, end time.Time) (task
 	return tasks, nil
 }
 
+// AllTasks returns all tasks in the store
+func (s *Storage) AllTasks() (tasks []*Task, err error) {
+	rng := util.Range{
+		Start: nil, Limit: nil,
+	}
+
+	tasks = make([]*Task, 0)
+
+	iter := s.db.NewIterator(&rng, nil)
+	defer iter.Release()
+
+	for iter.Next() {
+		tsk := &Task{}
+
+		err := json.Unmarshal(iter.Value(), tsk)
+		if err != nil {
+			return nil, err
+		}
+		tasks = append(tasks, tsk)
+	}
+	return tasks, nil
+}
+
 func NewMemoryTaskStorage() (*Storage, error) {
 	inmem := storage.NewMemStorage()
 	db, err := leveldb.Open(inmem, nil)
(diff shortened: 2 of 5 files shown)


laurentsenta:
I need to look into the internal to familiarize myself with the API but the approach looks good to me, thanks for sharing 👍

pkg/task/storage.go
```diff
@@ -230,6 +230,29 @@ func (s *Storage) rangeIter(prefix string, start time.Time, end time.Time) (task
...
+
+		err := json.Unmarshal(iter.Value(), tsk)
+		if err != nil {
+			return nil, err
+		}
+		tasks = append(tasks, tsk)
+	}
+	return tasks, nil
+}
+
```
laurentsenta, line 255: Do you know if this will contain ALL the tasks the cluster ran since its creation?
Nothing to fix now, but if that's the case we should plan for starting time issues on long-lived clusters.

brdji: Yes, that's correct - it will return all tasks present in the DB. However, I don't see this slowing down the startup time even when there are thousands of runs, unless the DB driver is very inefficient.
In either case, if we notice performance issues, we can refine this to retrieve the tasks from the past X days, instead of all tasks.

pkg/engine/engine.go
```diff
@@ -426,6 +427,40 @@ func (e *Engine) Kill(id string) error {
...
+	for _, aTask := range allTasks {
+		if aTask.State().State == task.StateProcessing {
+			// If the engine has a signal for the task, that means it's being tracked actively
+			if e.hasSignal(aTask.ID) {
+				logging.S().Debugw("Skipping cleanup of task, as it is still active", "task", aTask.ID, "state", aTask.State())
+				continue
+			}
+
+			logging.S().Infow("Cleaning up inactive task", "task", aTask.ID, "type", aTask.Name())
+			err = e.store.Delete(aTask.ID)
```
laurentsenta, line 448: Sorry for not replying earlier, I haven't had time to look into the storage part.

https://github.com/testground/testground/issues/1293 is talking about,
> containers lingering forever as completed in Kubernetes.

And the task is
> cleans up the Kubernetes namespace from in-progress / crashed / completed testplan runs on startup

But it looks like this PR is just removing tasks logs from Testground's leveldb database.
We don't want to remove logs from the history, we want to clean lingering resources.

This is the sort of code I expect to see:
https://github.com/testground/testground/blob/3400b0e10b58ba772f3a1a1c37f2c73fbc0e1972/pkg/runner/cluster_k8s.go#L346-L357

- Could you make sure we're not removing any tasks, but at least marking them as cancelled?
- Could you confirm this PR is cleaning up the Kubernetes namespace from in-progress / crashed / completed testplan on startup?

laurentsenta (changes requested):
I don't think this is fixing the issue, please use the thread I created above.

pkg/engine/engine.go
brdji: Thanks for the explanation! I think you're right - I'll take some time to look through the code and test a solution on the new cluster.

Since it's been a while, I seem to remember there being a reason why I removed them from the DB - I'll see if I can dig it up.

brdji:
Blocked by #1350 (since I have used the EKS cluster to test the solution).

I have tested a working solution for the cleanup, but two issues I have not been able to solve satisfactorily:
- I couldn't figure out the correct place in the code where to slot in the function. During startup, the `cluster_k8s` runner can throw an error if it attempts to initialize the pool before the sync-service has started. For a temporary solution, I have set the method to sleep for 10 seconds before attempting the cleanup - and it works.
- I couldn't determine how to parse a task's input into the expected `ClusterK8sRunnerConfig` type. For certain tasks, the cast operation would fail, indicating that the task's input is actually of type `map[string] interface{}`.

I would appreciate someone with more knowledge than me to assist me in finding the solution.

pkg/engine/engine.go
```diff
@@ -426,6 +428,67 @@ func (e *Engine) Kill(id string) error {
...
+						return fmt.Errorf("error deleting pod %s: %v", pod.Name, err)
+					}
+				}
+			}
+		}
+	}
+
+	return nil
+
+}
```
laurentsenta, line 490: That code looks very specific to kubernetes, we should stick to the `engine` abstraction and have something like

```
for every scheduled tasks:
   runner[task.runner].cleanup()
```

Then implement the cleanup in each runner (`docker` and `k8s`)

laurentsenta:
@brdji thanks for sharing this PR, I think some more work and testing is required, is it a PR you still intend to bring to completion?

brdji:
> @brdji thanks for sharing this PR, I think some more work and testing is required, is it a PR you still intend to bring to completion?

Unfortunately, I am unable to work on this task any further. We can close this PR, and keep our discussion and my work so far as a reference point on what needs to be done to solve this issue.

PR #1441, opened 2022-09-07
feat: add .testgroundignore file

### LINKED ISSUE #1170: Add a .testgroundignore kind-of feature
## Description

We should add a `.testgroundignore` kind of feature (similar to `.gitignore`). I came across this issue while building the `docker/node` builder (#1158). For Node.js projects, the `node_modules` directory is usually quite big, and uploading it to the daemon every time we need to run a test doesn't seem to be the wisest option.

Thus, I suggest creating an ignore list, in the format of a `.testgroundignore` file.

## What defines this endeavour to be complete?

- [ ] Support `.testgroundignore`

laurentsenta:
Fixes #1170


laurentsenta:
@hacdias 👋  just tagging you if you want to follow along, but no input is required from you.

pkg/client/client.go
```diff
@@ -221,6 +235,45 @@ func (c *Client) runBuild(ctx context.Context, r interface{}, path, plandir, sdk
...
+// getFilteredDirectory filters the directory dir according to the
+// ignored files specified in $dir/.testgroundignore. Returns a new
+// directory and a boolean indicating if such directory is temporary
+// and should be cleaned by the caller.
+func getFilteredDirectory(dir string) (string, bool, error) {
+	ignoreFilePath := filepath.Join(dir, ".testgroundignore")
+	_, err := os.Stat(ignoreFilePath)
+
+	if os.IsNotExist(err) {
+		return dir, false, nil
```
galargh, line 247: Don't we want to create a temp dir either way?

Is it possible that the dir files get modified during the run? If so, then if we always create a temp dir, no matter if `.testgroundignore` is provided or not, the behaviour is consistent and the file modifications are not reflected after the run.

pkg/client/client.go
```diff
@@ -221,6 +235,45 @@ func (c *Client) runBuild(ctx context.Context, r interface{}, path, plandir, sdk
...
+	} else if err != nil {
+		return "", false, err
+	}
+
+	tgIgnore, err := ignore.CompileIgnoreFile(ignoreFilePath)
```
galargh, line 252: Could we have a comment that explains that the output of `CompileIgnoreFile` is a set of regex patterns (as per https://github.com/sabhiram/go-gitignore/blob/master/ignore.go#L164)?

It wasn't immediately obvious to me and without this information it's hard to judge if the comparisons we're doing later are correct.

pkg/client/client.go
galargh: We also wouldn't need the temp boolean anymore if we always create a temp dir.

pkg/client/fixtures/with_ignore/.testgroundignore
```diff
@@ -0,0 +1,3 @@
+a
+b/c
```
galargh, line 2:
```suggestion
c
```

I think it'd be interesting to test if the above works. If we just have `b/c` here, then a naive implementation that doesn't turn lines here into regexes would still work but if we have only `c` then it wouldn't anymore. Does it make sense?

galargh (commented):
Cool change :) Left a few comments here and there but I like it a lot :)

pkg/client/client.go
brdji: I agree with @galargh  - I would like to see a more comments detailing this function.

laurentsenta: Updated

pkg/client/client.go
```diff
@@ -18,6 +18,9 @@ import (
 	"strings"
 	"sync"
 
+	copy2 "github.com/otiai10/copy"
```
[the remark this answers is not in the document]
laurentsenta: Probably a left over, I removed it

pkg/client/client.go
laurentsenta: Sounds good, I liked that first approach, but dropping it make sense, fixed.

pkg/client/fixtures/with_ignore/.testgroundignore
laurentsenta: Agreed, I updated the test!

pkg/client/client.go
laurentsenta: Added a comment with some clarification.

.github/workflows/integration-tests.yml
```diff
@@ -37,4 +36,4 @@ jobs:
           make install
       - name: Run tests
         run: |
-          go test -tags "integration,${{ matrix.suite }}" -timeout 20m -parallel 1 -v -shuffle=on ./pkg/integration/...
```
galargh, line 40: Why this change? At face value, I'd read it as an indication of the test suite getting worse.

galargh (approved):

.github/workflows/integration-tests.yml
laurentsenta: Agreed, but we're using 1.16 for now, and that option was added with go 1.17, it's messing up with the tests,
I added a comment to clarify this.

Issue #1368, opened 2022-06-28
EPIC: Speed up development with improved integration testing tooling (no more shell scripts or flakiness)


laurentsenta:
eta: 2023-01

It looks like we have 2 ways to run integration tests:

1. https://github.com/testground/testground/tree/master/pkg/cmd/itest
2. https://github.com/testground/testground/tree/master/integration_tests

Some of these are testing the same commands.
The `itest` package makes `go test -v ./pkg/...` is annoying to use because integration tests are creating a lot of noise amongts unit tests.

We should merge all these tests to a single, clearer location.

A few notes:

- Using go instead of shell make sense here, the [header.sh](https://github.com/testground/testground/blob/master/integration_tests/header.sh) script is already quite complex.
- We need to take into account the fact that some integration tests require setup (k8s)
- Bonus point for parallelizing tests, we might be able to use a github action's matrix.

## Design

- [x] Review approaches
     - [x] https://github.com/kubernetes/community/blob/master/contributors/devel/sig-testing/e2e-tests.md

## Tasks

- [x] Create a integration_tests package that relies on [golang's testing tools]
    - [x] Implement the tooling required to make these tests maintainable(https://github.com/testground/testground/tree/master/pkg/cmd/itest)
    - [x] Move these tests to a github CI action, ideally with a matrix per environment (docker, k8s, exec, etc), or if possible: per test since they are quite long.
- [x] port `test-integ-local-exec`
    - [x] Port the shell scripts to go
- [ ] port `test-integ-local-docker`
    - [ ] Port the shell scripts to go
- [x] port `test-integ-examples`
    - [x] Port the shell scripts to go

laurentsenta:
@galargh we had a few out-of-band discussions on this, do I remember correctly that from your experience, it'll be MUCH easier to move our test to go unit tests rather than trying things like the sharness library?

galargh:
> @galargh we had a few out-of-band discussions on this, do I remember correctly that from your experience, it'll be MUCH easier to move our test to go unit tests rather than trying things like the sharness library?

Let's definitely stay away from sharness 👅

PR #1442, opened 2022-09-07
#1432 If timeout is reached, mark task as canceled, and stop containers.

### LINKED ISSUE #1432: Delete tasks that have timeouted
## Description

Tasks that have reached the timeout should be killed, removed from the queue, and updated in the database.
Quote from @laurentsenta :
```
There is a task timeout for every test run in testground, when a task reaches this timeout we should kill it and update the state. Look for task_timeout_min in the codebase.
```

## What defines this endeavor to be complete

- [ ] Remove timeouted tasks from the queue
- [ ] Stop timeouted task execution (e.g. stopping the docker container)
- [ ] Update the state of the task in the database

Linked to #1417

bokimilinkovic:
Fix #1432 
When timeout is reached, task is removed from the queue, marked as canceled and stored back to DB.
But still there is some misbehaviours regarding task state that I noticed.. should be discussed further more in details..
I guess that `worker()` function could be a bit improved to handle that state overrides

### DIFF
--- a/pkg/engine/engine_test.go
+++ b/pkg/engine/engine_test.go
@@ -2,6 +2,13 @@ package engine
 
 import (
 	"encoding/json"
+	"github.com/otiai10/copy"
+	"github.com/rs/xid"
+	"github.com/stretchr/testify/require"
+	"github.com/testground/testground/pkg/config"
+	"io/ioutil"
+	"os"
+	"path/filepath"
 	"reflect"
 	"testing"
 	"time"
@@ -95,4 +102,157 @@ func TestUnmarshalTaskBuild(t *testing.T) {
 	if !reflect.DeepEqual(newTask, taskData) {
 		t.Errorf("Unmarshal Build task returned incorrect data")
 	}
+}
+
+func TestNewEngine(t *testing.T) {
+	env := &config.EnvConfig{
+		Daemon: config.DaemonConfig{
+			Scheduler: config.SchedulerConfig{
+				TaskRepoType:   "memory",
+				Workers:        1,
+				TaskTimeoutMin: 1,
+			},
+		},
+		Runners: map[string]config.ConfigMap{
+			"local:exec": {},
+		},
+		Builders: map[string]config.ConfigMap{
+			"exec:go": {},
+		},
+	}
+	err := env.Load()
+	require.NoError(t, err)
+
+	cfg := &EngineConfig{
+		Builders:  AllBuilders,
+		Runners:   AllRunners,
+		EnvConfig: env,
+	}
+
+	engine, err := NewEngine(cfg)
+	require.NoError(t, err)
+
+	t.Run("OK case", func(t *testing.T) {
+		testTask := createTestTask(t, "ok")
+		require.NoError(t, engine.queue.Push(testTask))
+		time.Sleep(30 * time.Second) // need to give it a time to finish because it is asynchronous function.
+		// get the task back from the database and check final state
+		taskFromDB, err := engine.store.Get(testTask.ID)
+		require.NoError(t, err)
+		require.Equal(t, task.StateComplete, taskFromDB.State().State)
+	})
+
+	t.Run("Stall case", func(t *testing.T) {
+		testTask := createTestTask(t, "stall")
+		require.NoError(t, engine.queue.Push(testTask))
+		time.Sleep(70 * time.Second) // need to give it a time to finish because it is asynchronous function.
+		// get the task back from the database and check final state
+		taskFromDB, err := engine.store.Get(testTask.ID)
+		require.NoError(t, err)
+		require.Equal(t, task.StateCanceled, taskFromDB.State().State)
+	})
+}
+
+// testcase could be one of these: 'ok' and 'stall'
+// OK will execute right away and stall will cause deadline exceed.
+func createTestTask(t *testing.T, testcase string) *task.Task {
+	cmp := api.Composition{
+		Global: api.Global{
+			Builder:        "exec:go",
+			Plan:           "placebo",
+			Case:           testcase,
+			TotalInstances: 1,
+			BuildConfig: map[string]interface{}{
+				"go_proxy_mode": "direct",
+			},
+			Runner: "local:exec",
+			RunConfig: map[string]interface{}{
+				"enabled": true,
+			},
+		},
+		Groups: []*api.Group{
+			{
+				ID:        "test",
+				Instances: api.Instances{Count: 1},
+			},
+		},
+	}
+
+	manifest := api.TestPlanManifest{
+		Name: "placebo",
+		Builders: map[string]config.ConfigMap{
+			"exec:go": {},
+		},
+		Runners: map[string]config.ConfigMap{
+			"local:exec": {},
+		},
+		TestCases: []*api.TestCase{
+			{
+				Name:      testcase,
+				Instances: api.InstanceConstraints{Minimum: 1, Maximum: 100},
+				Parameters: map[string]api.Parameter{
+					"param4": {
+						Type:    "string",
+						Default: "value4:default:manifest",
+					},
+				},
+			},
+		},
+	}
+	basedir, err := ioutil.TempDir("", "")
+	if err != nil {
+		return nil
+	}
+
+	plandir := filepath.Join(basedir, "plan")
+	err = copy.Copy("../../plans/placebo", plandir)
+	if err != nil {
+		return nil
+	}
+	t.Cleanup(func() {
+		os.RemoveAll(basedir)
+	})
+
+	id := xid.New().String()
+	taskData := &task.Task{
+		Type:        task.TypeRun,
+		Version:     0,
+		Priority:    0,
+		Plan:        "placebo",
+		Case:        testcase,
+		ID:          id,
+		Runner:      "local:exec",
+		Composition: cmp,
+		Input: RunInput{
+			RunRequest: &api.RunRequest{
+				CreatedBy:   api.CreatedBy{User: "test-user"},
+				BuildGroups: []int{0},
+				Composition: cmp,
+				Manifest:    manifest,
+			},
+
+			Sources: &api.UnpackedSources{BaseDir: basedir, PlanDir: plandir},
+		},
+		States: []task.DatedState{
+			{
+				State:   task.StateScheduled,
+				Created: time.Now().UTC(),
+			},
+		},
+		CreatedBy: task.CreatedBy(api.CreatedBy{User: "test-user"}),
+	}
+
+	jsonData, err := json.Marshal(taskData)
+
+	if err != nil {
+		return nil
+	}
+
+	newTask, err := UnmarshalTask(jsonData)
+
+	if err != nil {
+		return nil
+	}
+
+	return newTask
 }
--- a/pkg/engine/supervisor.go
+++ b/pkg/engine/supervisor.go
@@ -70,13 +70,41 @@ func (e *Engine) worker(n int) {
 
 			ch := make(chan int)
 			e.addSignal(tsk.ID, ch)
-
 			go func() {
 				select {
 				case <-ch:
 					e.deleteSignal(tsk.ID)
 					cancel()
 				case <-ctx.Done():
+					// Timeout reached,
+					// 1. Remove timeouted tasks from the queue -- It is going to be removed anyway, since we are popping from the queue.
+					// 2. Stop timeouted task execution (e.g. stopping the docker container)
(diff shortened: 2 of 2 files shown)


pkg/engine/supervisor.go
```diff
@@ -70,13 +70,42 @@ func (e *Engine) worker(n int) {
...
+						err = e.store.PersistProcessing(tsk)
+						if err != nil {
+							logging.S().Errorw("could not persist task", "err", err)
+							return
+						}
+
+						if tsk.IsCanceled() {
+							logging.S().Infow("Task reached timeout", "task_id", tsk.ID)
+						}
+					}
```
laurentsenta, line 108: Where do we stop the containers?

bokimilinkovic: As I understood, when the signal is deleted  from	`signals   map[string]chan int`  runner will stop all containers linked with that task.ID, correct me if I am wrong?

pkg/engine/engine_test.go
```diff
@@ -96,3 +103,156 @@ func TestUnmarshalTaskBuild(t *testing.T) {
...
+		require.NoError(t, engine.queue.Push(testTask))
+		time.Sleep(70 * time.Second) // need to give it a time to finish because it is asynchronous function.
+		// get the task back from the database and check final state
+		taskFromDB, err := engine.store.Get(testTask.ID)
+		require.NoError(t, err)
+		require.Equal(t, task.StateCanceled, taskFromDB.State().State)
+	})
+}
+
+// testcase could be one of these: 'ok' and 'stall'
```
galargh, line 156: Shouldn't this be an enum then?

pkg/engine/engine_test.go
```diff
@@ -96,3 +103,156 @@ func TestUnmarshalTaskBuild(t *testing.T) {
...
+// OK will execute right away and stall will cause deadline exceed.
```
galargh, line 157: Could you please point me to where that logic is implemented? I cannot seem to find it on my own.

galargh (commented):
The new test is currently not executed by the CI, could you add it to https://github.com/testground/testground/blob/6c348603a61ad4c83f1f524d8dfb9c05b989b799/Makefile#L59?

pkg/engine/supervisor.go
bokimilinkovic: I guess the same way that Kill also works

```
// Kill closes the signal channel for a given task, which signals to the runner to stop it
func (e *Engine) Kill(id string) error {
	e.signalsLk.RLock()
	if ch, ok := e.signals[id]; ok {
		close(ch)
	}
	e.signalsLk.RUnlock()

	return nil
}
```

pkg/engine/supervisor.go
```diff
@@ -70,13 +70,42 @@ func (e *Engine) worker(n int) {
...
 				case <-ch:
 					e.deleteSignal(tsk.ID)
 					cancel()
 				case <-ctx.Done():
+					// Timeout reached,
+					// 1. Remove timeouted tasks from the queue -- It is going to be removed anyway, since we are popping from the queue.
+					// 2. Stop timeouted task execution (e.g. stopping the docker container)
+					e.deleteSignal(tsk.ID)
+					// 3.  Update the state of the task in the database
+					logging.S().Infow("Timeout reached for task: ", "task_id", tsk.ID)
```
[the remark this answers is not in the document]
bokimilinkovic: I am updating database for the given task

```
// if task is still in processing state, mark it as canceled in DB
					if tsk.State().State == task.StateProcessing {
						canceledState := task.DatedState{
							Created: time.Now().UTC(),
							State:   task.StateCanceled,
						}
						tsk.States = append(tsk.States, canceledState)
						logging.S().Infow("Task state: ", "state", tsk.State().State)

						err = e.store.PersistProcessing(tsk)
						if err != nil {
							logging.S().Errorw("could not persist task", "err", err)
							return
						}
		.....
	}
```

pkg/engine/supervisor.go
```diff
@@ -70,13 +70,42 @@ func (e *Engine) worker(n int) {
...
+					// get the current state of the task from DB.
+					tsk, err := e.store.Get(tsk.ID)
+					if err != nil {
+						return
+					}
+					// if task is still in processing state, mark it as canceled in DB
```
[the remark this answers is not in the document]
bokimilinkovic: I also thought  about it, but I guess why it is not implemented that way firstly is that it doesn't mean that task is finished when the container is stopped, also there can be multiple containers for one task, so that's why the update is done from the same 'hierarchy' level as task starting is.

pkg/engine/supervisor.go
bokimilinkovic: Although I see that there is method over client that can stop container, I can not get to call that method from the right place where context is canceled, since 'engine' is not having that client as the dependency 

```
func (cli *Client) ContainerStop(ctx context.Context, containerID string, timeout *time.Duration) error {
   query := url.Values{}
   if timeout != nil {
      query.Set("t", timetypes.DurationToSecondsString(*timeout))
   }
   resp, err := cli.post(ctx, "/containers/"+containerID+"/stop", query, nil, nil)
   ensureReaderClosed(resp)
   return err
}
```

plans/_integrations/manifest.toml
```diff
@@ -7,9 +7,18 @@ runner = "local:docker"
 [builders."docker:go"]
 enabled = true
 
+[daemon.scheduler]
+task_timeout_min = 1
+task_repo_type = "disk"
+
 [runners."local:docker"]
 enabled = true
+run_timeout_min = 1
```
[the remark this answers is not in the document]
bokimilinkovic: You are right, we don't need `run_timeout_min` 👍 We can have integration tests longer that 1 minute, 1 minute is just minimum time to be set for daemon to wait, by default it is 10

plans/placebo/go.mod
```diff
@@ -3,3 +3,5 @@ module github.com/testground/testground/plans/placebo
 go 1.16
 
 require github.com/testground/sdk-go v0.3.1-0.20211012114808-49c90fa75405
+
+replace golang.org/x/sys => golang.org/x/sys v0.0.0-20220704084225-05e143d24a9e
```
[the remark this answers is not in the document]
bokimilinkovic: I was keep getting this error while executing test

```
go build failed: go: github.com/testground/sdk-go@v0.3.1-0.20211012114808-49c90fa75405 requires
	github.com/prometheus/client_golang@v1.7.1 requires
	golang.org/x/sys@v0.0.0-20200615200032-f1bc736245b1: missing go.sum entry; to add it:
	go mod download golang.org/x/sys
```

After that replace line, it was fixed and ci/cd tests were passing

pkg/engine/engine_test.go
bokimilinkovic: https://github.com/testground/testground/blob/master/plans/placebo/main.go
It is implemented here, I was just passing right test case as one of this from the map:
```
var testcases = map[string]interface{}{
	"ok":    ok,
	"panic": panickingTest,
	"stall": stall,
}
```
Ok returns no error
and Stall is sleeping for 24hrs

pkg/engine/supervisor.go
laurentsenta: I can't find that command in the codebase, https://github.com/testground/testground/search?q=ContainerStop,
It does look like an API, that's an entry point that the daemon exposes to clients, we don't want to use this internally.

Could you share links to the code? We might not be working on this part of the codebase when we review, and it's taking us time to move around the codebase even though you already did it.

You can also share permalink: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-a-permanent-link-to-a-code-snippet

Like this one: https://github.com/testground/testground/blob/63a8fce8cf08d576e745186dc322312ccf3af97c/pkg/engine/engine.go#L419-L427

The code you shared closes the channel, but are we sure closing this channel will kill the containers/pods/process in docker/k8s/exec ?

bokimilinkovic: I am providing you screenshots of docker containers running step by step

1. testground daemon
Screenshot 2022-09-22 at 14 11 24

2. running test case` testground run single --plan integrations --testcase issue-1432-task-timeout  --builder docker:go --runner local:docker --instances 1`
Screenshot 2022-09-22 at 14 12 07
Screenshot 2022-09-22 at 14 12 53

3. Task timeouted after 1 minute



Screenshot 2022-09-22 at 14 13 18

Screenshot 2022-09-22 at 14 13 32
Screenshot 2022-09-22 at 14 13 43

bokimilinkovic: Parent context is being used to create new context with cancel
https://github.com/testground/testground/blob/bb4451ee7bf0f1542278eabbdd5559667ca223a4/pkg/runner/local_docker.go#L486
that is used for starting containers:
`err := cli.ContainerStart(startGroupCtx, c.containerID, types.ContainerStartOptions{})`

this defered function is stopping running containers:
https://github.com/testground/testground/blob/bb4451ee7bf0f1542278eabbdd5559667ca223a4/pkg/runner/local_docker.go#L458

@laurentsenta

pkg/engine/engine_test.go
```diff
@@ -96,3 +110,156 @@ func TestUnmarshalTaskBuild(t *testing.T) {
...
+	}
+
+	newTask, err := UnmarshalTask(jsonData)
+
+	if err != nil {
+		return nil
+	}
+
+	return newTask
+}
```
laurentsenta, line 265: Testground's integration testing is quite messy at the moment, see https://github.com/testground/testground/issues/1368 for references.

While we wait to focus on the integration testing story,
Shell scripts have the benefit of introducing less legacy code (this is ~100+ lines of hardcoded configuration that already exists in the `plan` folder). Could you move this test to integration tests?

In the end, it looks like we are "just" testing the outcome:

```
testground run placebo:ok test
assert task status === "canceled"

testground run placebo:ok test
assert task status === "canceled"
```

plans/placebo/go.mod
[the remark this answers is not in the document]
laurentsenta: I'm surprised we need a `replace` here too,
could you make sure you're using the latest master?
Could you remove this line and let the CI run?

pkg/engine/supervisor.go
laurentsenta: I'll need more time to review this, could you fix the other requests? Thanks in advance, I'll try to have a second pass tomorrow.

laurentsenta (changes requested):
- Please remove the replace so that we can see the error in CI
- Please fix the other remarks regarding naming, etc.

pkg/engine/engine_test.go
```diff
@@ -96,3 +110,156 @@ func TestUnmarshalTaskBuild(t *testing.T) {
 		t.Errorf("Unmarshal Build task returned incorrect data")
 	}
 }
+
+func TestNewEngine(t *testing.T) {
```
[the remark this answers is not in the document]
bokimilinkovic: We are testing task timeout, but it can only be invoked upon calling constructor of Engine, because that is when it is popping tasks from the queue.. this was firstly intended to be integration test, but I would remove it totally to avoid confusion, since integration test is executed from the script, as you told me so.

bokimilinkovic:
https://app.circleci.com/pipelines/github/testground/testground/1512/workflows/53d587c7-11d2-4b9c-b57c-cda7a75bee73/jobs/6338/parallel-runs/0/steps/0-104
Failing go mod tidy for placebo plan

pkg/engine/engine_test.go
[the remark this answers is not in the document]
brdji: I would remove the tests, as well. The integration test covers the desired usage.

laurentsenta:
@bokimilinkovic, sorry I missed the fact that you updated this PR, could you add a note to the comments that should be resolved now?

Do you intend to bring this PR to completion?
In the description, there is a note that this is not the complete fix:

> But still there is some misbehaviours regarding task state that I noticed.. should be discussed further more in details..

Could you expand on that?

bokimilinkovic:
Hello @laurentsenta 
I tend to finish this PR, It was well tested and seems that tasks are properly updated when timeout-ed and containers are stopped. 
Regarding:
`But still there is some misbehaviours regarding task state that I noticed.. should be discussed further more in details..`
It was my mistake because I was trying to change engine's configuration from manifest, not testground home dir.
I will merge master into my branch and resolve conflicts.
You can take a one more look (or run 1 timeout test case if you have time to be sure that it works properly), and if everything seems fine so we can merge it

PR #1445, opened 2022-09-08
feat: introduce changelog

laurentsenta:


CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [0.6.0] - unreleased
```
galargh, line 7: Shouldn't this be just:
```suggestion
## [Unreleased]
```

As per the Keep a Changelog example. Then, we could analyse the changelog during the release to determine what version change is suitable.

Btw, we use similar changelog in a bunch of IPDX repos already, https://github.com/ipfs/download-ipfs-distribution-action/blob/master/CHANGELOG.md or https://github.com/pl-strflt/rust-sccache-action/blob/main/CHANGELOG.md

brdji (approved):

CHANGELOG.md
mxinden: I don't have a strong opinion on this.

In the projects I maintain, I require the unreleased version to be tracked while new pull requests come in. E.g.

- First pull request is a patch, thus adding an unreleased section with the patch version ( semver below `1.0.0`) bumped.
- Second pull request is a breaking change, thus updating the version in the unreleased section, bumping the minor (again semver below `1.0.0`) and undoing the patch version bump.
- Third pull request is a breaking change as well, though due to the second pull request, no changes beyond adding a changelog line are required.
- ...

My understanding of a patch is most accurate when I do the final review, thus I track the semver version change at this point in time, not later on when releasing.

mxinden (approved):
Very much in favor of keeping a changelog and keepachangelog.com.

CHANGELOG.md
galargh: Actually, that sounds like a better approach. Less likely to miss something during the release process.

BigLep:
@laurentsenta : what are the next steps before merging this?

laurentsenta:
@BigLep I'll merge tomorrow and start asking new PRs to contribute to the Changelog.

Issue #1416, opened 2022-08-03
EKS: Run the ping-pong test without making functional changes


brdji:
## Description
While running benchmarks on the EKS trial cluster, we have discovered that the `ping-pong` test cannot pass. It turns out that the test attempts to reach one pod from another, and it waits for this operation to finish and hangs indefinitely.

This is happening due to predefined sets of IP addresses that are allocated to running pods, which are not matched properly to the EKS setup/installation. We need to find a way of resolving the issue without impacting the basic functionality that the test offers.

## What defines this endeavor to be complete 
We should minimally modify the ping-pong test, so that it does not rely on pods receiving IP addresses that are in an expected range (e.g. from 16.0.0.0 to 16.0.0.1). In other words, the pods should somehow communicate their IP addresses, and not assume them beforehand.

The output of the command
`testground run single --plan network --testcase ping-pong --instances 2 --builder docker:go --runner cluster:k8s`
Should result in a `successful` job result.

One possible solution (using the sync service) would be:
- [x] Modify the ping-pong test so that the listening pod broadcasts its listening IP address to the sync-service, and the sending pod receives the address from the sync-service
- [x] Verify the test is working successfully on the EKS test cluster
- [x] Verify the test is working for the `local:docker` runner, as well

laurentsenta:
Does it depend on the CNI used?
Is this related to Testground and its sidecar or is it an issue that we can reproduce outside of Testground?
Do we have a plan? I guess something like:

- [ ] Reproduce the issue outside of Testground
- [ ] Create a minimal test case & share here
- [ ] Try with different combination of CNIs
  - [ ] CNI1-CNIA: pass / fail
  - [ ] CNI2-CNIA: pass / fail
  - [ ] CNI1-CNIB: pass / fail

dektech:
This is the plan, to verify whether it also happens on other CNI combinations.

Also, could you please clarify what does `reproducing the issue outside of Testground` mean? My understanding was that this is a TG-exclusive test that only works inside an environment running the daemon and sidecar.

dektech:
This has been happening on both `calico + weave` and  ` aws vpc cni + weave` clusters. Currently working on finding the exact issue and where is it happening. 
As for the last question, I really cannot answer that, maybe @brdji can but he is away until next week. 
You can try and let us know of the results.

Here is some additional info about the issue (tested on AWS VPC CNI + weave):

Command used to invoke the test:
[...]

PR #1462, opened 2022-09-15
Exchange IPs in pingpong test

brdji:
Instances should not guess each other's address, because the allocated IP range may not match the code.
This PR forces the test instances to exchange their data IPs, so that they will attempt to connect to the correct IP.

Fixes one part of #1416 .

### DIFF
--- a/plans/network/pingpong.go
+++ b/plans/network/pingpong.go
@@ -4,13 +4,21 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"strings"
 	"time"
 
 	"github.com/testground/sdk-go/network"
+	"github.com/testground/sdk-go/ptypes"
 	"github.com/testground/sdk-go/run"
 	"github.com/testground/sdk-go/runtime"
 	"github.com/testground/sdk-go/sync"
 )
+
+type ListenAddrs struct {
+	Addrs []string
+}
+
+var PeerTopic = sync.NewTopic("peers", &ListenAddrs{})
 
 func pingpong(runenv *runtime.RunEnv, initCtx *run.InitContext) error {
 	ctx, cancel := context.WithTimeout(context.Background(), 300*time.Second)
@@ -43,14 +51,14 @@ func pingpong(runenv *runtime.RunEnv, initCtx *run.InitContext) error {
 	runenv.RecordMessage("before netclient.MustConfigureNetwork")
 	netclient.MustConfigureNetwork(ctx, config)
 
-	seq := client.MustSignalAndWait(ctx, "ip-allocation", runenv.TestInstanceCount)
-
 	// Make sure that the IP addresses don't change unless we request it.
 	if newAddrs, err := net.InterfaceAddrs(); err != nil {
 		return err
 	} else if !sameAddrs(oldAddrs, newAddrs) {
 		return fmt.Errorf("interfaces changed")
 	}
+
+	seq := client.MustSignalAndWait(ctx, "ip-allocation", runenv.TestInstanceCount)
 
 	runenv.RecordMessage("I am %d", seq)
 
@@ -59,7 +67,7 @@ func pingpong(runenv *runtime.RunEnv, initCtx *run.InitContext) error {
 
 	config.IPv4 = runenv.TestSubnet
 	config.IPv4.IP = append(config.IPv4.IP[0:2:2], ipC, ipD)
-	config.IPv4.Mask = []byte{255, 255, 255, 0}
+	config.IPv4.Mask = []byte{255, 254, 0, 0}
 	config.CallbackState = "ip-changed"
 
 	var (
@@ -78,12 +86,21 @@ func pingpong(runenv *runtime.RunEnv, initCtx *run.InitContext) error {
 	runenv.RecordMessage("before reconfiguring network")
 	netclient.MustConfigureNetwork(ctx, config)
 
+	addrs, err := exchangeListenAddrs(ctx, runenv, client)
+	if err != nil {
+		return err
+	}
+
 	switch seq {
 	case 1:
+		runenv.RecordMessage("Listening at  ", listener.Addr())
 		conn, err = listener.AcceptTCP()
 	case 2:
+		addr := strings.Split(addrs.Addrs[0], ":")[0]
+		var targetIp = net.ParseIP(addr)
+		runenv.RecordMessage("Attempting to connect to ", targetIp)
 		conn, err = net.DialTCP("tcp4", nil, &net.TCPAddr{
-			IP:   append(config.IPv4.IP[:3:3], 1),
+			IP:   targetIp,
 			Port: 1234,
 		})
 	default:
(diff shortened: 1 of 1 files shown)


plans/network/pingpong.go
```diff
@@ -43,23 +51,23 @@ func pingpong(runenv *runtime.RunEnv, initCtx *run.InitContext) error {
...
+
 	runenv.RecordMessage("I am %d", seq)
 
 	ipC := byte((seq >> 8) + 1)
 	ipD := byte(seq)
 
 	config.IPv4 = runenv.TestSubnet
 	config.IPv4.IP = append(config.IPv4.IP[0:2:2], ipC, ipD)
-	config.IPv4.Mask = []byte{255, 255, 255, 0}
+	config.IPv4.Mask = []byte{255, 254, 0, 0}
```
galargh, line 70: Why are we using this mask? Not saying that there's anything wrong with that - I just genuinely don't know.

plans/network/pingpong.go
```diff
@@ -187,6 +204,92 @@ func pingpong(runenv *runtime.RunEnv, initCtx *run.InitContext) error {
...
+				fmt.Printf("%s does not contain %s\n", subnet, ip.IP)
+			}
+		} else {
+			panic(fmt.Sprintf("%T", addr))
+		}
+	}
+	return nil, fmt.Errorf("no network interface found. Addrs: %v", addrs)
+}
+
+// Shares all of this instance's addresses with all other instances in the test
```
galargh, line 267: I don't think I quite understand this. What other addresses than https://github.com/testground/testground/blob/330ca290d6aa44d06b0b39268744910e35924871/plans/network/pingpong.go#L210 are we sharing?

brdji: Nice catch - I'll rewrite this so that only a single address (data IP) gets shared.

plans/network/pingpong.go
brdji: The previous mask (/24) was too restrictive, and instances could not connect to each other over their assigned IP when running the test on EKS. Since the test subnet has a mask of /12 (e.g. 10.32.0.0/12), instances could receive an IP that would be excluded by the /24 mask, for example 10.33.0.0.

If we reduce the mask too much (e.g. /8), then the test will fail locally, as well.
By reducing the mask to the current value (/15), the test will work both locally and on the new EKS cluster.

galargh: Would you mind adding a comment about this here? I feel like this is a sort of information that could be easily forgotten.

plans/network/pingpong.go
galargh: Hey, thanks for making the updates. Sorry I wasn't clearer on what I meant - I had something like this in mind: https://github.com/testground/testground/pull/1470 (feel free to merge if you like it; I targeted your PR there) - i.e. taking advantage of the fact that we only ever need to publish a single string, not a list of address structs.

plans/network/pingpong.go
laurentsenta: Bumping this, sorry if this is silly, I'm catching up with this networking wizardry:

Our EKS Test Subnet might be something like `10.32.0.0/12`.
When we called this code before, it was possible that an instance received the address `10.33.0.1`. While it expected something with the subnet mask `/24`, which means:

- the parameter above `config.IPv4.IP = append(config.IPv4.IP[0:2:2], ipC, ipD)` is not really used in the case of EKS, correct? We get a different IP.
- somehow, that would break this instance results because having to reach an address outside its subnet somehow "broke" the instance, correct?

Now we're going to use a mask `/15`, which means:

- the instance might get assigned the address `10.40.0.01`, which is within the network defined by `10.32.0.0/12` but outside `10.32.0.0/15`.

Is this correct? Am I missing something?
Why do we need a subnet different from the one passed in `TestSubnet`?

plans/network/pingpong.go
```diff
@@ -187,6 +206,96 @@ func pingpong(runenv *runtime.RunEnv, initCtx *run.InitContext) error {
...
+	// since there is only 1 other instance in test, we expect a single address in the list
+	if len(otherAddrs) > 1 {
+		return "", fmt.Errorf("expected a single address to be returned, got %d", len(otherAddrs))
+	}
+
+	return otherAddrs[0], nil
+}
+
+// Returns the IP address belonging to the given subnet
+func getSubnetAddr(subnet *ptypes.IPNet) (*net.TCPAddr, error) {
```
laurentsenta, line 253: Could you use the SDK's `GetDataNetworkIP` ?

plans/network/pingpong.go
```diff
@@ -187,6 +206,96 @@ func pingpong(runenv *runtime.RunEnv, initCtx *run.InitContext) error {
...
+	}
+
+	_ = client.MustSignalAndWait(ctx, sync.State("got-other-addrs"), runenv.TestInstanceCount)
+
+	// since there is only 1 other instance in test, we expect a single address in the list
+	if len(otherAddrs) > 1 {
+		return "", fmt.Errorf("expected a single address to be returned, got %d", len(otherAddrs))
+	}
+
+	return otherAddrs[0], nil
```
laurentsenta, line 249: It looks like this function assumes a single `tcpAddr` from `getSubnetAddr` and at the same time acts like we're moving around arrays of addresses.

Could we be explicit about exchanging a single address? That would simplify the code a lot.

plans/network/pingpong.go
```diff
@@ -187,6 +206,96 @@ func pingpong(runenv *runtime.RunEnv, initCtx *run.InitContext) error {
...
+	if err != nil {
+		return "", err
+	}
+
+	otherAddrs := []string{}
+	// filter addresses to remove any own address
+	for _, addr := range allAddrs {
+		if _, ok := mine[addr]; ok {
+			continue
+		}
```
laurentsenta, line 238: See the comment below, if you have a single address, can you go for `if addr = tcpAddr`?

plans/network/pingpong.go
```diff
@@ -187,6 +206,96 @@ func pingpong(runenv *runtime.RunEnv, initCtx *run.InitContext) error {
...
+		return "", err
+	}
+
+	mynode := &ListenAddrs{}
+	mine := map[string]struct{}{}
+
+	mynode.Addrs = append(mynode.Addrs, tcpAddr.String())
+	for _, l := range mynode.Addrs {
+		mine[l] = struct{}{}
+	}
```
laurentsenta, line 224: See the comment below, if you have a single address, can you get rid of these constructions?

plans/network/pingpong.go
galargh: That's exactly my thinking - https://github.com/testground/testground/pull/1462#discussion_r976896605

laurentsenta: (resolving in favor of your comment @galargh, thank you o7)

plans/network/pingpong.go
```diff
@@ -43,23 +51,25 @@ func pingpong(runenv *runtime.RunEnv, initCtx *run.InitContext) error {
...
-	seq := client.MustSignalAndWait(ctx, "ip-allocation", runenv.TestInstanceCount)
-
 	// Make sure that the IP addresses don't change unless we request it.
 	if newAddrs, err := net.InterfaceAddrs(); err != nil {
 		return err
 	} else if !sameAddrs(oldAddrs, newAddrs) {
 		return fmt.Errorf("interfaces changed")
 	}
 
+	seq := client.MustSignalAndWait(ctx, "ip-allocation", runenv.TestInstanceCount)
```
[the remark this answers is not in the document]
brdji: To me, it makes more sense to first check if the network interfaces have changed, before moving onto the next logical step (acquiring a sequence number).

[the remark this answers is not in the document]
galargh: Do you know what could cause the addresses to change? I just want to make sure it's not something that can occur during wait for `ip-allocation`.

plans/network/pingpong.go
brdji: Oh, now I understand - thanks for making the PR, I merged it!

plans/network/pingpong.go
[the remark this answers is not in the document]
brdji: The check refers to line 52, and possible side-effects:
`netclient.MustConfigureNetwork(ctx, config)`
Should configure the network according to the specs in the `config` struct, but it should not change the data IP address, because a new address is not specified in the config. Therefore, if the addresses change, the network config func has most likely encountered an error and the test should stop.

The `ip-allocation` wait is handled by the sync-service, and does not change the network interfaces of the test instance; that's why it made sense for me to switch the order of the code.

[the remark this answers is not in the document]
galargh: Cool! Thanks for the explanation :)

galargh (approved):

plans/network/pingpong.go
galargh: @brdji This hasn't been resolved yet 🚨

brdji: Sorry - I missed the comment before merging;

>somehow, that would break this instance results because having to reach an address outside its subnet somehow "broke" the instance, correct?

Correct  - the mask should match the data subnet range, otherwise such an issue can pop up.

>Now we're going to use a mask /15

We've noticed this discrepancy in the other PR - the mask is set to `/12`.

laurentsenta: You're answering all my questions,

We're going to use a mask `/15`, which means:

- the instance might get assigned the address `10.40.0.01`, which is within the network defined by `10.32.0.0/12` but outside `10.32.0.0/15`.

Is this correct? Am I missing something?


> We don't need a different subnet, and they should actually match up. We've noticed this discrepancy in the other PR and fixed it (for EKS) - the mask is set to /12. Note the comment here: [Pingpong.go](https://github.com/testground/testground/blob/5611a78f4cdfb91026321bea0dfca47e37c08ec3/plans/network/pingpong.go).

What stops us from using the env value, passed by testground, instead of hardcoding it here?

brdji: > Is this correct? Am I missing something?

You are correct; that's why I mentioned the fix for the EKS PR. We noticed this discrepancy later (after I closed this PR, before responding).

>What stops us from using the env value, passed by testground, instead of hardcoding it here?

Nothing is stopping us, except that the env value must be present (and I'm not certain its present in every runner). Do you wish for me to change the test so that it uses the env value, if possible?

laurentsenta: Thanks for fixing this in the next PR,

> Nothing is stopping us, except that the env value must be present (and I'm not certain its present in every runner). Do you wish for me to change the test so that it uses the env value, if possible?

That looks like a better option but let's not introduce changes without being certain they are working,
I created a follow-up issue, https://github.com/testground/testground/issues/1486, and added a few notes here. No need to work on this now.

PR #1469, opened 2022-09-21
fix: go dependencies rewrite in exec:go

adlrocha:
I was trying to run a composition that used different dependencies for the different instances using `exec:go`. I was setting the instances' dependencies in the `composition.toml` as follows: 
```toml
[[groups]]
  id = "attacker"
  [groups.build]

  [[groups.build.dependencies]]
       module = "github.com/filecoin-project/lotus"
       version = "8f48631fc48773821fd853b6d9612380b9d74cb4"
```
When running the composition I kept getting the error: `task cancelled due to error     {"err": "task of type run cancelled: unable to add replace directives to go.mod; exit status 1; output: "}`.

Inspecting the code, I realized that the `dependency.Target` had to be set explicitly for it to work with `exec:go`. Modifying the dependency for the instance in the `composition.toml` to the following made it work.
```toml
[[groups]]
  id = "attacker"
  [groups.build]

  [[groups.build.dependencies]]
       module = "github.com/filecoin-project/lotus"
       target = "github.com/filecoin-project/lotus"
       version = "8f48631fc48773821fd853b6d9612380b9d74cb4"
```
However, I was then getting an error when building the assets because after the dependency replacement, we needed a `go mod tidy`. Adding the small command included in this PR completely fixed the issue, and I was able to run the composition with multiple versions seamlessly.

### DIFF
--- a/pkg/build/exec_go.go
+++ b/pkg/build/exec_go.go
@@ -98,7 +98,7 @@ func (b *ExecGoBuilder) Build(ctx context.Context, in *api.BuildInput, ow *rpc.O
 
 	// Calculate the arguments to go build.
 	// go build -o <output_path> [-tags <comma-separated tags>] <exec_pkg>
-	var args = []string{"build", "-gcflags='all=-N -l'", "-o", path}
+	var args = []string{"build", "-gcflags=all=-N -l", "-o", path}
 	if len(in.Selectors) > 0 {
 		args = append(args, "-tags")
 		args = append(args, strings.Join(in.Selectors, ","))
--- a/pkg/build/exec_go.go
+++ b/pkg/build/exec_go.go
@@ -85,7 +85,7 @@ func (b *ExecGoBuilder) Build(ctx context.Context, in *api.BuildInput, ow *rpc.O
 	}
 
 	// go mod tidy
-	cmd := exec.CommandContext(ctx, "go", append([]string{"mod", "tidy"})...)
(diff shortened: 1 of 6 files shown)


GlenDC (approved):

pkg/build/exec_go.go
```diff
@@ -84,6 +84,14 @@ func (b *ExecGoBuilder) Build(ctx context.Context, in *api.BuildInput, ow *rpc.O
 		}
 	}
 
+	// go mod tidy
+	cmd := exec.CommandContext(ctx, "go", append([]string{"mod", "tidy"})...)
```
laurentsenta, line 88:
```suggestion
	cmd := exec.CommandContext(ctx, "go", "mod", "tidy")
```

laurentsenta:
Thanks for the contribution @adlrocha and review @GlenDC,

A few notes:
- I'm looking into CI for your fork,
- If you don't mind I'll update this PR to fix the explicit `Target` issue and add an integration test,

adlrocha:
Sure, edit at will, let me know if you need anything from my side.

laurentsenta (approved):

brdji (changes requested):
Looks good to me, but please add notes to the [changelog](https://github.com/testground/testground/blob/8b80d79fdb503dc7fb15135b73069ba2e487bfd9/CHANGELOG.md)

laurentsenta:
Updated with a CHANGELOG (since I took on finishing this PR) and merging,

Thanks for the contribution @adlrocha !

Issue #1386, opened 2022-07-18
EPIC: implement Javascript & Browser support in Testground


KatarinaKeti:
## Description

Testground supports js test written for node, through the `docker:node` builder (legacy) and `docker:generic`.

We want to use Testgroung to test interoperability between different implementations of libp2p.
A critical need is to test browser-related features such as https://github.com/libp2p/specs/pull/412

## What defines this endeavour to be complete?

- [x] Create an `example-browser` plan with an `helloworld` and `failure`  tests (see steps below),
- [x] Create a more complex test case that does a ping between a browser instance and a node instance, using composition files.
- [ ] Finally, the team should be able to reuse the code to implement the browser ping test in https://github.com/libp2p/test-plans

### **Step 1: Run an helloworld test**

(merged in https://github.com/testground/testground/pull/1474)

- [x] 1.1 Succes Test: create a testground PR with a `plans/example-browser` test 
  - Likely will use the code from @GlenDC's [playwright demo](https://github.com/GlenDC/tg-playwright-test/tree/main/plans/playwright-test)
  - [x] It starts the node process, using docker generic, and we call `yarn run testground`
  - [x] We won't use a server, we'll just start the browser, log an HelloWorld, then record a success
- [x] 1.2 Failure test: create another case where we fail on purpose
  - [x] throw an error on the test.
- [x] 1.3 Turn into integration tests
  - [x] add a new test in `./integration_tests/` (feel free to update this ticket with questions)

### **Step 2: Run the helloworld test from WITHIN the browser runtime**

- [x] Use the js-sdk from within the browser to record message, metrics, etc.
- We should end up with a "wrapper" that loads the browser and pass in the environment. Then the test lives in the browser.
- This "should" work, testground-sync service was [rewritten](https://github.com/testground/sync-service/issues/4) a while ago to support connecting from a browser.
- :warning: we'll need to forward the environment variables for the testground SDK to work
- :warning: we might have issues with loading the js-sdk in the browser (missing dist package, etc)

Work threads:

- sdk-js PR: https://github.com/testground/sdk-js/pull/25
- testground PR: https://github.com/testground/testground/pull/1502

### **Step 3: Create a composition with node + js**

- [x] The high level goal is to create a ping-pong test between a node and a browser node.
    - done in https://github.com/testground/testground/pull/1539#issuecomment-1340703249

## Follow-up and other tasks

Discussion and notes raised during the talk about this feature:

- It would make sense to provide a typescript version of the js-sdk
[...]

PR #1474, opened 2022-09-24
Feat/1386 browser hello world test

GlenDC:
Add browser hello testplan and integration tests, making use of playwright and docker:generic,
in function of issue #1386 

Feature/Requirement checklist:

- [X] hello browser testplan 
- [x] hello browser integration test
    - [x] success integration test
    - [x] failure integration test

Usage example of hello world testplan:

```
testground plan import --from ./plans/example-browser
```

```
testground run single \
    --plan example-browser --testcase success \
    --instances 1 \
    --builder docker:generic --runner local:docker \
    --tp browser=firefox --wait
```

You can run firefox/chromium for both success and failure cases using the single integration test:

```
./integration_tests/example_04_browser.sh
```

~~Secondary Requirement checklist:~~

> Question: should we do that as part of this PR, or a new one?
>> Answer: a new PR

- ~~use testground sdk directly from browser~~
- ~~add test plan to showcase how to do this~~
- ~~add integration tests which makes use of this testplan~~

### DIFF
--- a/plans/example-browser/failure.js
+++ b/plans/example-browser/failure.js
@@ -3,9 +3,7 @@ module.exports = async (page, runenv, client) => {
     return window.navigator.userAgent
   })
   if (userAgent !== 'testground') {
-    throw new Error(`expected user agent 'testground', unexpected: ${userAgent}`)
+    throw new Error(`failure.js: If you see this message, that means the test failed as expected. User agent is ${userAgent}`)
   }
-
-  // we do not ever expect to reach here, this is a failure test
-  runenv.recordMessage('playwright: success: testground platform detected')
+  runenv.recordMessage('failure.js: If you see this message, something bad has occurred - this test is meant to fail')
 }
--- a/plans/example-browser/failure.js
+++ b/plans/example-browser/failure.js
@@ -1,0 +1,11 @@
+module.exports = async (page, runenv, client) => {
+  const userAgent = await page.evaluate(() => {
+    return window.navigator.userAgent
+  })
+  if (userAgent !== 'testground') {
+    throw new Error(`expected user agent 'testground', unexpected: ${userAgent}`)
+  }
+
+  // we do not ever expect to reach here, this is a failure test
+  runenv.recordMessage('playwright: success: testground platform detected')
+}
--- a/plans/example-browser/playwright-runner.js
+++ b/plans/example-browser/playwright-runner.js
@@ -2,24 +2,40 @@ const { chromium, firefox } = require('playwright')
 
 module.exports = (module) => {
   const testFn = require(module)
-  return (runenv, client) => {
-    runTestFn(runenv, client, testFn)
+  return async (runenv, client) => {
+    runenv.recordStart()
+    await runTestFn(runenv, client, testFn)
+    runenv.recordSuccess()
   }
 }
 
 async function runTestFn (runenv, client, fn) {
   let browser
+  let result
 
-  runenv.recordMessage('playwright: launching browser and opening new page')
-  if (/firefox/i.exec(runenv.testInstanceParams.browser)) {
-    browser = await firefox.launch()
-    runenv.recordMessage('playwright: firefox launched')
-  } else {
-    browser = await chromium.launch()
-    runenv.recordMessage('playwright: chromium launched')
+  try {
+    runenv.recordMessage('playwright: launching browser and opening new page')
+    if (/firefox/i.exec(runenv.testInstanceParams.browser)) {
+      browser = await firefox.launch()
+      runenv.recordMessage('playwright: firefox launched')
+    } else {
+      browser = await chromium.launch()
+      runenv.recordMessage('playwright: chromium launched')
+    }
+    const page = await browser.newPage()
+    runenv.recordMessage('playwright: new page opened')
+
+    result = await fn(page, runenv, client)
+  } finally {
+    if (browser) {
+      try {
+        await browser.close()
+        runenv.recordMessage('playwright: browser closed')
+      } catch (error) {
+        runenv.recordMessage(`playwright: failed to close browser: ${error}`)
+      }
+    }
   }
-  const page = await browser.newPage()
-  runenv.recordMessage('playwright: new page opened')
 
-  return await fn(page, runenv, client)
+  return result
 }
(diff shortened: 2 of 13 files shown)


GlenDC:
Placebo integration test seems to fail, which is unrelated from the work done for this PR?

plans/example-browser/playwright-runner.js
```diff
@@ -0,0 +1,41 @@
...
+  }
+}
+
+async function runTestFn (runenv, client, fn) {
+  let browser
+  let result
+
+  try {
+    runenv.recordMessage('playwright: launching browser and opening new page')
+    if (/firefox/i.exec(runenv.testInstanceParams.browser)) {
```
brdji, line 18: Is this the standard way of checking which browser to open? 
I would prefer a simple string check, and if the user inputs something like "opera", we should return an error.

plans/example-browser/playwright-runner.js
```diff
@@ -0,0 +1,41 @@
+const { chromium, firefox } = require('playwright')
+
+module.exports = (module) => {
```
brdji, line 3: Since this function forms the base of the tests, I would like to see comments on its behavior and usage; same goes for `runTestFn`

plans/example-browser/playwright-runner.js
GlenDC: No, not standard. Just my way of integrating that configurability into testground using a simple regex (to not have to worry about surrounding whitespaces or case sensitivity). So there are two things here:

1. require an exact string (be it case insensitive or not)
2. also expect for chromium an exact string, which would be the default

I can apply both things. Which I'll assume for now is what you want.

plans/example-browser/failure.js
```diff
@@ -0,0 +1,11 @@
...
+  const userAgent = await page.evaluate(() => {
+    return window.navigator.userAgent
+  })
+  if (userAgent !== 'testground') {
+    throw new Error(`expected user agent 'testground', unexpected: ${userAgent}`)
+  }
+
+  // we do not ever expect to reach here, this is a failure test
+  runenv.recordMessage('playwright: success: testground platform detected')
+}
```
brdji, line 11: I would tweak the Error / success messages to signify that the expected behavior is for this test to fail, and not reach the end of the function.

plans/example-browser/playwright-runner.js
GlenDC: Sure, is there a particular format of docstring you expect here? Wasn't under the impression that this needed any comments given the fact that it's (a) an example and (b) very small in scope. But sure, it also cannot hurt, especially for an example. So I'll add it with pleasure.

runFn is a private function though, but I'll never the less add a comment.

Note that for the second example (the one I still need to add and which will run testground-sdk from within the browser) the flow will still be a bit different and slightly more complex)

plans/example-browser/playwright-runner.js
brdji: Ignoring whitespaces or case sensitivity sounds like a good idea; I would apply both points.

plans/example-browser/playwright-runner.js
brdji: The functions are clear enough, but I have future developers in mind - when someone wishes to use testground browser tests to their own project, they will use this example (or others, if we replace this). It makes sense to keep the example as clear and documented and possible, so in the future, people don't have to question why / how something was done.

Does this make sense?

plans/example-browser/failure.js
brdji: If the `failure.js` is integrated into the CI, or if someone runs the test on their own, the messages can be a bit confusing. Here's what I would do: the default output should be something like 
`failure.js: If you see this message, that means the test failed as expected. User agent is ${}.`
And for the failure
`failure.js: If you see this message, something bad has occured - this test is meant to fail`

That way, if I see the message `failure.js - PASSED X` (or similar), it won't be as confusing.
Thoughts?

laurentsenta: Let's simplify and use the same code that we used in `failure.js`
it's explicit and self-documented.
https://github.com/testground/testground/blob/62c18ac2150aae5f49a091b03150872440e8c564/plans/example-js/failure.js

plans/example-browser/playwright-runner.js
laurentsenta: Both are for me in this version,

In future iterations, let's favor exact string matches when we can: It's hard to plan for how a parameter might evolve and what edge cases might bite us in the future. (remember windows 9? :)

plans/example-browser/playwright-runner.js
laurentsenta: Thanks for the note @brdji, huge 👍 to aiming for more documentation and more comments,

A few notes to make sure we're staying on the pragmatic side of things:

- We expect this code to change a lot over the next few weeks, so it's fine to hold off "user-oriented" documentation (this is the `testground:edge` version). A `README.md` would be a great start with an example or two coming from the integration test, no blockers for me.
- We need "maintainers-oriented" comments when a piece of code hides some specialized knowledge or some context that cannot be guessed by a maintainer having a quick read. If we feel like a piece of code is self-explanatory, let's not worry about comments.

laurentsenta:
The CI issues should have been fixed, could you rebase on `master` @GlenDC ?

laurentsenta (approved):

plans/example-browser/playwright-runner.js
GlenDC: Resolved as an exact string, as requested.

plans/example-browser/playwright-runner.js
GlenDC: Added the desired comments.

plans/example-browser/failure.js
GlenDC: Went with @brdji's proposal, seems the most fair IMHO.

GlenDC:
> @GlenDC thanks for the ping, could you rebase over the latest master? I'll make sure this goes in next.
> @brdji could you approve the PR if you are satisfied with the latest changes?

Ok. Master rebased. Ready for approval and merge.
Locally I already am working on step 2 and I'll open a Draft PR once this one is merged in.

brdji (approved):
Good job!

GlenDC:
@laurentsenta ball is in your camp again, as I do not have merge rights. All good now AFAIK.

mxinden:
Thanks for all the work here.

//CC @achingbrain for visibility.

PR #1479, opened 2022-09-26
Fix SKIP_LOG_PARSING condition

kerzhner:
The `if` condition is always true. With the change, the `if` condition is true only when `SKIP_LOG_PARSING` is set to a non-empty string.

``` bash
> echo $SKIP_LOG_PARSING
>
> if [ SKIP_LOG_PARSING ]; then echo "skip"; fi
> skip
> if [[ -n "$SKIP_LOG_PARSING" ]]; then echo "skip"; fi
> SKIP_LOG_PARSING=true
> if [[ -n "$SKIP_LOG_PARSING" ]]; then echo "skip"; fi
> skip
```

PR #1481, opened 2022-09-27
Additional hosts

kerzhner:
https://github.com/testground/testground/pull/1479 precedes this PR.

This PR contains the changes from https://github.com/testground/testground/pull/1286 in addition to an integration test for the changes. @hannahhoward describes a use-case for the changes.

Our team has been using the additional hosts feature for the following scenario. Our testground testsuite contains many state-channels golang peers that interact with each other. During test setup and teardown stages, the peers submit Ethereum transactions to and listen to events from a local Ethereum node (which is run via [Hardhat](https://hardhat.org/hardhat-network/docs/overview)). The Ethereum node is spun up outside of testground network. It manually added to the `testground-control` docker network. The Ethereum node persists for the duration of many test runs.


integration_tests/20_docker_additional_hosts.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+my_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "$my_dir/header.sh" "$my_dir/20_env.toml"
+
+docker build ./plans/additional_hosts/http_server -t http_server
+docker run -it -d -p 8080:8080 --name http_server http_server
+docker network connect testground-control http_server
```
laurentsenta, line 7: I believe we want to call `testground healthcheck --runner local:docker --fix` first,
this will create the control network if it's not there yet.

laurentsenta, line 7: - I believe we need to call `testground healthcheck --runner local:docker --fix` first (we might run this on a fresh session without the control network).
- Could you use something like https://hub.docker.com/r/hashicorp/http-echo/ ? it's 4mo instead of go's 300.

laurentsenta (changes requested):
This is pretty exciting, thanks for sharing and adding clear integration tests.
I requested a few changes, will take another look tomorrow.

I also created an EPIC with a few follow-up tasks to turn this into a first-class feature, https://github.com/testground/testground/issues/1485 I haven't used it (yet) so don't hesitate to leave comments if something is missing.

integration_tests/20_docker_additional_hosts.sh
kerzhner: Good call on both points!
- [3668f03](https://github.com/testground/testground/pull/1481/commits/3668f0322aac37b167f58d16c1a60add9295e7d4) moves the healthcheck toward the start of the integration test script.
- [8137ce1](https://github.com/testground/testground/pull/1481/commits/8137ce14812647639d0025e82b88c7377ff9bf37) removes the custom http server and replaces it with hashicorp/http-echo.

laurentsenta:
The fixes look good, thanks for bringing yet another feature into testground's master @kerzhner & state-channels team,

I'll get this in next, could you rebase over the current master?

I would appreciate it if you could add a line to the CHANGELOG.md file, but no blocker here (we just introduced this file).

Something like:

```md
### Added
- support adding containers to the control network See [PR 1481]

[PR 1481]: https://github.com/testground/testground/pull/1481
```

Would do the job.


@brdji should we wait for your review?

kerzhner:
> I'll get this in next, could you rebase over the current master?
> 
> I would appreciate it if you could add a line to the CHANGELOG.md file, but no blocker here (we just introduced this file).

Thanks for the support @laurentsenta! The feature branch has been rebased on the latest master and CHANGELOG is updated in [45adf75](https://github.com/testground/testground/pull/1481/commits/45adf7537f78600eaa6577af52683e9af8e0cc91).

laurentsenta (approved):
👍 thanks for updating the Changelog

pkg/sidecar/docker_reactor.go
```diff
@@ -79,6 +80,12 @@ func (d *DockerReactor) ResolveServices(runid string) {
 		os.Getenv(EnvInfluxdbHost),
 	}
 
+	additionalHosts := strings.Split(os.Getenv(EnvAdditionalHosts), ",")
+	logging.S().Infow("additional hosts", "hosts", os.Getenv(EnvAdditionalHosts))
+	if len(additionalHosts) != 1 || additionalHosts[0] != "" {
+		wantedRoutes = append(wantedRoutes, additionalHosts...)
+	}
```
galargh, line 87: Can't we just do the following here?
```suggestion
	wantedRoutes = append(wantedRoutes, additionalHosts...)
```

pkg/runner/local_docker.go
```diff
@@ -136,14 +138,24 @@ func (r *LocalDockerRunner) Healthcheck(ctx context.Context, engine api.Engine,
...
+	additionalHosts := "ADDITIONAL_HOSTS="
+	envHosts, hasHosts := engine.EnvConfig().Runners["local:docker"]["additional_hosts"].([]interface{})
+	if hasHosts {
+		for i, host := range envHosts {
+			if i > 0 {
+				additionalHosts += ","
+			}
+			additionalHosts += host.(string)
+		}
+	}
```
galargh, line 150: Could we simplify this piece with the following?
```suggestion
	envHosts, hasHosts := engine.EnvConfig().Runners["local:docker"]["additional_hosts"].([]string)
	if hasHosts {
		additionalHosts += strings.Join(envHosts, ",")
	}
```

galargh (approved):
Left 2 code suggestions that I think could simplify things a bit. Overall looks great though :) Thank you for doing this! It's an exciting new feature!

pkg/sidecar/docker_reactor.go
kerzhner: Yep, that's a good idea!

pkg/runner/local_docker.go
kerzhner: Good idea! The suggestion has been applied.

laurentsenta:
Merged, thanks for the contribution @kerzhner !

PR #1497, opened 2022-10-19
ROADMAP.md: refine

laurentsenta:


ROADMAP.md
```diff
@@ -98,7 +98,7 @@ We want to ensure Testground is valuable and stable before we grow its feature s
 
 - Delivery: Q4 2022
```
galargh, line 99: I feel `- The team can use the latest Testground version, create new releases, and upgrade the cluster.` might spill into Q1 2022.

galargh (approved):

ROADMAP.md
galargh: That's cause I suspect we'd want to have `How to fix common EKS failure scenarios` covered here too.

BigLep (approved):

ROADMAP.md
laurentsenta: Moving to Q1 2023 sounds good to me,

The cut I wanted to make is:
- `(2. Refresh Testground's EKS support)` is about getting a cluster deployed and usable by testground maintainers & users willing to deal with experimental mode. It covers Bloxico's work + some automation. 
- `(4. Provide a Testground As A Service Cluster used by libp2p & ipfs teams)` is about getting a cluster deployed and usable by "everyone". Which covers stability & maintenance. I would expect we create the cluster on day 1 of this milestone and learn about failure scenarios during the milestone.

galargh: Got it! Let's leave it for Q4. TLDR milestone 2 - we can use deliverables from Bloxico; milestone 4 - we are good at it.

PR #1501, opened 2022-10-20
README.md: update with install.sh instruction

jxs:
Hi, this seems to be missing in the instructions.
Plan to do the same to https://github.com/testground/docs with the final agreed wording after this one is merged.
Thanks :)


jxs:
Sorry I just noticed that this is probably an older version. I was able to use the testground from the $GOPATH.

PR #1502, opened 2022-10-24
examples: add node & cross-browser example

GlenDC:
Add advanced browser example testplan Goal of this plan is to show how you can define a testplan which can be run in a node environment as well as within a browser, without having to change anything to your testplan code.

This does mean we also need to allow the `@testground/sdk` javascript package to be run in the browser as-is. To allow this I have a branch which I run this testplan against at <https://github.com/glendc/sdk-js/tree/feat/support-browser>:

- the FS features are not possible in the browser, so these need to be disabled;
- the environment variables are not available when creating the runtime from env in browser,
  so for the browser we instead assume that these are available in the `window` object as `window.testground.env`;

For both these changes we assume one is in the browser when `process.title === 'browser'`, this because the `browser` npm (polyfill webpack) module sets that as the `process` global variable.

Within the example the testplan is found under `src/`, the other three directories form together the "magic" solution to allow the node testplan to also be run in the browser:

- `scripts/`: contains a polyfill folder to inject an import in `src/index.js` which is required to polyfill some logic required by winston (an import by `@testground/sdk`) 
- `browser/`: the entry point for the `docker:generic` testplan, and which will spawn and control everything else;
     - `browser/server/`: simple express served — controlled by the `browser/` and which will serve the web-pack'd testplan;

TODO:
- ~~allow debug breakpoints in `chrome://inspect` without having to refresh;~~ (not an issue for me, and don't think there is a way to do it otherwise)
- [x] make the tests actually run and pass (as for now we no longer have errors and everything seems polyfill'ed, but the test doesn't seem to run (despite the fact that we do see logs coming through);
- [x] open a `@testground/sdk` MR and align on the approach there: my goal is that this sdk can be used just the same from the browser as well as the node env;
- [x] make the browser magic solution more smart in terms of how to wait for test to finish: somehow hook into an event which will be called when testplan is finished, not before and not later;

Related PRs:

- ~~add browser support to `@testground/sdk`: https://github.com/testground/sdk-js/pull/25~~
   - replaced with new approach in `@testground/sdk`: https://github.com/testground/sdk-js/pull/26
- support docker runner (additional) environment arguments (required to allow the use of environment variables to choose what browser engine to test one, instead of having to rebuild the docker container): https://github.com/testground/testground/pull/1507

Potential improvements but probably out of scope:

- it would be great if we can choose what exposed port we bind to on the host machine (instead of currently any available open port) as it would mean we can always inspect against the same chrome address (but given this is defined by the docker Go library I do not have much hope for this one);
   - this one can be tackled in future, for now it is tracked in the following issue: https://github.com/testground/testground/issues/1506

---

In function of https://github.com/testground/testground/issues/1386.


GlenDC:
FYI @laurentsenta , `ci/circleci: build-test-linux-k8s ` fails, but I think that's once again for a reason different than this PR.

integration_tests/example_05_browser_node.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
```
laurentsenta, line 1: Could you add this test to the Makefile? I'd like to see the CI failing if this is not yet passing.

laurentsenta: (please ignore the k8s test suite, it is flaky)

GlenDC: I was planning to do only once my test is actually running. For now it is blocked on sending over a WS to the sync service. Planning to get that working this week.

GlenDC: I already opened the PR more to sync on the direction and what is still planned to be done as part of this PR

GlenDC: @laurentsenta added the PR for `@testground/sdk` as well in this PR description: <https://github.com/testground/sdk-js/pull/25>

plans/example-browser-node/browser/index.js
```diff
@@ -0,0 +1,105 @@
+const { chromium, firefox, webkit } = require('playwright')
+const { exit } = require('process')
+const { spawn } = require('child_process')
+
+const spawnServer = require('../server')
```
laurentsenta, line 5: Is the server only used to enable browser support?
Could we move the code inside the `browser` folder to make the split between browser support, node support, and actual test code, more obvious?

plans/example-browser-node/browser/index.js
```diff
@@ -0,0 +1,105 @@
...
+    }, process.env)
+
+    console.log('opening up testplan webpage on localhost')
+    await page.goto('http://127.0.0.1:8080')
+
+    // TODO: wait for the test to finish somehow :|
+
+    console.log('start browser exit process...')
+
+    if (process.env.HALT_BROWSER_ON_FINISH === 'true') {
```
laurentsenta, line 82: Why is this needed?

plans/example-browser-node/manifest.toml
```diff
@@ -0,0 +1,34 @@
...
+[builders."docker:node"]
+enabled = true
+
+[builders."docker:generic"]
+enabled = true
+
+[builders."docker:generic".build_args]
+BROWSER_KIND = "firefox"
+BROWSER_DEBUG_PORT = "5321"
+HALT_BROWSER_ON_FINISH = "false"
```
laurentsenta, line 16: It should be possible to override these values at runtime right?

We should be able to build one image and override the browser variable when you start the container, which means we need the ability to pass custom env's to a testground plan.

plans/example-browser-node/scripts/polyfill.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+# polyfill setimmediate for winston in browser
+sed "1s/^/require('setimmediate')\n\n/" src/index.js > src/index_polyfill.js
+mv src/index_polyfill.js src/index.js
```
laurentsenta, line 5: Could you share some context here? Could we use the `browser` field? https://github.com/defunctzombie/package-browser-field-spec

plans/example-browser/README.md
```diff
@@ -0,0 +1,35 @@
+# Example Browser Testplan
```
laurentsenta, line 1: Is this plan destined to be removed and replaced by the `example-browser-node`?

laurentsenta:
(remember to use "convert to draft" for your draft PRs)

plans/example-browser-node/browser/index.js
```diff
@@ -0,0 +1,105 @@
...
+    page.on('console', (message) => {
+      const loc = message.location()
+      console.log(`[${message.type()}] ${loc.url}@L${loc.lineNumber}:C${loc.columnNumber}: ${message.text()} — ${message.args()}`)
+    })
+
+    console.log('prepare page window (global) environment')
+    await page.addInitScript((nodeEnv) => {
+      const env = {}
+      for (const k in nodeEnv) {
+        if (k.startsWith('TEST_')) {
```
laurentsenta, line 68: This might work now, but it will bite us eventually,

Since the testground SDK is the one official owner of reading and parsing testground envs,
could you make sure we rely on it to load & forward the env to the browser runtime?

Adding a serialize / deserialize feature in https://github.com/testground/sdk-js/blob/master/src/runtime/params.js might be enough.

plans/example-browser-node/browser/index.js
GlenDC: Sure. it is used to serve the testground plan in the browser, but indeed, we can make it part of the `browser` folder, np!

plans/example-browser-node/browser/index.js
GlenDC: It is not needed. However, it can be useful to be able to prevent the test from exiting upon failure / success, and instead allow it to be open. This to be able to connect to it for a remote debugging session and restart the test by refreshing the browser with your breakpoints now set.

plans/example-browser-node/manifest.toml
GlenDC: I agree, this is for now, as discussed on slack, a limitation of testground. As currently you cannot pass args to the `local:docker` runner, while you can do to the `docker:generic` builder. So once the first is possible as well we can get around that indeed.

Unless I am missing something of course.

plans/example-browser-node/scripts/polyfill.sh
GlenDC: Not sure, might need some help there, I am certainly no expert in the NodeJS world. What would be the diference with our current approach? It's certainly not the case that I can just run winston as-is without any shimming in the browser, unless you just need to enable some config for that?!

plans/example-browser/README.md
GlenDC: I was under the impression that this was not the case, but we can remove it if you want it.
To me it is fine to stay as this is a very different approach. In the original plan we run the test from within NodeJS, and do not try to run testground-sdk in the browser.

plans/example-browser-node/browser/index.js
GlenDC: What exactly might be us eventually? Also what would the serialize/deserialize feature solve here? I am open to anything, I'm asking as I am not following exactly what you're referring to.

plans/example-browser-node/manifest.toml
GlenDC: PR open: https://github.com/testground/testground/pull/1507

plans/example-browser/README.md
GlenDC: @laurentsenta want me to remove it, or what is your take on it?

plans/example-browser-node/browser/index.js
GlenDC: Could you please clarify on this? A meeting isn't needed, would be useful if you can simply respond.

laurentsenta: @GlenDC this code relies on the fact that every variable is prefixed with `TEST_`, which is not enforced anywhere, it's an implicit specification.

The reliable way to access Testground's environment variables is to use the SDK. The corresponding namespace is defined in: https://github.com/testground/sdk-js/blob/master/src/runtime/params.js.

I want to make sure we rely on the SDK to retrieve these values and don't rely on implicit constraints like "every env variable is prefixed with `TEST_`, which might become untrue anytime).

plans/example-browser-node/scripts/polyfill.sh
laurentsenta: (I think we developed this discussion in the js-sdk already and this can be resolved, let me know if you have more questions)

plans/example-browser/README.md
laurentsenta: If I understand correctly:
they both answer the same question: "as a test author, how do I create a test that runs in a browser runtime?" and there's no obvious reason to use the version that can't access the testground sdk.

I would like to make sure we have one "obvious" way to do things, that we might make "Testground native" eventually. It will be easier for maintainers to improve and communicate about it, and it'll be easier for users to get started and rely on our code.

plans/example-browser-node/browser/index.js
```diff
@@ -0,0 +1,105 @@
...
+          env[k] = nodeEnv[k]
+        }
+      }
+      window.testground = { env }
+    }, process.env)
+
+    console.log('opening up testplan webpage on localhost')
+    await page.goto('http://127.0.0.1:8080')
+
+    // TODO: wait for the test to finish somehow :|
```
laurentsenta, line 78: We could set a global flag in the window and use https://playwright.dev/docs/api/class-page#page-wait-for-function to detect when it's toggled. Add the toggle after `await invokeMap(testcases)` for testing, then maybe move it to the js SDK, at the end of `invokeMap`.

I wish we could rely on the SDK here and listen to the success / failure events (https://github.com/testground/sdk-go/blob/master/runtime/runenv_events.go#L13-L19), but these events do not contain the instance ID, and we might lose metrics and files written AFTER we called RecordSucess.

plans/example-browser-node/browser/index.js
laurentsenta: Got it, we have similar behavior with the `keep_containers` flag. Could you use this approach? Something like `KeepOpenedBrowser`?

GlenDC: Sure, will look into it.

plans/example-browser-node/browser/index.js
GlenDC: Nice, will look into it. Thanks!

plans/example-browser-node/browser/index.js
GlenDC: Moved `server` to `browser/server`: https://github.com/testground/testground/pull/1502/commits/01a187cf80f594a3eb8001da79734b28fae8b3fe

plans/example-browser-node/browser/index.js
GlenDC: For now I have already named it `TEST_KEEP_OPENED_BROWSERS`, but seems like a proper runner solution like `keepOpenedBrowser` variable seems better suited for once we have a `docker:browser`, and then we can also see how to integrate browser-only things into the local docker runner. Resolving it as such, but please re-open if you do not agree.

plans/example-browser/README.md
GlenDC: Fine for now, it's anyway in the Git history so we can always refer to the version on the older commit if we ever have to. Will delete it. RIP.

plans/example-browser-node/browser/index.js
GlenDC: Added functionality in sdk-js PR to be able to do this: https://github.com/testground/sdk-js/commit/06df3ce6c9c904d52fa6d081aa1ffd204b5de69a

This I have implemented also in this PR already (the integrated functionality): https://github.com/testground/testground/pull/1502/commits/d7f0b02e72293586268dd127c2bf11c6bd4616ac

plans/example-browser-node/manifest.toml
GlenDC: Once the above PR is merged in, I'll merge master into this branch (again) and I'll switch over the logic I can move over :)

GlenDC: At that point let's also to sit together so we can try to get the plan to work on the 3 browsers (success + failures) without requiring 3 different docker builds (not that I would know how to do that, as not sure how to pass args to the cfgs using build/run, not sure even possible?

plans/example-browser-node/browser/index.js
GlenDC: Added a hook in https://github.com/testground/sdk-js/pull/26/commits/03fc268de60f1cc4286ec490fe5ef9eb9e402263, which will make sure `window.testground.result` is defined at the end. If we ever want we can distinguish between success (`true`) and something else (e.g. `Error` for a failure), but for now I just wait until it is defined: <https://github.com/testground/testground/pull/1502/commits/77e20fe4de06b1fea81f08e75a290a5caf3e54fc>

GlenDC:
@laurentsenta: MR is ready for a final review for what is me concerned.

- all threads that I can resolve have been resolved
- the code now works and is tested in CI (using chromium)
- all logging is nicely working and interaction with sync also is smooth as butter

Only comment that is left is regarding the browser selection (https://github.com/testground/testground/pull/1502#discussion_r1010637049):
- this will require the docker local args PR to be merged in (https://github.com/testground/testground/pull/1507)
- it will also require us to pair program to get this setup in the integration file, as I'm not sure how to pass on those parameters to our unit tests, according to your message we might need to use a composition file for that, so for that I might need some help :)

GlenDC (commented):
I self-reviewed the PR, and all looks good according to me.

GlenDC:
Latest update posted on testground's slack channel: https://filecoinproject.slack.com/archives/C03KQ91M0E8/p1668298843557339

Makefile
```diff
@@ -81,4 +81,18 @@ test-integ-examples:
 	./integration_tests/example_01_rust.sh
 	./integration_tests/example_02_js_pingpong.sh
 	./integration_tests/example_03_generic_artifact.sh
-	./integration_tests/example_04_browser.sh
\ No newline at end of file
+	./integration_tests/example_04_browser_node.sh
+
+kind-cluster:
```
laurentsenta, line 86: That code was removed, something went wrong during a rebase over master.

laurentsenta (changes requested):
Almost there, thanks for sharing! These are the 3 fixes I'd like to solve before merging (nothing big):

- [x] fix the `Makefile` rebase issue
- [x] let's not block on the env feature (https://github.com/testground/testground/issues/1386#issuecomment-1313473589)
- [x] we're blocked by https://github.com/testground/sdk-js/pull/26

cc @galargh and @nonsense if they want to review too.

Makefile
GlenDC: Resolved in bfaa5b396bdf23ef48cc8236229176e2be9a2ea0.

GlenDC:
> * [x]  fix the `Makefile` rebase issue
> * [ ]  let's not block on the env feature ([#1386 (comment)](https://github.com/testground/testground/issues/1386#issuecomment-1313473589))
> * [ ]  we're blocked by [testground/sdk-js#26](https://github.com/testground/sdk-js/pull/26)

First step is done. Third I cannot do yet until merged (if you want to block on that).
For step 2 I'll require some help as I'm not sure how I can pass these build args while using `testground build` I suppose it is `build-cfg` or something like that, but not sure how to use it.

integration_tests/example_04_browser_node.sh
```diff
@@ -0,0 +1,147 @@
...
+# Browser (Chromium): success
+
+testground run single \
+    --plan=testground/example-browser-node \
+    --testcase=output \
+    --builder=docker:generic \
+    --use-build=testplan:example-browser-node-browser \
+    --runner=local:docker \
+    --instances=1 \
+    --tp BrowserKind=chromium
```
laurentsenta, line 49: You miss the `\`, which means the shell interprets the next line as a new command.

laurentsenta: (have you tried running the tests on your machine? That might help with iteration speed).

GlenDC:
All tests pass @laurentsenta and last comments have been applied AFAIK.
Please give it another review and merge if possible.

laurentsenta (approved):
lgtm, thanks for sharing; it's pretty exciting to see this used soon!

I added @galargh and @achingbrain to give them a chance to review as well,

A few follow-up tasks, let's not block on this:

- [ ] run the sync test also, ideally a single test that connects all the browsers + node.
- [ ] move the build process to a single builder (`docker:generic` I guess), so that the runtime is actually "just" a flag.
- [ ] make a release for the `js-sdk` and tag a version.

Issue #1466, opened 2022-09-19
Is it possible to change --file when using docker:generic builder?


francardoso93:
How to use a dockerfile that is not called Dockerfile?

francardoso93:
https://github.com/testground/testground/blob/eaab2784579967f0cab559242a7124c161243273/pkg/build/docker_generic.go#L65

Seems that it isn't possible.. 
Could I open a PR for this?

laurentsenta:
That makes sense, feel free to add me as a reviewer when it's ready

PR #1503, opened 2022-10-24
Make Dockerfile name possible to specify

ackintosh:
This PR closes https://github.com/testground/testground/issues/1466 and also relates https://github.com/testground/sdk-rust/issues/40.

This PR makes Dockerfile name possible to specify via `manifest.toml` file.

```diff
  name = "testplan"

  [defaults]
  builder = "docker:generic"
  runner = "local:docker"

  [builders."docker:generic"]
  enabled = true
+ dockerfile = "testplan.Dockerfile"

  [[testcases]]
  ...
  ...
```

### DIFF
--- a/pkg/build/docker_generic.go
+++ b/pkg/build/docker_generic.go
@@ -26,8 +26,9 @@ type DockerGenericBuilder struct {
 
 type DockerGenericBuilderConfig struct {
 	// Custom base path where we find the test source
-	Path      string             `toml:"path" default:"./"`
-	BuildArgs map[string]*string `toml:"build_args"` // ok if nil
+	Path       string             `toml:"path" default:"./"`
+	Dockerfile string             `toml:"dockerfile" default:"Dockerfile"`
+	BuildArgs  map[string]*string `toml:"build_args"` // ok if nil
 }
 
 // Build builds a testplan written in Go and outputs a Docker container.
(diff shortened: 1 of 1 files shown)


ackintosh:
Hmm, seems the `default` doesn't work.

https://github.com/testground/testground/blob/95005b0878cf3490b5c50e70fc321785e408c529/pkg/build/docker_generic.go#L30

ackintosh:
BurntSushi/toml doesn't support `default` 👀 

https://github.com/BurntSushi/toml/issues/171

ackintosh:
Okay that works, but the code may not be good because I'm not familiar with Go.

pkg/build/docker_generic.go
```diff
@@ -26,43 +26,60 @@ type DockerGenericBuilder struct {
...
+
+	planPath := cfg.Path
+	basePathForPlan := path.Join("/plan", planPath)
+
 	opts := types.ImageBuildOptions{
 		Tags:        []string{in.BuildID},
 		BuildArgs:   cfg.BuildArgs,
 		NetworkMode: "host",
-		Dockerfile:  filepath.Join(basePathForPlan, "Dockerfile"),
+		Dockerfile:  filepath.Join(basePathForPlan, cfg.Dockerfile),
```
laurentsenta, line 82: It looks like this was the core change; why do you need to introduce `newDockerGenericBuilderConfig`?

laurentsenta (changes requested):
Thanks for sharing this change,

- [ ] a question on the generic function we implemented,
- [ ] we need a test.

PR #1507, opened 2022-11-02
feat: support environment variables for local docker

### LINKED ISSUE #1505: support runner args for local:docker
## Description

Today, one can already define in the `manifest.toml` file of a testplan builder-args for a `docker:generic` builder.

It is however not yet possible to define args that can be given to the `local:docker` runner when running an already built image.
As a result it is currently only possible to add dynamic features to a Generic dockerfile using build args, rather than also
supporting args that can be injected without having to rebuilt the image.

If we can support this it can make our`docker:generic` images a lot more flexible.

GlenDC:
- Add `additional_envs` option to `docker:runner` which lets a user override envs.

Closes #1505


GlenDC:
@laurentsenta can this be reviewed so we can aim to merge?

laurentsenta:
@GlenDC 👍 Could you add tests and a line in the changelog?

GlenDC:
> @GlenDC 👍 Could you add tests and a line in the changelog?

Sure thing. Will do and I'll let you know once it is done!

GlenDC:
@laurentsenta the changelog line is added, and I've added this to an existing example plan.
Would that suffice for you, if not, can you help me by specifying more exactly what more you want?

How do you actually pass run config values when running a plan via CLI? There is `run-cfg` but I do not see anything using it? Are there examples of this, as I suppose that would be the way to pass custom run cfgs, e.g. our custom environment variables.

pkg/runner/local_docker.go
```diff
@@ -70,6 +70,12 @@ type LocalDockerRunnerConfig struct {
 	// (default: ["nofile=1048576:1048576"]).
 	Ulimits []string `toml:"ulimits"`
 
+	// EnvironmentVariables can be used to set additional environment variables.
+	// Note to use unique environment variables not used by Testground,
+	// as your custom environment variables will get overwritten by the ones
+	// also set by Testground.
+	EnvironmentVariables map[string]string `toml:"environment_variables"`
```
laurentsenta, line 77: Thanks for the clear documentation,

Could you rename it to `AdditionalEnvs`? It'll be explicit that these are "bonus" env variables, which don't overwrite Testground's and will be consistent with `AdditionalHosts`

GlenDC: Alright. Done :)

GlenDC:
> How do you actually pass run config values when running a plan via CLI? There is `run-cfg` but I do not see anything using it? Are there examples of this, as I suppose that would be the way to pass custom run cfgs, e.g. our custom environment variables.

Could you also explain me that @laurentsenta, as I would want to do that for my browser plan I imagine, in between different tests on same image?

plans/example/output.go
```diff
@@ -1,14 +1,20 @@
...
 
 // Demonstrate test output functions
 // This method emits two Messages and one Metric
 func ExampleOutput(runenv *runtime.RunEnv, initCtx *run.InitContext) error {
-	runenv.RecordMessage("Hello, World.")
+	if msg := os.Getenv("TESTGROUND_HELLO_MESSAGE"); msg != "" {
+		runenv.RecordMessage(msg)
+	} else {
+		runenv.RecordMessage("Hello, World.")
+	}
```
laurentsenta, line 17: That's an example but not a test: Remember we're building TESTground, we want our codebase to be tested thoroughly, if we break this feature in 3 months, this example won't detect it.

Here is an example of testing additional hosts:
https://github.com/testground/testground/blob/master/integration_tests/20_docker_additional_hosts.sh
https://github.com/testground/testground/blob/master/plans/additional_hosts/main.go

Note how the code fails if the host is not set.

Another example with the mixed builders feature:
https://github.com/testground/testground/blob/master/integration_tests/15_docker_mixed_builders_configuration.sh
https://github.com/testground/testground/blob/9c13f3bbe956c3cc29591db4b8b3c63b6fa1f6dc/plans/_integrations_mixed_builders/go/main.go#L11

Note how the test will fail if the values are not set correctly.

plans/example/manifest.toml
```diff
@@ -20,6 +20,9 @@ enabled = true
 [runners."local:docker"]
 enabled = true
 
+[runners."local:docker".additional_envs]
+TESTGROUND_HELLO_MESSAGE = "Hello, Docker."
+
 [runners."local:exec"]
 enabled = true
```
laurentsenta, line 28: (commenting here to get a thread)

> How do you actually pass run config values when running a plan via CLI? There is run-cfg but I do not see anything using it? Are there examples of this, as I suppose that would be the way to pass custom run cfgs, e.g. our custom environment variables.
> 
> Could you also explain me that @laurentsenta, as I would want to do that for my browser plan I imagine, in between different tests on same image?

At the moment, you would use a composition file, then pass the parameters there, for example:

https://github.com/testground/testground/blob/9c13f3bbe956c3cc29591db4b8b3c63b6fa1f6dc/plans/_integrations_mixed_builders/_compositions/issue-1357-mix-builder-configuration.toml#L22-L23

plans/example/output.go
laurentsenta: (The scripts are registered in the Makefile for now: https://github.com/testground/testground/blob/9c13f3bbe956c3cc29591db4b8b3c63b6fa1f6dc/Makefile#L61)

GlenDC: Ok going to see to add `21_docker_additional_envs.sh` then and the like :)

plans/example/manifest.toml
GlenDC: Can you show me how that would work? As I am not sure how to go from a composition file parameter to a runner config value?!

plans/example/output.go
GlenDC: Added the additional test and also enabled the `20_*.sh` test, as that one wasn't enabled yet.

GlenDC:
@laurentsenta all tests have been enabled and PR is ready for being merged what is me concerned.

GlenDC (commented):
Self-Reviewed it again as well, looks all good to me!

GlenDC:
@laurentsenta what more would you need to be done here before we can more towards a merge path?

laurentsenta:
Code looks fine to me, adding @galargh and @nonsense if they have time for a second look

GlenDC:
Ok perfect. @galargh and @nonsense , do let me know if something needs to be done or how we can move this PR forward :) Thanks!

GlenDC:
Going to close this one as for now there is not enough motivation to really add this feature, given that it's considered better practise to use parameters (SDK) for these purposes.

PR #1516, opened 2022-11-08
feat: support interop matrix

### LINKED ISSUE #1493: EPIC: Implement multiple runs per compositions
We're working with the libp2p team on multi-dimensional testing:
https://github.com/libp2p/test-plans/issues/53

They need a way to describe large test matrixes,
one composition might combine every libp2p version able to communicate over webrtc and use yamux,
another might combine every libp2p version able to use mutex X and transport Y.

The main risk is the composition complexity: Every test run would need a new composition file at the moment.
And the configuration used in libp2p/test-plan is quite complex.
This is the build configuration for go libp2p master: https://github.com/libp2p/test-plans/blob/c71602e88c8a20c6ecc7c18fe1f5b5cf570d3b11/ping/_compositions/go-rust-interop.toml#L33-L56

Contributes to Milestone #1514 

## Current Solution

Add a `[runs]` field to compositions

```toml
[[groups]]
# the list of instances

[[runs]]
# the list of runs
# a run is of the form:
    id = 'run_id'

    [runs.test_params]
    # custom param for the whole run

    [[runs.groups]]
    id = 'id of a test instance'
    # TODO: maybe use 2 values: `id` and `group_id`,
    #       and default `id` to `group_id`
    instances = {count = 1}

    [[runs.groups.test_params]]
    # custom param for this instance

    [[run.groups]]
    id = 'another instances'
    # and so on.
```

- demonstration: https://github.com/libp2p/test-plans/tree/feat/interop-dashboard/composer
- discussion: https://github.com/libp2p/test-plans/issues/55#issuecomment-1278541910

```bash
testground run composition --file=./composition-interop.toml --run-index=42 # build and run a single test
testground run composition --file=./composition-interop.toml # build & run all tests
testground run composition --file=./composition-interop.toml ... --result ./results.csv
```

## Tasks

- [x] list the templates required from https://github.com/libp2p/test-plans/tree/feat/interop-dashboard/composer and implement them
- [x] implement `pick` and `toml` and `withEnv` (https://github.com/libp2p/test-plans/blob/3fccfb27bbaba204437b93a73d8d718f4b372442/composer/demo/composition-interop.toml#L87-L97)
- [x] implement the runs field and add an option to build and run a single id (fail in other cases)
- [x] implement a way to run "all" runs at the same time and figure out a way to merge outcomes
- [x] implement a way to output the outcomes to a csv file, differentiate fail (build) and fail (run)

### Follow-up Tasks

- [ ] document features
    - [ ] new templates
    - [ ] new cli parameters
- [ ] Add a feature to extract the list of artifacts at the end of a build?
    - https://github.com/libp2p/test-plans/tree/feat/interop-dashboard/composer#follow-ups
- [ ] Fix the composition -w parameter, do not overwrite the file, accept a path instead
- [ ] implement a `(partial "./template.toml")`
- [ ] implement a `--continue-on-build-error` that runs "as many tests as possible" even on build failures
- [ ] Update go versions (enable generics & fix issues with golangci-lint)
- [ ] Refactor the integrations tests to make them maintainable,
- [ ] Drop the concept build groups request during a run
  - [ ] sending a composition with missing artifacts is enough
- [ ] Templating follow-ups
  - [ ] Extract templating to another cli command (composable testground)
  - [ ] Investigate: https://github.com/Masterminds/sprig
  - [ ] Investigate "missingkey=error"
- [ ] improve client-side run/builds
  - [ ] make the request retriable
  - [ ] generalize the use of strategy, which should reduce code required and make it unit-testable
- [ ] Refactor how we manage tasks and generate outcomes,
  - [ ] Make outcomes data type shared between every runner
  - [ ] Make the outcome for build tasks consistent with the others
- [ ] implement server-side runs feature
  - [ ] refactors tasks to support task groups
  - [ ] refactor the supervisor to support task groups
  - [ ] refactor the request system to support task groups
  - [ ] remove the need for waiting on the client side

Something similar was built in lotus:
https://github.com/filecoin-project/lotus/tree/master/testplans/composer

### How to deal with build & run failures:

If I try to run N tests for K different versions, and the build for 2 versions is failing:
Should Testground keep building, and run the tests for the builds that succeeded?

Proposal:

Add an option that tells testground: keep going:

```
testground run composition .... --continue-on-build-error
```

### Are we happy with current templates?

Templating is a very useful scripting engine we integrate into testground. For "simple" use case like parametrization (custom go version, generating large tests, etc) it makes a lot of sense.

Are we happy with this approach? Another approach could be: the user generates toml files with their own scripting engine (js, go, etc).

Proposal: make template reusables with `partial`

```toml
(partial "./groups-go.toml" .) 
```

This would import another file and use it as a template.

laurentsenta:
Fixes #1493

Dear reviewers, extra care is taken to have clean commit steps. In case we want to split this PR.
It will be much nicer to review by going through the git history (in github: open the first commit then use the Prev / Next buttons).

- [x] Implement `pick` and `toml` and `withEnv`
- [x] Update composition
    - [x] Split `composition.go` into 3 different files (prepare, validate, and data)
    - [x] Simplify the validation processes to clarify validation/preparation
    - [x] Fix encoding/decoding using `mapstructs`
    - [x] Add the `runs` field
- [x] Implement client-side runs field
    - [x] add an option to build and run a single id (fail in other cases)
    - [x] add an option to gather results into a CSV
    - [x] we start multiple runs on the client side for now (build happens once)
- [x] Turn runs step into a Strategy object to make follow-up refactoring easier
    - [x] This lets us "name" operations and reuse them more easily
    - [x] This prepares for simplification (remove duplicate code in build and run operations) and testing (mock API calls).
- [x] Implement server-side runs field
    - [x] Add validation
    - [x] Process a run with the `runids` parameters
    - [x] Fix the composition returned in the task
- [x] Add tests, etc.

### the `runs` field is processed client-side

We have two options when sending a composition with multiple `runs` either:
- create one task that contains a group of runs,
   - This is the ideal way to do this, and I'd like to implement this later. It means cleaning up the way we represent tasks and gather outcomes, etc. At the moment even outcomes between two runners are processed very differently. I prepared this feature by this workflow as much as possible.
- create multiples tasks, one for each run,
    - This is the less intrusive way. We could do it server side, but then the server has to hold long-running states, we have to serialize it, etc. Which will end up looking like a weak implementation of task groups. Since we also want to implement retries and improvements to the client's CLI, I moved the code there.

## Follow-up Tasks

See https://github.com/testground/testground/issues/1493


pkg/api/composition_validation.go
```diff
@@ -0,0 +1,123 @@
...
+		m := make(map[string]bool, len(r.Groups))
+		for _, x := range r.Groups {
+			if _, ok := m[x.ID]; ok {
+				return fmt.Errorf("group ids not unique; found duplicate: %s:%s", r.ID, x.ID)
+			}
+			m[x.ID] = true
+		}
+	}
+
+	// Recalculate instance counts
```
galargh, line 65: Why are we doing it here?

pkg/cmd/run.go
```diff
@@ -220,98 +249,368 @@ func run(c *cli.Context, comp *api.Composition) (err error) {
...
-			User:   cfg.Client.User,
-			Repo:   c.String("metadata-repo"),
-			Branch: c.String("metadata-branch"),
-			Commit: c.String("metadata-commit"),
+	// Execute!
+
+	// Compute priority
+	isCollecting := c.Bool("collect")
+	isMultiple := len(runIds) > 1
+	isWaiting := c.Bool("wait") || isCollecting || isMultiple
```
galargh, line 257: I understand because I saw the implementation but it seems a little unexpected from the user perspective.

pkg/cmd/template.go
```diff
@@ -20,10 +21,34 @@ type compositionData struct {
 func compileCompositionTemplate(path string, input *compositionData) (*bytes.Buffer, error) {
 	templateDir := filepath.Dir(path)
 
+	// Investigate: https://github.com/Masterminds/sprig
+	// Investigate "missingkey=error"
```
galargh, line 25: I think we might be missing some context here. I wouldn't be too sure how to proceed based on this.

galargh (approved):
Thanks for the explanation yesterday 🙇 Looks good, nothing blocking on my side 🚀

pkg/api/composition.go
```diff
@@ -273,294 +310,195 @@ type Dependency struct {
...
-	runners := make([]string, 0, len(manifest.Runners))
-	for k := range manifest.Runners {
-		runners = append(runners, k)
-	}
-	sort.Strings(runners)
-	if sort.SearchStrings(runners, c.Global.Runner) == len(runners) {
-		return nil, fmt.Errorf("plan does not support runner %s; supported: %v", c.Global.Runner, runners)
-	}
+func WriteCompositionToFile(comp *Composition, file string) error {
+	f, err := os.Create(file)
```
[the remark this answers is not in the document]
laurentsenta: It overrides the file (truncate)

pkg/cmd/template.go
laurentsenta: moved to follow-ups

pkg/api/composition_validation.go
laurentsenta: Legacy code, the validation was doing some validation + maths.

pkg/api/composition.go
```diff
@@ -273,294 +310,195 @@ type Dependency struct {
...
+		ids = append(ids, x.ID)
+	}
+	sort.Strings(ids)
+	return ids
+}
 
-	return result
+// CalculatedInstanceCount returns the actual number of instances in this group.
+//
+// Validate MUST be called for this field to be available.
```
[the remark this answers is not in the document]
laurentsenta: Agreed,
but there's no "good" check for this other than rethink the whole thing into a getter rather than precompute it here.
(checking that the value is != 0 would hide bugs like "I changed the composition without recomputing indices")

I would like to simplify and refactor the composition format as a whole,
This PR is "only" adding a comment that makes this explicit, the issue existed before.

pkg/api/composition.go
```diff
@@ -273,294 +310,195 @@ type Dependency struct {
...
-	// $TESTGROUND_HOME/plans; the server doesn't care about our local
-	// paths.
-	c.Global.Plan = manifest.Name
-
-	// validate the test case exists.
-	_, tcase, ok := manifest.TestCaseByName(c.Global.Case)
-	if !ok {
-		return nil, fmt.Errorf("test case %s not found in plan %s", c.Global.Case, manifest.Name)
-	}
+// Validate MUST be called for this field to be available.
```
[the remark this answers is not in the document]
laurentsenta: Same as above

pkg/api/composition.go
```diff
@@ -273,294 +310,195 @@ type Dependency struct {
...
-	// $TESTGROUND_HOME/plans; the server doesn't care about our local
-	// paths.
-	c.Global.Plan = manifest.Name
-
-	// Is the builder supported?
-	if manifest.Builders == nil || len(manifest.Builders) == 0 {
-		return nil, fmt.Errorf("plan supports no builders; review the manifest")
-	}
+// FrameForRuns clones this composition, retaining only the specified run ids and corresponding groups
+// TODO: Verify that our composition and nested objects are immutables / deep copies.
```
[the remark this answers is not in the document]
laurentsenta: Removed, I verified while doing the implementation.

The next step is to rewrite composition files (see below)

pkg/cmd/run.go
laurentsenta: What about I create a follow-up:

- [ ] implement server-side runs feature
  - [ ] refactors tasks to support task groups
  - [ ] refactor the supervisor to support task groups
  - [ ] refactor the request system to support task groups
  - [ ] remove the need for waiting on the client side


?

pkg/api/composition.go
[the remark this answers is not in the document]
laurentsenta: (resolving to merge)

pkg/api/composition.go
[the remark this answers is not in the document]
laurentsenta: (resolving to merge)

pkg/api/composition_validation.go
laurentsenta: (resolving to merge)

pkg/cmd/run.go
laurentsenta: (resolving to merge)

pkg/api/composition.go
```diff
@@ -273,294 +310,195 @@ type Dependency struct {
...
-		}
+	return result
+}
 
-		// Update the group's calculated instance counts.
-		if g.calculatedInstanceCnt = g.Instances.Count; g.calculatedInstanceCnt == 0 {
-			g.calculatedInstanceCnt = uint(math.Round(g.Instances.Percentage * float64(totalInstances)))
+// PickGroups clones this composition, retaining only the specified groups.
+func (c Composition) PickGroups(indices ...int) (Composition, error) {
+	for _, i := range indices {
```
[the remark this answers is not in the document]
laurentsenta: (resolving to merge)

Issue #1355, opened 2022-06-15
EPIC: Clarify and Simplify logging


laurentsenta:
There are a few assumptions made by testground about logging that looks legacy / quite specific to the go-sdk,

The go-sdk writes json payloads to both the stdout and files like the `run.out` and `run.err`.

There are also different ways to process these outputs: the `pretty.go` is specific to local docker runner and might change test outcomes, the `run.out` files are collected by testground and used during testing, etc.

This EPIC aggregates discussion about this subject until we're ready to tackle this.

## Related Discussions

https://github.com/testground/testground/pull/1327#discussion_r896576693
https://github.com/testground/testground/issues/1322#issuecomment-1141209321

laurentsenta:
Another issue raised during IPFS Thing is that testground logging is quite cryptic at first look. The dashboard logs don't have syntax coloring (they display raw terminal color escapes) and it's hard for a new user to tell what is happening when looking at the logs.

I believe prefixing/detailing build logs (instead of dumping them from docker directly), and making sure each step is clearly described might help.

Making sure users note the existence of the dashboard might help too.

laurentsenta:
Another issue:
when the daemon logs warnings they are not sent back to the client, which means that these can be found ONLY in the daemon logs. Warnings should be forwarded to the client, they are much much more visible there.

laurentsenta:
Difference between stdout and record_message is not clear: https://github.com/testground/testground/pull/1520#discussion_r1022881066

PR #1520, opened 2022-11-11
example-rust: improve test-case:

jxs:
- use tokio's TcpListener and TcpStream to not block the executor thread.
- remove remaining unwraps.
- cargo fmt

### DIFF
--- a/plans/example-rust/src/main.rs
+++ b/plans/example-rust/src/main.rs
@@ -1,4 +1,5 @@
-use std::net::{Ipv4Addr, TcpListener, TcpStream};
+use std::net::Ipv4Addr;
+use tokio::net::{TcpListener, TcpStream};
 
 const LISTENING_PORT: u16 = 1234;
 
@@ -7,11 +8,10 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let (client, _run_parameters) = testground::client::Client::new().await?;
     client.wait_network_initialized().await?;
 
-    let local_addr = &if_addrs::get_if_addrs()
-        .unwrap()
+    let local_addr = &if_addrs::get_if_addrs()?
         .into_iter()
         .find(|iface| iface.name == "eth1")
-        .unwrap()
+        .ok_or("Could not find interface eth1")?
         .addr
         .ip();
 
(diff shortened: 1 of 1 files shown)


laurentsenta (approved):
Looks good to me 👍

plans/example-rust/src/main.rs
```diff
@@ -7,51 +8,43 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            let listener = TcpListener::bind((*addr, LISTENING_PORT)).await?;
 
             client.signal("listening".to_string()).await?;
 
-            for _stream in listener.incoming() {
-                println!("Established inbound TCP connection.");
-                break;
-            }
+            println!("Established inbound TCP connection.");
+            let _ = listener.accept().await?;
```
mxinden, line 27:
```suggestion
            let _ = listener.accept().await?;
            println!("Established inbound TCP connection.");
```
This is in the wrong order, right?

plans/example-rust/src/main.rs
```diff
@@ -7,51 +8,43 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-                .await?;
+            client.barrier("listening".to_string(), 1).await?;
 
             let remote_addr: Ipv4Addr = {
                 let mut octets = addr.octets();
                 octets[3] = 2;
                 octets.into()
             };
-            let _stream = TcpStream::connect((remote_addr, LISTENING_PORT)).unwrap();
-            println!("Established outbound TCP connection.");
```
mxinden, line 44: Why drop the `println` on the dialer side?

plans/example-rust/src/main.rs
jxs: yes, thanks!

plans/example-rust/src/main.rs
jxs: No reason, I switched the `println!`'s for `client.record_message` to understand what would be the difference on the logs and then missed this when reverting thanks for catching that Max.  

I noticed it's that `println!`'s use the `OTHER` scope while `client.record_message` uses the `MESSAGE` scope in the logs. Is there a place where the difference between using one or the other is explained? 
[`record_message`](https://docs.rs/testground/0.4.0/testground/client/struct.Client.html#method.record_message) doc seems to be missing a description also, I can then update it.

mxinden (approved):
Thanks @jxs!

plans/example-rust/src/main.rs
laurentsenta: Not really, 
`record_message` sends structured messages to the sync service, it should be capable to hold more info (like log severity, etc)
`println!` is "just" captured by testground and forwarded to the client.

Related: https://github.com/testground/testground/issues/1355

Issue #1491, opened 2022-10-10
testground 2022 Q4/2023 Q1 Roadmap


laurentsenta:
Current Status of Roadmap:

- 2022-10-10: We don't have full maintainer alignment on Roadmap items and priorities. As we resolve these, we will update the roadmap.

Unresolved questions:

- TBD

Description
This issue is intended to capture discussion around Testground's Roadmap
Because GitHub doesn't have any feature to comment on Markdown files, please use this issue to leave any feedback, proposals, etc.

Roadmap drafts:
- 1st draft: https://github.com/testground/testground/pull/1484

laurentsenta:
Our current roadmap "make sense" with the needs that came up from using Testground this year,

Once the dust settles, I'd like to make sure we review the 2019 ROADMAP, no doubt it contains works and improvements discussion we want to reuse. 
 ttps://github.com/testground/docs/blob/88f7e03f7b7ee5715c3c36072787dada4b2b12ca/misc/ROADMAP.md

laurentsenta:
Open discussion: https://github.com/testground/testground/pull/1484#discussion_r992168145

BigLep:
@laurentsenta : thanks for creating this issue. 
1. Minor: should we maybe add a section to the README to improve discoverability of the ROADMAP?  (Also, I pinned this issue.)
2. We say things like "Effort: approx. 6 months".  Does this mean "Estimate of 6 developer months from start to finish, assuming a developer worked on this initiative full time"?
3. I'm not disagreeing with your sequencing and I'm not pressuring to reduce estimates, but I would love for us to see if there are any creative ways to bring in "Provide a Testground As A Service Cluster used by libp2p & ipfs teams".  It feel like a bummer to be waiting until Q3 for this :(  If the answer is in hiring or in collaborations, I'd like to see us do a focused investment in Q4 so we can pull this in.
4. "Support libp2p's interoperability testing story and ProbeLabs work as a way to drive "critical" Testground improvements"
   1. I think we need to solve "Javascript & Browser support in Testground" earlier given its criticality for libp2p's interoperability story and for libp2p to be able to validate the new functionality they're working on with webrtc and webtransport. 
   2. Of everything else in the list, I would put remote runners and performance benchmarking at the top (but those maybe just deserve their own milestone).  We can tackle that when we get closer.

In reading through this roadmap, it mostly made me want to find a way to pull in the milestones sooner.  That's a good sign.  There's lots of critical stuff to get delivered as soon as possible.

[...]

Issue #1378, opened 2022-07-06
Publishing from 1 instance out of N(rest subscribes only) hangs further test logic


Bidon15:
#### Version information:

2b3b5d92bac73cd2a78a37555de0c5b8884161bc

#### Description:
[This doc](https://docs.testground.ai/v/master/writing-test-plans/communication-between-instances) explains how many-to-many approach works perfectly for testground.

Unfortunately, if we go with the approach that only 1 out of N publishes per topic and the rest subscribes to the event, then further logic hangs. 
I've uploaded some examples into this [repo](https://github.com/Bidon15/test-ground)

if this behaviour is an expected one and not a bug, then what is the idiomatic way to implement the 1 pub and many subscribes scenario?

Bidon15:
Oh yes. You are absolutely right @laurentsenta! I was blind sided on this for quite some time. 
Closing as your point actually helped me a lot on reviewing/fixing the main test logic that caused this 🙏 
Appreciate your time 👍

Issue #1336, opened 2022-05-13
Docs: Role assignment/usage in test-case example


Bidon15:
## Description
After closely following the [learning-example-tg](https://github.com/testground/learning-example-tg/blob/master/manifest.toml) , I've seen that the example is using sequence numbers for role assignment of producer and consumers. 

It will be awesome if there is an example where explicit [role](https://github.com/testground/sdk-go/blob/49c90fa754052018b70c63d87b7f1d37f6080a78/runtime/runparams.go#L31) assignment/usage is displayed for devs to understand how to write tests where there are multiple actors(like bootstrappers, clients, full nodes, etc. or just simple A, B, C) involved in the test case

Right now, the only place where they(roles) are explicitly defined is [lotus-soup/testplans](https://github.com/filecoin-project/lotus/blob/master/testplans/lotus-soup/_compositions/baseline-docker-5-1.toml), which involves understanding of compositions/global params

Hence, the learning curve can be eased out if there is a simple roles' assignment/usage + some record messages in the learning examples repo(s)

## What defines this endeavour to be complete?

- [ ] Simple example like existing [learning-example-tg](https://github.com/testground/learning-example-tg) where roles are included in the manifest and how they are utilised/assigned in the test-case
   - [ ] This can even be simplified like what is mentioned in the docs for [params and test-cases](https://docs.testground.ai/v/master/writing-test-plans/paramaters-and-testcases)
   - [ ] A dockerfile example will also help understand how everything is packed

brdji:
Sorry for not replying sooner, I think this is a great idea! 

I've started working on this yesterday, and have found that roles are not read from the manifest file, but apparently only from the  composition file. Here's the `TestCase` struct in `manifest.go`:
https://github.com/testground/testground/blob/43e6b3ec43eb1be9b7c79ded99eb54c9b375d7a6/pkg/api/manifest.go#L29-L34

This would mean that this example would have to be written using a simple `composition` file, which isn't too bad, and can be useful as a learning tool not just for roles, but for composition files, too.

brdji:
I have added an example using a simple composition file. 

> Is it planned to make the roles readable from manifests as stand-alone option without compositions?

I'm not sure that roles would make sense without adding instance groups, instance numbers, etc. to the manifest file, as well. That would mean the manifest file would basically turn into the composition file, and one of them would be superfluous. 
I believe that roles, or other additional parameters are intended to be passed in through the `[testcases.params]` section in the manifest.

Issue #1275, opened 2021-10-28
empty results for executed test-plan


Bidon15:
#### Version information:

git rev-parse --verify HEAD
0b997a15b2df79356d262e2afe69f85fadd08545

Setup 1 - MBP 2018: 
- Intel x86 64bit
- OS Big Sur 11.6
- Go ver: 1.17

Setup 2 - Digital Ocean Droplet : 
- x86 64bit
- Ubuntu 20.04.2 LTS
- Go ver: 1.17

#### Description:

Pre-Reqs:

- Cloned testground repo
- Installation done via `make install`

Steps to reproduce: 

1. Open terminal window
2. Execute `testground daemon`
3. Open another terminal window
4. Execute `testground plan import --from ./plans/network`
5. `testground run single --plan=network --testcase=ping-pong  --builder=docker:go --runner=local:docker --instances=2`

Expected result: 
- Results info log should be provided with status of executed test plan (success/fail/etc.)

Actual result:
- No logs for results like `INFO run finished successfully ...`

Note:

I initially thought these logs should be visible after collecting logs. However, `results.out` are empty, too

hacdias:
Hey! Testground, by default, runs asynchronously. To get logs live, you have to append the flag `--wait`. You can also fetch a list of the current tasks with `testground tasks`.

If you also want to collect a tarball with the logs outputted by your run, you either append `--collect`, as well as `--wait`, or you wait until the execution of the run ends and execute `testground collect --runner local:docker [run-id]`.

Let me know if you need more help!

Bidon15:
Hey! Thanks for the hint @hacdias 
Indeed --wait helped a lot with following up on what is happening in the running test plan terminal.
Though, I still don't get the idea why the results.out files are empty on both 0/1 after execution of the test plan 
Overall logs are clearly indicating that everything is fine
`{"ts":1639132301477845363,"msg":"","group_id":"single","run_id":"c6pip2gnt2ojiuff6v7g","event":{"success_event":{"group":"single"}}}`

Or does results mean the execution of the plan itself with no report on fail/pass/skip for test cases/scenarios?

Bidon15:
Oh, I guess I found the answer myself: `At the moment Testground doesn't provide any tools for post-processing of outputs. You are free to submit ideas and pull requests and contribute to this feature`

I assume it's still true,  so I'll close this issue

PR #1532, opened 2022-11-25
doc: add documentation and rollups example to roadmap's vision

Bidon15:
Notes on why imho 3 points should be included in the vision section: 

### Documentation is king for wider adoption

Currently, our documentation is very `todo-list`ish and users(me included back in time) faced harshed realities of implementation for their projects' needs. Async hand-holding is needed as I've heard multiple times that nailing down the concept of testground is not straight-forward and is causing issues like: 
- https://github.com/testground/testground/issues/1336 - compositions are hard to nail down. Especially when a user opens the `runenv` APIs and discovers that there are multiple ways to treat the test-run instance's perspective
- https://github.com/testground/testground/issues/1378 - classic Dunning-Kruger effect that should be populated somewhere in docs that blocking calls should be treated with extra-caution when using sync-service APIs
- https://github.com/testground/testground/issues/1275 - what is results exactly? Somebody can think like me that it's the pass/failure which is not the case 

### Large-scale networks(example for a rollup)

Going forward with the L1-L2-L(N) narrative, it will be beneficial if we think forward what users might need from testground to test their rollups, or even a complete e2e system test, like:
A Button call-> Triggers a smart contract -> state transition in the rollup -> dumps data to a DA Layer. And then we check the new data at the UI level. Now, imagine how many variations we can have for a single test-case (network wise as well as VM, DA, instances count, etc.)

### Selenium/Cypress

Both are golden standards for web-browser automation and I do believe that testground should be a one-stop shop for blockchains' / p2p network testing/experimentation

Ref: #1491

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -57,7 +57,7 @@ Testground As A Service embodies our long-term vision.
 - An easy-to-follow documentation and guidelines for users to jump straight into action,
   - An example: pages that can hand-hold through the learning curve, thus fostering community-driven tutorials. As well as APIs section that can help navigate through the vast possibilities of testground.
 - The ability to experiment with large-scale networks and simplify the integration testing of "any" network-related code,
-  - An example: experiment with a set of rollup networks that deploy their data to a different DA network. A unified/standardised way of doing network(s)-to-network(s) experiments would eliminate a lof of bike-shedding that future teams will go through. 
+  - An example: having the ability to define standard experiments with a set of roll-up networks that deploy their data to a different DA network. This would help teams compare solutions using the same metrics.
 - The ability to track the impact of a change in terms of stability & performance across multiple projects,
   - An example: having the ability to run IPFS benchmarks and simulations with different combinations of libraries. This would help us measure regression and improvements as soon as they occur in upstream dependencies.
 
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -54,7 +54,10 @@ _This section is still a very high-level draft_
 Testground As A Service embodies our long-term vision.
 
 - A single, scalable, installation that one or more organizations can use in all their projects,
+- An easy-to-follow documentation and guidelines for users to jump straight into action,
+  - An example: pages that can hand-hold through the learning curve, thus fostering community-driven tutorials. As well as APIs section that can help navigate through the vast possibilities of testground.
 - The ability to experiment with large-scale networks and simplify the integration testing of "any" network-related code,
+  - An example: experiment with a set of rollup networks that deploy their data to a different DA network. A unified/standardised way of doing network(s)-to-network(s) experiments would eliminate a lof of bike-shedding that future teams will go through. 
 - The ability to track the impact of a change in terms of stability & performance across multiple projects,
   - An example: having the ability to run IPFS benchmarks and simulations with different combinations of libraries. This would help us measure regression and improvements as soon as they occur in upstream dependencies.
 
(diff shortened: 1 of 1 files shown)


Bidon15:
I haven't added this into the PR but leaving another point here:
It's community plugins for testground being featured in the repo

ROADMAP.md
```diff
@@ -54,14 +54,18 @@ _This section is still a very high-level draft_
 Testground As A Service embodies our long-term vision.
 
 - A single, scalable, installation that one or more organizations can use in all their projects,
+- An easy-to-follow documentation and guidelines for users to jump straight into action,
+  - An example: pages that can hand-hold through the learning curve, thus fostering community-driven tutorials. As well as APIs section that can help navigate through the vast possibilities of testground.
 - The ability to experiment with large-scale networks and simplify the integration testing of "any" network-related code,
+  - An example: experiment with a set of rollup networks that deploy their data to a different DA network. A unified/standardised way of doing network(s)-to-network(s) experiments would eliminate a lof of bike-shedding that future teams will go through.
```
laurentsenta, line 60: let's drop the wording around `bike-shedding`, maybe:


```suggestion
  - An example: having the ability to define standard experiments with a set of roll-up networks that deploy their data to a different DA network. This would help teams compare solutions using the same metrics.
```

ROADMAP.md
```diff
@@ -54,14 +54,18 @@ _This section is still a very high-level draft_
...
 - The ability to experiment with large-scale networks and simplify the integration testing of "any" network-related code,
+  - An example: experiment with a set of rollup networks that deploy their data to a different DA network. A unified/standardised way of doing network(s)-to-network(s) experiments would eliminate a lof of bike-shedding that future teams will go through. 
 - The ability to track the impact of a change in terms of stability & performance across multiple projects,
   - An example: having the ability to run IPFS benchmarks and simulations with different combinations of libraries. This would help us measure regression and improvements as soon as they occur in upstream dependencies.
 
 Products with similar ideas but specialized in different areas:
 
 - database: [CockroachDB performance tracker](https://cockroachdb.github.io/pebble/?max=local),
 - browser: [Webkit Performance Dashboard](https://perf.webkit.org/v3/)
+- selenium : [Selenium](https://www.selenium.dev/)
```
laurentsenta, line 68: Could you share why selenium is useful as a comparison here?

The idea behind Testground As A Service is a unified platform with a dashboard that lets you track changes to your performance metrics, pinpointed to a single commit (similar to https://perf.webkit.org/v3), and eventually run new experiments.

I know selenium as a programmable browser, like playwright or pupetteer, which is more one tool your test might use in TaaS.

Bidon15: > Could you share why selenium is useful as a comparison here?
> 
> I know selenium as a programmable browser, like playwright or pupetteer, which is more one tool your test might use in TaaS.

Selenium/Cypress and others are a common one stop solutions for those devs/qa engineers to develop e2e automated tests on their stack (front+back). Usually, such tests are executed on a job basis(e.g. every RC promotion) and then a team does manual testing if needed later

IMHO, testground can cover those e2e automated tests needs for blockchains. Especially, when we are moving towards a multi/cross chain vision in the industry. In addition, we are heavily relying on manual labour of community members is public testnets to perform checks. Some of the manual labour(especially simple like token transfer between accounts, chains) can be eliminated with testground spinning up a full fledge chain(s) and perform checks without humans' participation during the whole execution process(regression, release, smoke, sanity, etc.).  

Now, we don't have a golden standard in writing e2e big network tests where I can schedule a job with a set of test-plans to execute on every release candidate of XYZ client release.

Once we have the standard, a floodgate of new engineers will chime in to learn about QA not only smart-contract wise testing but protocol testing itself, which I believe is crucial considering that appchains' narrative is gaining momentum

laurentsenta:
(rebased on latest master)

laurentsenta (approved):

PR #1537, opened 2022-11-30
feat: port example and local exec tests to go tests and github CI

laurentsenta:
Contributes to #1368

Currently, the tests are written in shell, which is hard to maintain, improve, and reuse.
With this PR, we move (some tests) to golang. We'll split this work into a few PR to make it
easier to review.

We call the CLI object from the go tests and start the daemon by creating an actual server object.
We don't go through a testground binary. Which means we can get code coverage and a simpler debugging experience.

- [x] Create an integration test package that relies on golang tool, instead of shell scripts
    - [x] Implement the tooling required: a few run commands, a few `requireXXX` assertions
- [x] port local exec tests
- [x] port docker examples
- [x] make it more practical to use the CLI from a test
    - [x] refactor the generic parser for response to make decoding/printing explicit
    - [x] update the CLI tooling to rely on the passed stdout/stderr rather than printing to the console directly
- [x] Add a basic CI workflow

Follow-up Tasks:

- [ ] port docker tests and add more of the tooling required,
- [ ] move `itest` content to these tests,
- [ ] remove the shell integration tests


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,37 @@
+on: [push, pull_request]
```
galargh, line 1: I'd suggest the following so that we don't double work as much.

```suggestion
on:
  push:
    branches:
      - master
  pull_request
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,37 @@
...
+        os: [ "ubuntu" ]
+        go: [ "1.16.x" ]
+    env:
+      COVERAGES: ""
+    runs-on: ${{ format('{0}-latest', matrix.os) }}
+    name: "[${{ matrix.suite }}] on ${{ matrix.os }} (go ${{ matrix.go }})"
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: recursive
```
galargh, line 23: Do we have submodules here? It rings a bell but I cannot remember for what.

pkg/integration/utils/cli.go
```diff
@@ -0,0 +1,82 @@
...
+		}
+	}
+}
+
+func ExtractTarGz(src, dst string) error {
+		targzStream, err := os.Open(src)
+		if err != nil {
+			return err
+		}
+		defer targzStream.Close()
```
galargh, line 35:
```suggestion
	targzStream, err := os.Open(src)
	if err != nil {
		return err
	}
	defer targzStream.Close()
```

pkg/integration/utils/cli.go
```diff
@@ -0,0 +1,82 @@
...
+	"io"
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+
+	"testing"
+)
+
+// use the CLI and call the command `docker pull` for each image in a list of images
```
galargh, line 19: Do we want to add reasoning why we're not using the docker library? I'm not saying that we should. Just thought it could be nice to have it mentioned.

pkg/integration/utils/entrypoints.go
```diff
@@ -0,0 +1,135 @@
...
+			if err := os.RemoveAll(collectedDestination); err != nil {
+				t.Fatal(err)
+			}
+		}
+	}
+
+	return result, err
+}
+
+func RunAComposition(t *testing.T, params RunComposition) (*RunResult, error) {
```
galargh, line 93: I feel the naming here is a bit tricky (because it's almost the same but isn't). Would it make sense to rename the `RunAComposition` to something else? Maybe `RunCompositionTest`? And then `Run` func would become `RunTest` too?

pkg/integration/utils/requires.go
```diff
@@ -0,0 +1,79 @@
...
+	require.NoError(t, err)
+	require.False(t, f.IsDir())
+	require.NotZero(t, f.Size())
+
+	// require file is empty
+	f, err = os.Stat(singleErr)
+	require.NoError(t, err)
+	require.False(t, f.IsDir())
+	require.Zero(t, f.Size())
+}
```
galargh, line 79:
```suggestion
}

```

pkg/integration/utils/entrypoints.go
galargh: Or maybe the types should become `RunSingleParams`, `RunCompositionParams`, etc.

pkg/integration/utils/utils.go
```diff
@@ -0,0 +1,58 @@
...
+			t.Fatal(err)
+		}
+	}
+
+	if err = os.Chdir(dir); err != nil {
+		t.Fatal(err)
+	}
+
+	return nil, dir, cleanup
+}
```
galargh, line 58:
```suggestion
}

```

galargh (approved):

.github/workflows/integration-tests.yml
laurentsenta: We don't at the moment, I removed it 🙏

pkg/integration/utils/entrypoints.go
laurentsenta: Thanks for catching this; agreed, I fixed the naming 👍

pkg/integration/utils/cli.go
laurentsenta: Mostly for simplicity, I updated the comment to clarify that the fact we use CLI is secondary.

PR #1539, opened 2022-12-05
feat: improve browser examples

GlenDC:
Implement follow ups in my power as mentioned by @laurentsenta in https://github.com/testground/testground/pull/1502#pullrequestreview-1199287441.

Summary:

- replace the `BrowserKind` test paramater with `runtime`, which by default is `node`, but
   - can also be set to `chromium`, `firefox` and `webkit`.
 - enable the sync test for the browser-node example
    - also added a sh script to run that same test for the regular `js` example (while I was at it)

For the sdk-js (third) point I created an issue at https://github.com/testground/sdk-js/issues/28. I do not mind to also work on that one everything that still might need to be done before you can make a release. The release however will have to be done by the sdk-js maintainers (you?).

In function of #1386


laurentsenta:
Thanks for the care you put into this PR, it looks great,
with the sync test, at the moment, we run 1 sync test with 2 node instances, 1 sync test with 2 firefox instances, etc.

Since the next step for this work is to test interoperability, we would cover more features by having 1 sync test that runs with 1 node instance, 1 firefox, and 1 chrome. together. Could you look into creating this composition? I can help with writing it.

GlenDC:
> Thanks for the care you put into this PR, it looks great, with the sync test, at the moment, we run 1 sync test with 2 node instances, 1 sync test with 2 firefox instances, etc.
> 
> Since the next step for this work is to test interoperability, we would cover more features by having 1 sync test that runs with 1 node instance, 1 firefox, and 1 chrome. together. Could you look into creating this composition? I can help with writing it.

@laurentsenta the integration test cross-runtime for sync has been added. This one tests using an instance of node, chromium, firefox and webkit. Can you please re-review?

plans/example-browser-node/compositions/sync-cross-runtime.toml
```diff
@@ -0,0 +1,39 @@
...
+name    = "sync-cross-runtime"
+author  = "Glen De Cauwsemaecker"
+
+[global]
+plan    = "example-browser-node"
+case    = "sync"
+builder = "docker:generic"
+runner  = "local:docker"
+
+total_instances = 4
```
laurentsenta, line 11: For what it's worth, if you don't use percentages in your groups, this field is optional since Testground can compute the value.

laurentsenta (approved):

plans/example-browser-node/compositions/sync-cross-runtime.toml
GlenDC: Addressed it for the sake of completness.

PR #1544, opened 2022-12-07
feat: Use XDG Base Directory Specificationfor TESTGROUND_HOME directory

dougEfresh:
Having TESTGROUND_HOME directory in `~/testground` pollutes my $HOME. There are standards for application data/cache/config across operating systems, see https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html and https://github.com/adrg/xdg

For linux/unix:
`~/.config/testground`

For Mac/Darwin:

`~/Library/Application Support/testground`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `[[runs]]` field to compositions. See [PR 1516]
 - Add `--run-ids` option during `testground run`. See [PR 1516]
 - Add `--result-file ./output.csv` option during `testground run`. See [PR 1516]
+- Move default `TESTGROUND_HOME` from `~/testgraound` to xdg directory specification. See [PR 1544]
 
 ### Fixed
 - Fix dependencies rewrites in the `exec:go` builder. See [PR 1469]
@@ -24,3 +25,4 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [PR 1469]: https://github.com/testground/testground/pull/1469
 [PR 1481]: https://github.com/testground/testground/pull/1481
 [PR 1537]: https://github.com/testground/testground/pull/1537
+[PR 1544]: https://github.com/testground/testground/pull/1544
--- a/install.sh
+++ b/install.sh
@@ -1,8 +1,13 @@
 #!/usr/bin/env sh
 set -e
 set -o pipefail
+unameOut="$(uname -s)"
+TESTGROUND_HOME="$HOME/.config/testground"
+case "${unameOut}" in
+    Darwin*)    TESTGROUND_HOME="$HOME/Library/Application Support";;
+esac
 
-mkdir -p ~/testground && cd ~/testground
+mkdir -p "$TESTGROUND_HOME" && cd "$TESTGROUND_HOME"
 
 docker pull iptestground/testground:edge
 docker pull iptestground/sync-service:edge
@@ -12,7 +17,7 @@ docker pull iptestground/sidecar:edge
 docker run -v ${PWD}:/mount --rm --entrypoint cp iptestground/testground:edge /testground /mount/testground
 
 if [ -z $GITHUB_PATH ]; then
-    echo "Testground was installed to ~/testground. Add this to your path."
+    echo "Testground was installed to \"$TESTGROUND_HOME\". Add this to your path."
 else
-    echo "~/testground" >> $GITHUB_PATH
+    echo "$TESTGROUND_HOME" >> $GITHUB_PATH
 fi
--- a/pkg/config/loader.go
+++ b/pkg/config/loader.go
@@ -6,7 +6,7 @@ import (
 	"path/filepath"
 
 	"github.com/BurntSushi/toml"
-
+	"github.com/adrg/xdg"
 	"github.com/testground/testground/pkg/logging"
 )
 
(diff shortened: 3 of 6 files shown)


laurentsenta:
Thanks for the contribution @dougEfresh 
Before merging, we'll have to verify the impact on other setups (like EKS infra).

Could you add a line to the CHANGELOG.md?

pkg/config/loader.go
```diff
@@ -59,25 +59,22 @@ func (e *EnvConfig) EnsureMinimalConfig() error {
...
-			return fmt.Errorf("failed to obtain user home dir: %w", err)
-		}
-		home = filepath.Join(v, "testground")
 	}
 
 	switch fi, err := os.Stat(home); {
 	case os.IsNotExist(err):
 		logging.S().Infof("creating home directory at %s", home)
-		if err := os.MkdirAll(home, 0777); err != nil {
+		if err := os.MkdirAll(home, 0770); err != nil {
```
laurentsenta, line 77: I think you want to leave read and execute here, could you clarify why we want a `0`?

dougEfresh: I personally do not like world readable directories, but I'll change it back.

dougEfresh:
Could someone test this on a mac? I only run linux :)

dougEfresh:
> * Could you also update related documentation? like: https://docs.testground.ai/v/master/table-of-contents/getting-started#running-testground

See https://github.com/testground/docs/pull/24

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `[[runs]]` field to compositions. See [PR 1516]
 - Add `--run-ids` option during `testground run`. See [PR 1516]
 - Add `--result-file ./output.csv` option during `testground run`. See [PR 1516]
+- Move default TESTGROUND_HOME from ~/testgraound to xdg directory specification. See [PR 1544]
```
laurentsenta, line 15:
```suggestion
- Move default TESTGROUND_HOME from `~/testground` to xdg directory specification. See [PR 1544]
```

install.sh
```diff
@@ -12,7 +17,7 @@ docker pull iptestground/sidecar:edge
...
 
 if [ -z $GITHUB_PATH ]; then
-    echo "Testground was installed to ~/testground. Add this to your path."
+    echo "Testground was installed to \"$TESTGROUND_HOME/testground\". Add this to your path."
 else
-    echo "~/testground" >> $GITHUB_PATH
-fi
\ No newline at end of file
+    echo "$TESTGROUND_HOME" >> $GITHUB_PATH
+fi
```
laurentsenta, line 23: @galargh would you have time to look into this part? It's supposed to run in CI as well.

CHANGELOG.md
dougEfresh: Also add link to bottom of CHANGELOG

galargh:
Hey! Thanks for proposing the change 😄 

Before we proceed with merging it straight away, I'd want us to consider the impact it might have on the current testground users. From what I gather, if someone is using testground now, doesn't have `TESTGROUND_HOME` set and upgrades to the version with this change in it, then they're imported plans won't work anymore. As such, in my opinion it has to be treated and communicated as a breaking change. And if so, do we want to introduce it on its' own or would we rather batch it up with other breaking changes?

dougEfresh:
> As such, in my opinion it has to be treated and communicated as a breaking change. And if so, do we want to introduce it on its' own or would we rather batch it up with other breaking changes?

How about we make this backward compatible ?  if `~/testground` exists, just use that. Shouldn't be so hard to implement.

dougEfresh:
@galargh I updated this PR to check for legacy directory $HOME/testground and use that.

galargh (commented):
Thank you, Douglas, for the changes! The check for `~/testground` existence addresses my initial concerns well :) 

I added the last set of small review comments. I think we'll be ready to merge after the next iteration 🥳

pkg/config/loader.go
```diff
@@ -59,19 +59,26 @@ func (e *EnvConfig) EnsureMinimalConfig() error {
...
-	var home string
+	// fallback to $HOME/.config/testground (unix/linux) via xdg standard
+	home := filepath.Join(xdg.ConfigHome, "testground")
+	v, err := os.UserHomeDir()
+	if err != nil {
+		return fmt.Errorf("failed to obtain user home dir: %w", err)
+	}
+	legaacyHomePath := filepath.Join(v, "testground")
+	_, err = os.Stat(legaacyHomePath)
+	if err == nil {
```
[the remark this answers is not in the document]
dougEfresh: I thought about this  as well. I don't think at this point in the code to check a specific error. If there is some issue, then later on in the code  `switch fi, err := os.Stat(home);` will catch it. 
But I will add is a check that stat us a Directory.

pkg/config/loader.go
```diff
@@ -59,19 +59,26 @@ func (e *EnvConfig) EnsureMinimalConfig() error {
...
+	}
+	if v, ok := os.LookupEnv(EnvTestgroundHomeDir); ok {
+		// we have an env var.
+		home = v
+	}
+
 	if v, ok := os.LookupEnv(EnvTestgroundHomeDir); ok {
 		// we have an env var.
 		home = v
-	} else {
```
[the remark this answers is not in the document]
dougEfresh: Added a function to check legacy dir or use default home. I think this makes the code more clear.  Plus, easy to remove when the time comes to deprecate $HOME/testground 
Let me know what you think.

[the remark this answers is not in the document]
dougEfresh: Also, one note of interest, `os.UserHomeDir()` has been replaces with `xdg.Home` . Which defaults to `/` if `$HOME` not set.  I think this is acceptable, but wanted to point this out to you.

galargh:
I moved the call to the function you created so that we don't print the warning if the default is not used. I also fixed the test setup and made the warning more explicit. 

All looks good now. I'll test it on mac just to be sure and merge. Thank you for your contribution :)

galargh (approved):

Issue #1542, opened 2022-12-06
Stalled test (likely network failure)


MarcoPolo:
#### Version information:

commit: https://github.com/testground/testground/commit/8629aa281a02bf6a193573a8e0e3959307aadf1d

MacOS M1
#### Description:

Output of test: https://gist.githubusercontent.com/MarcoPolo/f4a152873f8059a88b566d775497adb4/raw/b46e89dfcd40defe418454d8e3fc95a74804aee8/gistfile1.txt

If a container exits with a non-zero status code we should terminate the test.

laurentsenta:
@MarcoPolo I don't have full logs, but I think what you see here is the rust test stalling ([We've seen this happen](https://github.com/libp2p/test-plans/pull/41#issuecomment-1256025245) in the libp2p/test-plan repos before).

So the go test completes, the rust test stalls, and there is probably a 60 minutes timeout by default in your test: https://github.com/libp2p/test-plans/blob/f8250cc3786ab1f9da2d1f999cbda22ed4943521/.github/actions/setup-testground/action.yml#L69

You could drop the timeout to 5 minutes in the testground conf, it won't be a big deal since we're not building with Testground (or disable the daemon in `testground-setup`).

> If a container exits with a non-zero status code we should terminate the test.

If ALL containers exit we terminate the test. Testground waits for the whole test to complete before gathering logs, outputs, metrics, and outcomes.

I'll update the description to emphasis the need to be more explicit about outputs.

## A note on the test itself:

- I tried running the test with only rust: https://github.com/laurentsenta/test-plans/actions/runs/3695981611#summary-10092695972 (passed)
- Only go: https://github.com/laurentsenta/test-plans/actions/runs/3696122500 (passed)
- Then all of them: https://github.com/laurentsenta/test-plans/actions/runs/3697441875

```
"rust-v0.50.0_x_go-v0.22.0(tcp,_noise,yamux)",ced154vo4a4bpm1mo1s0,failure,
"rust-v0.50.0_x_go-v0.23.4(tcp,_noise,yamux)",ced17ofo4a4bpm1mo1u0,failure,
```

- I also tried running twice all the go tests, in case that was an issue with the number of tests, it passed as well: https://github.com/laurentsenta/test-plans/actions/runs/3696390392

As I write this (I re-run all the tests just in case), It seems that the error happened with rust-v0.50.0 x go both times,

laurentsenta:
[...]

PR #1549, opened 2022-12-15
local_docker: make it more explicit when a test is killed when reaching a timeout

laurentsenta:
Relates to #1542
Closes https://github.com/testground/testground/issues/1551

When a job reaches the timeout, it's not very clear that it was canceled.

### DIFF
--- a/pkg/runner/local_docker.go
+++ b/pkg/runner/local_docker.go
@@ -287,11 +287,19 @@ func (r *LocalDockerRunner) Run(ctx context.Context, input *api.RunInput, ow *rp
 	}
 
 	defer func() {
+		log.Infow("run completed", "err", err)
+
 		if err != nil && result.Outcome == "" {
+			log.Infow("run failed", "err", err)
 			result.Outcome = task.OutcomeFailure
 		}
 		if ctx.Err() == context.Canceled {
-			result.Outcome = task.OutcomeCanceled
+			log.Infow("run canceled")
+			result.Outcome = task.OutcomeFailure
+		}
+		if ctx.Err() == context.DeadlineExceeded {
+			log.Infow("run canceled after reaching the task timeout")
+			result.Outcome = task.OutcomeFailure
 		}
 	}()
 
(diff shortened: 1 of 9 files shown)


galargh:
Is it necessarily what we want? To me a timeout feels closer to a failure rather than cancellation because I expect the latter to be intentional and initiated by user. I'd see reaching a timeout as a failure with timeout as a reason. Do you know other testing frameworks that make distinction between failure and cancellation? Maybe it's just my initial thinking that's off :P

laurentsenta:
I see your point, "canceled" might sound like a user action.

I would prefer being explicit:
- a failure happens when your code failed,
- Testground might end a test for various reasons, because it's configured as such, timeout is one,

To answer directly your question: I don't think go does, but it's famously know to be the bare minimum, jest logs a "timeout reached" I think, I'd have to look into others.

What about github actions? There is a global timeout on the platform, do you know what happens when we reach this timeout?

galargh:
Here's how it goes for GH Actions:
- timeout case: https://github.com/galargh/.github/actions/runs/3704316375
- cancelled case: https://github.com/galargh/.github/actions/runs/3704333550

In both cases the job is marked as cancelled. but in the timeout case the workflow run is marked as failure whereas in the cancelled case it's marked as cancelled.

laurentsenta:
Ok, let's mimick github on that one, please don't merge this PR until I have time to rework it, I'd like to at least log an explicit message on timeout.

laurentsenta:
@galargh PR updated and ready for review!

pkg/runner/local_docker.go
```diff
@@ -287,11 +287,19 @@ func (r *LocalDockerRunner) Run(ctx context.Context, input *api.RunInput, ow *rp
...
+		log.Infow("run completed", "err", err)
+
 		if err != nil && result.Outcome == "" {
+			log.Infow("run failed", "err", err)
 			result.Outcome = task.OutcomeFailure
 		}
 		if ctx.Err() == context.Canceled {
-			result.Outcome = task.OutcomeCanceled
+			log.Infow("run canceled")
+			result.Outcome = task.OutcomeFailure
```
galargh, line 298:
```suggestion
			result.Outcome = task.OutcomeCanceled
```

galargh (approved):
We lost `OutcomeCancelled` in the process :P Looks good otherwise. I'll let you decide if we need a test case for this.

Approving so that you can merge after applying suggestions 👍 👍 👍

PR #1558, opened 2023-01-03
chore: make install.sh install testground to pwd

galargh:
I don't think we want to install `testground` binary to `TESTGROUND_HOME`. It seems more obvious to me that `install.sh` would download the binary to the working directory. In that case, the user would be able to easily decide what that should be.

Here's how we could adjust `libp2p/test-plans` afterwards: https://github.com/libp2p/test-plans/compare/test-install-to-pwd?expand=1

And here's this setup part used in a workflow run still working as expected: https://github.com/libp2p/test-plans/actions/runs/3829913805/jobs/6517126745

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,23 +1,18 @@
 #!/usr/bin/env sh
 set -e
 set -o pipefail
-unameOut="$(uname -s)"
-TESTGROUND_HOME="$HOME/.config/testground"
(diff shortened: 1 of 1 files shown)


laurentsenta (approved):
looks good, thanks for the simplification 👍

PR #1591, opened 2023-09-26
ci: explicitly switch macos build to silicon and upgrade xcode

galargh:
CircleCI is removing support for Apple Intel executors on Oct 2 - see https://discuss.circleci.com/t/macos-resource-deprecation-update/46891. After that date only Apple Silicon executors will be available. This PR ensures the macos builds are executed on Apple Silicon executors only.

FYI, I hardcoded the Go version in the pipeline because something was off with env var substitution. This should be updated if testground is under active development again.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,7 @@ commands:
     description: "install go, checkout and restore cache"
     steps:
       - go/install:
-          version: '{{ .Environment.GOVERSION }}'
+          version: "1.16"
       - checkout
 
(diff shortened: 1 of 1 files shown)


galargh:
As per https://github.com/filecoin-project/rust-fil-proofs/pull/1722#issuecomment-1739139554, it turns out I misunderstood the announcement (timeline) slightly. I'm going to close this PR now and recreate it with correct commit messages and description. I'm sorry for the extra noise.

PR #1594, opened 2024-05-23
feat: support nix builder

yihuang:
nix is very powerful for creating docker image for complex setups, we use it to package a whole python environment.


yihuang:
we'll build the custom image and use the artifact directly.

